PR #13, opened 2020-07-30
Cij dates

Jennit07:
Derived cij dates removed, cij dates added into acute, maternity and mental health for all years. Tested 1617 and 1819 syntax, all seems to work fine.


2015-16/C01 Make Episode File.sps
```diff
@@ -369,7 +362,38 @@ aggregate
     /cij_ppa = Max(PPA).
 
 sort cases by chi keydate1_dateformat.
-
+<<<<<<< HEAD
```
Moohan, line 365: This shouldn't be here. Was there a merge conflict?

2015-16/C01 Make Episode File.sps
```diff
@@ -373,7 +366,7 @@ sort cases by chi keydate1_dateformat.
 
 *  We are not currently including Social Care data for some years but we still want the variables for consistency.
 * Social Care.
- * String
+String
```
Moohan, line 369: This and the CH variables below should not need to be added in the 1516 file

2016-17/C01 Make Episode File.sps
```diff
@@ -354,11 +364,11 @@ aggregate
     /Break chi cij_marker
     /cij_ppa = Max(PPA).
 
-sort cases by chi CIJ_start_date.
+sort cases by chi keydate1_dateformat.
 
 *  We are not currently including Social Care data for some years but we still want the variables for consistency.
 * Social Care.
- * String
+ String
```
Moohan, line 371: Same as 1516 the CH and sc_send_lca should already exists

2015-16/C01 Make Episode File.sps
Jennit07: Line removed

2015-16/C01 Make Episode File.sps
Jennit07: CH variables added back in 1516 file, not too sure about numeric variables. Can you please double check what i have added back in/deleted is correct

2016-17/C01 Make Episode File.sps
Jennit07: CH variables added back in 1617 file, not too sure about numeric variables. Can you please double check what i have added back in/deleted is correct

PR #15, opened 2020-08-24
B tests patch

Jennit07:
changes to DN, CMH B tests (adding in a test for healthboard/LCA) and adding in homelessness tests

### DIFF
--- a/2015-16/B07X DN_tests.sps
+++ b/2015-16/B07X DN_tests.sps
@@ -16,8 +16,38 @@ End if.
 * Flags to count missing values.
 If sysmis(dob) No_DoB = 1.
 
-* Flag to check how many HBs we have.
-AutoRecode hbtreatcode /into HBs.
+*Flag to count how many episodes in each HB by treatment code. 
+If hbtreatcode = 'S08000015' NHS_Ayrshire_and_Arran = 1.
+If hbtreatcode = 'S08000016' NHS_Borders = 1. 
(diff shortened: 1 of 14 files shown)


2015-16/B07X DN_tests.sps
```diff
@@ -16,8 +16,38 @@ End if.
...
+If hbtreatcode = 'S08000022' NHS_Highland = 1.
+If hbtreatcode = 'S08000023' NHS_Lanarkshire =1. 
+If hbtreatcode = 'S08000024' NHS_Lothian = 1. 
+If hbtreatcode = 'S08000025' NHS_Orkney = 1. 
+If hbtreatcode = 'S08000026' NHS_Shetland = 1. 
+If hbtreatcode = 'S08000028' NHS_Western_Isles = 1. 
+If hbtreatcode = 'S08000029' NHS_Fife = 1. 
+If hbtreatcode = 'S08000030' NHS_Tayside = 1. 
+
+*Change missing HB values to 0.
```
Moohan, line 35: You could write this as Recode NHS_A&A to NHS_Tayside (SYSMIS = 0).
Just for future reference, much less writing!

Moohan: Actually I'm not sure it's even necessary to recode them at all, I think the sum below will work fine without it,

Moohan (approved):
Looks good, I think you could cut some blocks of code out without affecting anything but if it works I'm happy to merge!

PR #17, opened 2020-08-24
Allow Highland and Dundee city data for 16/17

Moohan:
Modified slightly to allow data from Highland and Dundee city data for 2016/17 based on comments from David Readhead and Data management.


Jennit07 (approved):
Yeah, looks good to me. Happy to merge in

PR #31, opened 2021-02-25
Add code for testing LTC file

### LINKED ISSUE #30: Tests for LTC file
Create a test that checks for unique upi numbers in the LTC file from IT. Inspired by an issue that came up in the March 2020 update. 

Could possibly test for other things too. Could also do the same uniqueness test on the Deaths file from IT.

Moohan:
Should now output a file to the year folder with any duplicate records, or an empty file in the usual case with no errors.

Partially resolves #30

### DIFF
--- a/2018-19/B12 Create CHI - LTC Lookup.sps
+++ b/2018-19/B12 Create CHI - LTC Lookup.sps
@@ -69,6 +69,39 @@ End Repeat.
 If postcode = "NA" postcode = "".
 
 sort cases by chi postcode.
+
+ * Highlight different CHI numbers.
+add files file = *
+    /first = first_chi
+    /last = last_chi
+    /by chi.
+
+ * Highlight different CHI number / postcode combinations.
+add files file = *
+    /first = first_chi_postcode
+    /last = last_chi_postcode
+    /by chi postcode.
+
+ * Create flags to identify duplicates.
+Numeric duplicate_chi duplicate_chi_postcode (F1.0).
+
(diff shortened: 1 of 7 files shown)


Moohan:
Just realised these changes need copying to all years before merging!

Jennit07:
Think the code works well to show the checks on LTC, however due to earlier fix on the LTC extract, no duplicates in file so the temp save doesnt work to view an error when tested on 1920. Could we do something similar to E01 where the syntax checks for an error before saving the anon_chi? - So use the same syntax for the LTC fix and add in something similar to E01 after identifying LTC duplicates

2018-19/B12 Create CHI - LTC Lookup.sps
```diff
@@ -98,10 +98,11 @@ Else.
     End if.
 End if.
 
-Temporary.
-select if duplicate_chi = 1 or duplicate_chi_postcode = 1.
+If duplicate_chi = 1 or duplicate_chi_postcode = 1 Error = 1.
+Frequencies Error.
```
Moohan, line 102: Could we change this to `crosstabs Error by chi.` that way if there is an error we will also get the duplicate CHI(s) printed in the output.

Jennit07:
Included crosstabs Error by CHI and moved the test code into its own syntax called B12X CHI-LTC_tests and copied to all years. Ready for merging into march-21-update if the changes look okay

Moohan (commented):
Looks good 👍

PR #37, opened 2021-03-25
#24 Move costs into main repo

### LINKED ISSUE #24: Move costs into main repo
Move code into the main repo so it will be tracked. Will need to update file locations in code etc.

Moohan:
Resolve #24.

Still needs checking - I had to change all file paths and update to use macros so there might be issues there.

In the DN costs lookup it reads the HSCP population estimates, I didn't create a macro for this as it will probably clash with other macro changes.

### DIFF
--- a/All years/costs_care_home.sps
+++ b/All years/costs_care_home.sps
@@ -1,0 +1,84 @@
+﻿* Encoding: UTF-8.
+*2. LookUp for costs for Care Homes.
+
+*Get COSLA Value tables.
+*The values are obtained from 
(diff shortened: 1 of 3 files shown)


Moohan:
@Jennit07  I think we discussed this and checked it already?

If you're happy can you merge and close, please?

Jennit07 (approved):
Looks good. Final check before merging i noticed the host at the beginning wasnt copying right in DN and OOH it was missing 'Costs/' in the file path so added this in and happy to merge.

PR #40, opened 2021-04-14
Merge so all_years branch is up to date

Jennit07:
We want to merge master onto the all years branch so the branches are up to date. The most significant change on master is the update of email addresses.


Moohan:
I'll fix all the merge conflicts next week... Looks like a GitKraken job!

PR #42, opened 2021-04-21
adjust discharge dates according to death date

Moohan:
If the discharge occurs 1 - 5 days after death use the death date in place of the discharge date.
Need to adjust all sc_date_2 (the final discharge date) and only the keydate_2 when this is the last record in the full episode.
Use the `death_date` by preference, which is the NRS death date unless this is missing, in which case it is the CHI death date, otherwise try the CHI death date.

### DIFF
--- a/2018-19/B08 Process SOURCE Care Homes.sps
+++ b/2018-19/B08 Process SOURCE Care Homes.sps
@@ -2,21 +2,28 @@
 get file = !Extracts_Alt + "All Care Home episodes.zsav".
 
 * Adjust discharge dates according to death dates.
+* Match on the death dates from the deceased lookup (year specific).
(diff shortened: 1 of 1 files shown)


Moohan:
I think this is ok but just wanted you to have a look @Jennit07 as it is changing ~8% of episodes, which is more than expected.

Have discussed this with Martin, but basically, there is a large(ish) number of episodes where the death_date is before the dis_date, with a lot of these being 3 days before (weirdly) we haven't figured out why that might be, but this is supposed to be a correction to bring dis dates in line with death dates, where they are close but dis is after death.

You might need to check B13 and C06 for some context on the death variables and how they are used in other bits.

PR #44, opened 2021-05-25
Avoid warnings in test files

Moohan:
When the `existing_value` was zero / missing the test syntax was generating warnings for a divide by zero, this needlessly complicates the output. This fix just avoids that warning.


Jennit07 (approved):
Nice fix for stopping the warnings coming up. Tested and is much better

PR #50, opened 2021-08-03
28 Add sigfac to A&E

### LINKED ISSUE #28: Add sigfac to A&E
Currently not included although we do filter to only include records with sigfac = 32 or missing.

Moohan:
Closes #28


Jennit07 (approved):
Merging sigfac in A&E

Issue #49, opened 2021-07-30
Add preventable admissions count and preventable beddays to indiv


Moohan:
Used in Matrix dashboard, should be simple to count in the same way as CIJ admissions.

Moohan:
Code used in matrix:
```
* Use the PLICS episode file to calculate the number of admissions.
get file = !hscdiip_slf + "source-episode-file-20" + !FY + ".zsav"
    /Keep Anon_CHI recid CIJ_marker CIJ_PPA yearstay.

select if any(recid, "01B", "GLS", "02B", "04B").
select if CIJ_PPA = 1.

aggregate outfile = *
    /break Anon_CHI CIJ_marker
    /Preventable_Beddays = sum(yearstay).

aggregate outfile = *
    /Presorted
    /break Anon_CHI
    /Preventable_Admissions = n
    /Preventable_Beddays = sum(Preventable_Beddays).
```

PR #53, opened 2021-08-04
29 Make new cij_delay variable

### LINKED ISSUE #29: Make new cij_delay variable
Variable to indicate if the stay (`cij_marker`) has a delay episode or episodes associated with it.

Moohan:
This is probably not the most efficient way as we're already doing a CIJ level aggregate in `C01` for the `CIJ_PPA` variable. Ideally we'd do this and that together but that would require a new script and re-labelling all the scripts etc.

Assuming you don't spot any issues I think this is fine for now since we'll move everything to R at some point anyway!

Closes #29


Moohan:
@Jennit07 If you manage to get the PPA stuff done (#49) I might piggyback off that code to also include a CIJ_delay count in the indiv file.

Moohan:
@Jennit07 to add some lines to the individual file to create a count of delayed stays in the year.

Jennit07:
As discussed, count for cij_delay added. Will have merge conflicts with ppa count as this was added on same line 75. Still needs copied across all years once confirmed :)

Moohan:
I think this is good to go. Can you merge if you're happy with it too please?

Jennit07 (approved):
New cij_delay variable added to episode file, checked over, added a count for variable to individual file and test variables added to D05X. Ready to merge changes into Sep21 update branch

PR #55, opened 2021-08-05
Add ppa count

Jennit07:
Creates a variable so that preventable admissions and beddays can be counted in the individual file as used in the matrix dashboard


2019-20/D01 Make Individual File.sps
```diff
@@ -72,6 +72,15 @@ Else if cij_pattype_code = 2.
     Compute CIJ_mat = Distinct_CIJ.
 End if.
 
+*Create variables to count preventable admissions and preventable beddays. 
+Do if cij_ppa = 1 and Distinct_CIJ = 1.
+    Compute preventable_admissions = 1.
+    Compute preventable_beddays = yearstay.
```
Moohan, line 78: This will only count days from the first episode

Moohan (changes requested):
Changes as discussed :)

2019-20/D01 Make Individual File.sps
```diff
@@ -72,6 +72,15 @@ Else if cij_pattype_code = 2.
     Compute CIJ_mat = Distinct_CIJ.
 End if.
 
+*Create variables to count preventable admissions and preventable beddays. 
+Do if cij_ppa.
+    If Distinct_CIJ preventable_admissions = 1.
+    Compute preventable_beddays = datediff(Max(!startFY, CIJ_start_date), Min(!endFY, CIJ_end_date), "days").
```
Moohan, line 78: These dates need swapping around as it is producing negative time-lengths at the moment.

Moohan (changes requested):
Can you make the change to the dates please and then just run it to check it looks ok. I just opened the standard SLF ep file manually then renamed `anon_chi` to `chi`, then this code block should be able to run ok.

It looks good other than the negative beddays - no stays over 365 days!

Moohan (changes requested):
Could you add the variables to the final `D07` file, please? And possibly into the test files to track differences between updates.

Same for the Delay variables #53

2019-20/D05X Individual file tests.sps
```diff
@@ -282,6 +283,7 @@ aggregate outfile = Old_Summary
     /Age_mean = mean(age)
     /sex_Males sex_Females = Sum(Male Female)
     /n_NSUs n_Dead = SUM(NSU deceased)
+    /n_preventable_admissions n_preventable_beddays = Sum(preventable_admissions preventable_beddays)
```
Moohan, line 286: Could this be `total_preventable_beddays` please

Jennit07: Sure! That makes more sense :) Just testing the negative beddays code above then ill copy across years if all looks good to you

Jennit07:
Changes copied across years, tested code for negative beddays and everything looks good to me :)

Moohan (approved):
Looks good. I'll merge now.

PR #56, opened 2021-08-06
Add 2021/22 syntax

Moohan:
Mostly copied from 2020/21 syntax with only changes to the macro file. Just asking for a review in case I'm forgetting something?

For HHG and SPARRA, I'm about to add that, so 21/22 will have 'start_fy' only and 20/21 will have start and end.


Jennit07 (approved):
Looks good to me. Only suggestion is to remove CMH and DN syntax to avoid those empty files while processing

Jennit07:
Looks good, happy to merge :)

PR #63, opened 2021-09-08
Fix Care Home duplicates bug

Moohan:
Was previously removing all instances of a reords where it was duplicated, instead of all but one. This is now corrected!


Jennit07:
Code ran through and checked, looks like it is working properly and now removing all duplicates except one. Ready to merge into sept21 branch

PR #71, opened 2021-11-03
Get GP Practice Cluster data from the Open Data platform

Moohan:
Hey, I've ran and checked this and think it's fine. If you're ok with it I'll remove the extra checking code.


Jennit07 (changes requested):
Just checked all the macros, file paths and the R script. All looks good to me and the ordering makes sense. Happy to approve if you want to remove your checking code?

PR #72, opened 2021-11-10
Home care

Moohan:
Hey, just opening this as a  draft for you to have a look at / comment on.

I finally fixed the 'episode count' bit (pretty sure!).

The overall idea is the same as with Care Home i.e. Clean up the data (fix variables etc), then merge the records together on person, start_date, service type and reablement type (where reablement follows the same methodology as for nursing_care).

One extra bit is that we have to sum all the hours for the merged episodes but also keep track o which quarter the hours were submitted in / assigned to.


Moohan:
Need to:

- [ ] apply the fix for old start dates
- [ ] Update news

Jennit07 (approved):
Looks good. Ready to merge into Dec21 update

PR #101, opened 2021-11-12
Add nsu

### LINKED ISSUE #64: Get and add NSU cohort for 20/21
Need to do an extract of existing CHIs from the latest 20/21 file and send this to the CHILLI team with a request to get an NSU extract for that year.

Keep the temp-episode-3 file for 20/21 as we can restart from there when we have the extract which should save quite a bit of time and faff!

Jennit07:
Adds nsu syntax to 2021 C04 and D04. New variable 'gender' has been added to the extract for years 1920 onwards so this has been added to the syntax. This will need updating for older years with NSU available when we receive new NSU extracts for previous years. Closes #64


Moohan (approved):
Looks good

PR #105, opened 2021-11-16
Adding Social Care to 21/22

Moohan:
* Extended the range for valid data to Q1 2021/22
 * Changed the Care Home extract to SQL
 * Added relevant syntax to 2021/22

### DIFF
--- a/2019-20/C01 Make Episode File.sps
+++ b/2019-20/C01 Make Episode File.sps
@@ -36,9 +36,6 @@ add files
     /file = !Year_dir + "care_home_for_source-20" + !FY + ".zsav"
     /By chi.
 
-* All records should be sorted by CHI, if the above fails, remove the "/By chi" and run again then run the below sort.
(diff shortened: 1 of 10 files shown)


Jennit07 (changes requested):
All other tweaks and tidy code looks good to me. In regards to the SQL, i have tested this a few ways: 
-  on its own with an execute (working)
- without an execute and run to the alter type (not working - throws an SQL error?) 
- Tested with the execute and down to the alter type (working)

It looks like it is working now but takes a bit longer to run. I've just requested changes so you can check you are happy with it before merging. But other than that i think it looks good

Moohan:
Great thanks, I suspected the `execute` might be an issue as I tested with it but then removed it before committing which was silly!

1 down - demographics and client still to go...

FYI - I've been using 'squash and merge' so it makes just one commit to the other branch which I think is quite nice.

PR #108, opened 2021-11-18
Change the Social Care demographics extract to SQL

Moohan:
I've ran this all through to the save and it works for me but could you please run and check it all before merging?


Jennit07:
Ran and checked this - all looks good now :)

PR #109, opened 2021-11-19
Changed the client extract to use SQL

Moohan:
Had to use some macro hack to be able to use our macro in the filter, seems to work fine though.
I also added in lost of other variables based on a recent question, I need to speak to Social Care about actually including them though!


Jennit07:
Looks good!

PR #117, opened 2021-12-03
Uplift costs by 1% per year.

### LINKED ISSUE #116: Uplift costs as standard
Costs are not currently uplifted.

Need to build an uplift in for all costs. Use 1% for each year after the available one (can hopefully do this with macros?) e.g. Latest available cost?

Do we possibly apply this in C01 - for PLICS costs, and in the costs lookup for GP OoH, DN and CH?

Could also apply it in each dataset processing code.

Moohan:
Apply an uplift of 1% per year to all costs for all years after the latest year where 'real' costs are available. Closes #116 

Done with the PLICS side of things.

I need to update the cost syntax for:

- [x] Care Home
- [x] GP Out of Hours
- [x] District Nursing

### DIFF
--- a/2014-15/C10 Final tidy for Episode file.sps
+++ b/2014-15/C10 Final tidy for Episode file.sps
@@ -1,6 +1,12 @@
 ﻿* Encoding: UTF-8.
 get file = !Year_dir + "temp-source-episode-file-8-" + !FY + ".zsav".
 
+!create_uplift_var.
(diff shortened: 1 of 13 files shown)


Jennit07 (approved):
Checked through the code and looks good, PLICS costs working. CH, OOH and DN costs ran with the new uplifted costs applied.

PR #119, opened 2021-12-08
Update Individual file for Home Care

Moohan:


2014-15/D01 Make Individual File.sps
```diff
@@ -410,14 +411,19 @@ Else if (recid = "HC").
     Compute HC_episodes = 1.
 
     * Hours count.
-    Compute HC_total_hours = hc_hours.
+    Compute HC_total_hours = hc_hours_q1 + hc_hours_q2 + hc_hours_q3 + hc_hours_q4.
```
Moohan, line 414: Need to replace this bit with `hc_hours_annual` once that's been updated/added in the ep file.

Jennit07 (approved):
Code looks good to me. Think you have got all the variables in the right places and they make sense too!

PR #127, opened 2022-01-18
Updates to the file path functions

Moohan:


Jennit07:
**Notes**

- Fixed typo in cohorts function - glue function was calling Demographic cohort for the` get_service_use_cohorts_path()`
- While testing `get_it_prescribing_path()` function will return error message when applying the parameter "1819" as the current file that exists in the folder has an older IT reference number. Possibly need to make another function similar to `previous_update()` and apply this to the older IT extracts? This will depend on how often we will use the old IT extract numbers whether we need this or not? However the function is working properly and `get_it_prescribing_path()` will return a file path and tell you whether this exists in the directory.
- In `get_practice_details_path()` this references the file` Practice Details.sav `, however in the folder there is a file called `practice_details_Dec_2021.sav`. Should we be referencing this file instead? if that is the case, `get_practice_details_path()` will need to call the function` latest_update()`

Moohan (commented):
Great spot, thanks :)

Moohan:
> **Notes**
> 
> * Fixed typo in cohorts function - glue function was calling Demographic cohort for the` get_service_use_cohorts_path()`
> * While testing `get_it_prescribing_path()` function will return error message when applying the parameter "1819" as the current file that exists in the folder has an older IT reference number. Possibly need to make another function similar to `previous_update()` and apply this to the older IT extracts? This will depend on how often we will use the old IT extract numbers whether we need this or not? However the function is working properly and `get_it_prescribing_path()` will return a file path and tell you whether this exists in the directory.
> * In `get_practice_details_path()` this references the file`Practice Details.sav`, however in the folder there is a file called `practice_details_Dec_2021.sav`. Should we be referencing this file instead? if that is the case, `get_practice_details_path()` will need to call the function` latest_update()`

Happy to make those changes.

- [ ] Update practice details to point to new file(s)
- [ ] Add param to IT functions to allow specifying an alternative ref number

Moohan:
Changes made. @Jennit07 can you review again, please?

Jennit07:
Looks good :). As discussed on call, changes to: 

- [ ] regular expression
- [ ] Adding comments 
- [ ] Pull out substring
- [ ] Rename IT extracts to "tasknumber_PIS_year" - When requesting IT extracts to ask IT to rename accordingly

Jennit07:
> Looks good :). As discussed on call, changes to:
> 
> * [ ]  regular expression
> * [ ]  Adding comments
> * [ ]  Pull out substring
> * [ ]  Rename IT extracts to "tasknumber_PIS_year" - When requesting IT extracts to ask IT to rename accordingly

I dont think removing the substring works, it comes up with error cannot find the file. This is with me applying "1819" and "2018" as a parameter and giving an older reference number. However leaving substring and applying both of these as a parameter allows the correct file to be identified

Moohan:
Have made those changes. I think we should handle the file name changes as a separate thing as that needs some other bits in place as well.

I'll also do the functions to swap between 'standard' (1718) and 'alternate' (2017) fy in a separate PR too.

Jennit07:
I agree with splitting out the other two PR's. I tested this yesterday and thought it wasn't working but realised i was using the function wrong. Tried again today and the function changes FY to CY and pulls through the correct extract path. I tried this a few ways inputting alternative task numbers for years without the current task number and it works okay. Happy to merge :)

Jennit07 (approved):

PR #130, opened 2022-01-26
Fy formats 

### LINKED ISSUE #129: Create functions to convert between Financial Year types
We mostly use FY in the form of 1718 (2017/18) but sometimes we also need it as 2017. 

Make functions that swap between the two formats.

catherineholland1:
closes #129 
Functions created to convert normal financial year form into the alternate form and the reverse. 

i.e 1718 to 2017 and the reverse


Source Linkage Files.Rproj
```diff
@@ -0,0 +1,13 @@
+Version: 1.0
```
Moohan, line 1: I'm not sure why this file was created. Could you just delete it please? or revert the commit which added it (would be better but slightly more work!)

R/convert_year_types.R
```diff
@@ -0,0 +1,30 @@
...
+#' @param year vector of years in the form '2017'
+#'
+#' @return a vector of years in the normal form '1718'
+#' @export
+#'
+#' @examples
+#' convert_year_to_fyyear(c("1718", "1819"))
+convert_year_to_fyyear <- function(year) {
+  fyyear <- paste0(substr(year, 3, 4), as.numeric(substr(year, 3, 4)) + 1)
+  return(as.numeric(fyyear))
```
Moohan, line 29: I think we probably want to return a character?

At least it should be consistent i.e. both functions take and return the same type (either integer or character)

tests/testthat/test-convert_year_types.R
```diff
@@ -0,0 +1,61 @@
+test_that("Can convert a year from normal financial year to the alternate form", {
```
Moohan, line 1: Tests are good.
No need to test all years, one or two is fine. Testing edge cases is good though i.e. earliest file = 2011/12.  Also, test incorrect inputs e.g. what if we run `convert_fyyear_to_year("2017)`?

Moohan (changes requested):
Looks good. There's a couple of changes suggested :)

Source Linkage Files.Rproj
catherineholland1: commit reverted

R/convert_year_types.R
catherineholland1: removed as.numeric() from function to return a character variable instead

tests/testthat/test-convert_year_types.R
catherineholland1: removed unnecessary year tests

Would I include an error message to make sure that the correct form has been inputted?

Moohan:
/style

Just testing this GH action ^^ in theory it should run styler and re-commit 😉

catherineholland1 (commented):
all changes complete

R/convert_year_types.R
```diff
@@ -0,0 +1,46 @@
...
+#'
+#' @examples
+#' fyyears <- c("1718", "1819")
+#' convert_fyyear_to_year(fyyears)
+#' [1] "2017" "2018"
+convert_fyyear_to_year <- function(fyyear) {
+  for (i in 1:length(fyyear)) {
+    if (substr(fyyear[i], 1, 2) > substr(fyyear[i], 3, 4)) {
+      stop ("Year has been entered in the wrong format, try again using form `1718` or use function `convert_year_to_fyyear` to convert to the financial year form.")
+  }}
```
Moohan, line 18: You can get rid of the whole loop construct with `any()` i.e. if any of the supplied values are bad throw an error.
```suggestion
    if (any(substr(fyyear[i], 1, 2) > substr(fyyear[i], 3, 4))) {
      stop("Year has been entered in the wrong format, try again using form `1718` or use function `convert_year_to_fyyear` to convert to the financial year form.")
  }
```

We could also make the check more strict than just greater than, i.e. second part must be first part + 1:
```suggestion
  first_part <- as.integer(substr(fyyear, 1, 2))
  second_part <- as.integer(substr(fyyear, 3, 4))
  if (any(first_part != second_part + 1L)) {
    stop("Year has been entered in the wrong format, try again using form `1718` or use function `convert_year_to_fyyear` to convert to the financial year form.")
  }
```

R/convert_year_types.R
```diff
@@ -0,0 +1,46 @@
...
+#'
+#' @examples
+#' years <- c("2017", "2018")
+#' convert_year_to_fyyear(years)
+#' [1] "1718" "1819"
+convert_year_to_fyyear <- function(year) {
+  for (i in 1:length(year)){
+    if (substr(year[i], 1, 2) != "20") {
+      stop ("Year has been entered in the wrong format, try again using form `2017` or use function `convert_fyyear_to_year` to convert to alternate year form.")
+  }}
```
Moohan, line 43: Replace the loop with `any()`

R/convert_year_types.R
```diff
@@ -0,0 +1,46 @@
...
+#'
+#' @param fyyear vector of financial years in the form '1718'
+#'
+#' @return a vector of years in the alternate form '2017'
+#' @export
+#'
+#' @examples
+#' fyyears <- c("1718", "1819")
+#' convert_fyyear_to_year(fyyears)
+#' [1] "2017" "2018"
```
Moohan, line 13: Remove this line please, the examples are run and the output is generated when you run `devtools::document`. So this isn't needed (and possibly breaks the documentation output)

R/convert_year_types.R
```diff
@@ -0,0 +1,46 @@
...
+#'
+#' @param year vector of years in the form '2017'
+#'
+#' @return a vector of years in the normal financial year form '1718'
+#' @export
+#'
+#' @examples
+#' years <- c("2017", "2018")
+#' convert_year_to_fyyear(years)
+#' [1] "1718" "1819"
```
Moohan, line 38: Remove this line as above, please.

Moohan (changes requested):
Almost there, a couple more changes though please:

- [ ] Replace loops with use of `any()`
- [ ] Remove 'output' lines from documentation comments
- [ ] Add a test to check for the error e.g. `expect_error(convert_fyyear_to_year("2017"), "Year has been entered in the wrong format")`
- [ ] Run [styler](https://styler.r-lib.org/index.html) - Just install the package then use the addin 😉

R/convert_year_types.R
catherineholland1: Just the [1] line or calling the function as well?

R/convert_year_types.R
catherineholland1: When I include your above code the test fails? but unsure how as all of the test years second part are +1 from the first

R/convert_year_types.R
Moohan: Yep just the 'output' `[1]` line as you've done 👍

R/convert_year_types.R
Moohan: Hmm, I'll have a quick look but I'm happy with it how it is too, I'll either push a change or approve shortly!

Moohan (approved):

Moohan:
All changes approved. @catherineholland1 I'll leave this to you to merge it 😉 I would usually do a 'squash and merge' (see the drop-down arrow) for this type of change as ultimately it's quite a simple addition. Squash and merge will combine all the commits into one (it combines all the messages too and gives you chance to edit it).

PR #135, opened 2022-02-02
#88 LTC Extract Processing Conversion

### LINKED ISSUE #88: Convert LTC extract processing

catherineholland1:
Closes #88

B12 LTC extract processing conversion to R


LTC_extracts_DRAFT.R
```diff
@@ -0,0 +1,156 @@
...
+
+
+# combine ltc and ltc_date
+LTC <- cbind(data[, 1:2], ltc, ltc_date)
+
+
+
+## sort data by chi ##
+LTC <- LTC %>%
+  dplyr::arrange(chi)
```
Jennit07, line 101: This is good - you can put this line in a pipe after line 51: rename_with(.cols = ends_with("flag"), .fn = ~str_remove(.x, "_date_flag"))

LTC_extracts_DRAFT.R
```diff
@@ -0,0 +1,156 @@
...
+
+
+
+
+## Save out to Year folder - /conf/hscdiip/SLF_Extracts/LTCs ##
+# fs::path(get_slf_dir(), "LTCs", "<filename>")
```
Jennit07, line 107: Remember to remove lines that are unrelated - Okay for now while you are testing :)

LTC_extracts_DRAFT.R
```diff
@@ -0,0 +1,156 @@
...
+# haven
+haven::write_sav(LTC, paste0("LTC_patient_reference_file-20", FY),
+  path = "/conf/hscdiip/SLF_Extracts/LTCs", compress = TRUE
+)
+
+readr::write_rds(LTC, paste0("LTC_patient_reference_file-20", FY),
+  path = "/conf/hscdiip/SLF_Extracts/LTCs", compress = "gz"
+)
+
+
```
Jennit07, line 118: You can write "End of script here" and also use style r on your script to tidy it up :)

Jennit07 (changes requested):
Good work - you have a really good understanding of R and what we are trying to do in this script. Some parts of your code are not working quite as we would expect so I've tried to give a lot of detail through comments and suggestions throughout. If there is anything you are not sure about, let me know and i'd be happy to help explain things. A summary of the changes: 

- [ ] Add comments to top of script 
- [ ] Read in CSV with correct column types 
- [ ] Rename variables separately 
- [ ] Create flags using tidyverse code instead of base R in one dataframe 
- [ ] Decide whether to use a function for whole script or leave as it is

Hope you have a nice weekend :)

LTC_extracts_DRAFT.R
```diff
@@ -0,0 +1,156 @@
...
+  for (j in 1:ncol(ltc)) {
+    ifelse(is.na(ltc_date[i, j] == TRUE | ltc_date[i, j] > fyyear_date | ltc_date[i, j] == fyyear_date),
+      ltc[i, j] <- 0, ltc[i, j] <- 1
+    )
+  }
+}
+
+
+# combine ltc and ltc_date
+LTC <- cbind(data[, 1:2], ltc, ltc_date)
```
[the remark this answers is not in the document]
Moohan: Just to clarify the logic here, the flag should be:
  1 - if there is a date that falls before the end of the financial year.
  0 - otherwise i.e. no date (NA) or a date after the end of the financial year.
  
I think Catherine's logic was flagging any non-missing date as 1 (even if it was 'in the future') but Jen's logic will only flag as 1 if it falls within the year.

I think this would be correct... Need some tests 😝
```suggestion
## Create LTC flags 1/0 ##

# Should be a function which takes fyyear 
# and returns the correct date.
end_fy <- dmy("01/04/2017")

LTC <- data %>%
  mutate(across(ends_with("date"), list(flag = ~if_else(is.na(.x) | .x > end_fy, 0, 1)))) %>%
  rename_with(.cols = ends_with("flag"), .fn = ~str_remove(.x, "_date_flag"))
```

LTC_extracts_DRAFT.R
```diff
@@ -0,0 +1,156 @@
+
+## financial year in question ##
+FY <- 1718
+year <- 2017
+
+
+# output1718 <- haven::read_sav("/conf/hscdiip/SLF_Extracts/LTCs/LTCs_patient_reference_file-201718.zsav")
```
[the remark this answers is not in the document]
catherineholland1: comment removed

LTC_extracts_DRAFT.R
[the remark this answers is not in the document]
Jennit07: Thanks James, for some reason i thought this wasn't working when i was testing this, but i can see how my change wasn't right either. Makes more sense with your logic :)

LTC_extracts_DRAFT.R
```diff
@@ -0,0 +1,156 @@
+
```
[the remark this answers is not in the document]
catherineholland1: comments added at beginning of the script

LTC_extracts_DRAFT.R
```diff
@@ -0,0 +1,156 @@
...
+## financial year in question ##
+FY <- 1718
+year <- 2017
```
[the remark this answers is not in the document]
catherineholland1: changed year to the function: `convert_fyyear_to_year()`

LTC_extracts_DRAFT.R
```diff
@@ -0,0 +1,156 @@
...
+
+
+# output1718 <- haven::read_sav("/conf/hscdiip/SLF_Extracts/LTCs/LTCs_patient_reference_file-201718.zsav")
+
+
+
+## Read data ##
+it_extract_ref()
```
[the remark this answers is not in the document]
catherineholland1: removed

LTC_extracts_DRAFT.R
```diff
@@ -0,0 +1,156 @@
...
+  "chd_date",
+  "hefailure_date",
+  "ms_date",
+  "parkinsons_date",
+  "refailure_date",
+  "congen_date",
+  "bloodbfo_date",
+  "endomet_date",
+  "digestive_date"
+)
```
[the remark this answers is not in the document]
catherineholland1: changed way data read in, as above for reproducibility

LTC_extracts_DRAFT.R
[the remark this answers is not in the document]
catherineholland1: edited the flag code using %>% with the code above

LTC_extracts_DRAFT.R
catherineholland1: added into earlier pipe

LTC_extracts_DRAFT.R
catherineholland1: end of script included

catherineholland1 (commented):
comments implemented

LTC_extracts_DRAFT.R
```diff
@@ -0,0 +1,137 @@
...
+    `HEART_DISEASE_DIAG_DATE` = col_date(format = "%Y/%m/%d %T"),
+    `MULT_SCLEROSIS_DIAG_DATE` = col_date(format = "%Y/%m/%d %T"),
+    `PARKINSONS_DIAG_DATE` = col_date(format = "%Y/%m/%d %T"),
+    `RENAL_FAILURE_DIAG_DATE` = col_date(format = "%Y/%m/%d %T"),
+    `CONGENITAL_PROB_DIAG_DATE` = col_date(format = "%Y/%m/%d %T"),
+    `BLOOD_AND_BFO_DIAG_DATE` = col_date(format = "%Y/%m/%d %T"),
+    `OTH_DIS_END_MET_DIAG_DATE` = col_date(format = "%Y/%m/%d %T"),
+    `OTH_DIS_DIG_SYS_DIAG_DATE` = col_date(format = "%Y/%m/%d %T")
+  )
+)
```
Jennit07: Sorry the col date format was working for my acute file but this is different in the LTCs - it uses %d-%m-%Y instead. I'll commit the change

LTC_extracts_DRAFT.R
```diff
@@ -0,0 +1,137 @@
...
+    hefailure = NA,
+    ms = NA,
+    parkinsons = NA,
+    refailure = NA,
+    congen = NA,
+    bloodbfo = NA,
+    endomet = NA,
+    digestive = NA
+  )
+
```
Jennit07: I've removed this and commited the change as the code on line 114 will create the columns so no need to specify here

LTC_extracts_DRAFT.R
```diff
@@ -0,0 +1,137 @@
...
+
+
+# Set flags to 1 or 0 based on FY
+# then sort by chi
+end_fy <- lubridate::dmy(paste0("01-04-", as.numeric(substr(year, 3, 4)) + 1))
```
Jennit07: I think this is fine, but @Moohan should this be end of FY i.e 31-03-2018? i would think 01-04-2017 is the startFY - Just want to check this

Jennit07 (changes requested):

LTC_extracts_DRAFT.R
Moohan: Yep, I agree end_fy should be 31/03/XXXX we then need to be careful with `>` vs `>=` etc.

LTC_extracts_DRAFT.R
```diff
@@ -0,0 +1,103 @@
...
+## packages ##
+
+library(dplyr)
+library(stringr)
+library(readr)
+
+## financial year in question ##
+FY <- 1718
+year <- convert_fyyear_to_year(FY)
+end_fy <- lubridate::dmy(paste0("01-04-", as.numeric(substr(year, 3, 4)) + 1))
```
Moohan, line 24: We don't need lubridate here, and `dmy` is usually slower because it's 'clever'. We can just specify the date in a more standard format(i.e. YYYY-MM-DD)

```suggestion
FY <- 1718
end_fy <- as.Date(paste0(convert_fyyear_to_year(FY), "-03-31")
```

Moohan (approved):
Just a minor suggestion from me but that could also be picked up when doing #137

LTC_extracts_DRAFT.R
catherineholland1: changed endfy to "31/03/XXX" 

included in the flag code an extra constraint: ".. | .x == end_fy, ... "

LTC_extracts_DRAFT.R
catherineholland1: lubricate removed

added +1 to first part of the date as we are looking at the end of fy year so should be "2017 + 1"? please check my logic is correct @Moohan

Moohan: Yep you're correct, that looks good now

Jennit07 (approved):
Looks good - Ready to merge

PR #143, opened 2022-02-14
Rename IT extracts (SPSS)

### LINKED ISSUE #128: Rename IT extracts
Opening an Issue to remember to change IT extracts name to the format "tasknumber_PIS_year" - will need to make this macro change in SPSS Syntax before running the march-22-update.

Jennit07:
IT extracts now in the format tasknumber_PIS_year or tasknumber_LTC/Deaths. This has been updated in the IT extract folder and also in syntax. This should be in line with the request that new extracts are also renamed. Closes #128

### DIFF
--- a/2014-15/A01b Set up 1415 Macros.sps
+++ b/2014-15/A01b Set up 1415 Macros.sps
@@ -24,10 +24,3 @@ Define !FY()
 Define !NextFY ()
     "1516"
 !EndDefine.
-
-*This related to the file in the IT extracts directory and should unzip this in B06 Process PIS extract.
-*Change this to the relevant number specific to FY.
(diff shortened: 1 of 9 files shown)


Moohan (commented):
Looks good. I'm not going to approve just yet as I think we need to wait and see what IT / BI actually do with the requests...

Would you mind adding an equivalent PR for the R versions, please?

Jennit07:
> Looks good. I'm not going to approve just yet as I think we need to wait and see what IT / BI actually do with the requests...
> 
> Would you mind adding an equivalent PR for the R versions, please?

Thanks James, no problem i'll add the new PR :)

Moohan (approved):

PR #145, opened 2022-02-14
New function to find the latest file given a some search parameters

Moohan:
Added this the other day as I had written it for some other bit of work. Had issues with git so just pushing now!

I can see this being useful for grabbing the lookups (for example) - you point it to the right folder and 'type' of file and it will return the latest version that meets that criteria.

In the doc example, I've used the SPD, which would mean we would never have to check the name of the latest SPD!


Jennit07 (commented):
This is working, however i have to specify spd_dir as an object for it to work. Is the idea to incorporate this into read_lookups functions for selecting the spd? If so, i think  this would work with a few tweaks to the read_lookups functions

catherineholland1:
Looks good but I got same as Jen for spd_dir - if this is incorporated into other functions where this is defined would work.

catherineholland1 (changes requested):
Looks good but I got same as Jen for spd_dir - if this is incorporated into other functions where this is defined would work.

R/find_latest_file.R
```diff
@@ -13,9 +13,9 @@
 #' @export
 #'
 #' @examples
-#' find_latest_file(spd_dir, regexp = "Scottish_Postcode_Directory_.+?\\.rds", recurse = TRUE)
-find_latest_file <- function(dir, ...) {
-  fs::dir_info(path = dir, type = "file", ...) %>%
+#' find_latest_file(get_lookups_dir(), regexp = "Scottish_Postcode_Directory_.+?\\.rds")
+find_latest_file <- function(dir, ..., recurse = TRUE) {
+  fs::dir_info(path = dir, type = "file", ..., recurse = recurse) %>%
```
Moohan, line 18: Discussed this with Jen.

The example was my mistake and is now fixed!

We also thought having `recurse = TRUE` be the default made sense, so by default it will look in all sub-folders.

PR #149, opened 2022-02-21
#77 Convert Postcode lookup

### LINKED ISSUE #77: Convert Postcode Lookup processing

catherineholland1:
Closes #77 conversion 

the file patch functions are included until the functions have been updated in the package.

a new function `ca_to_lca` is included in this PR too - I have created and tested the function too. 

I couldn't figure out how to use the regex matching in the last selection of variables/how to get a new regex for the quintile variables, or are they the same just replace the `sc_decile`?


R/convert_ca_to_lca.R
```diff
@@ -0,0 +1,57 @@
...
+
+
+
+# library(phsopendata)
+
+# ca_codes <- readr::read_csv(
+#    "https://www.opendata.nhs.scot/dataset/9f942fdb-e59e-44f5-b534-d6e17229cc7b/resource/967937c4-8d67-4f39-974f-fd58c4acfda5/download/ca11_ca19.csv")
+
+# ca <- phsopendata::opendata_get_resource(res_id = "967937c4-8d67-4f39-974f-fd58c4acfda5")
+## phs package not working
```
Moohan, line 57: Remove commented code, please.
```suggestion
```

R/convert_ca_to_lca.R
```diff
@@ -0,0 +1,57 @@
...
+    ca == "S12000038" ~ "26", # Renfrewshire
+    ca == "S12000027" ~ "27", # Shetland
+    ca == "S12000028" ~ "28", # South Ayrshire
+    ca == "S12000029" ~ "29", # South Lan
+    ca == "S12000030" ~ "30", # Stirling
+    ca == "S12000040" ~ "31", # West Lothian
+    ca == "S12000013" ~ "32" # Na h-Eileanan Siar
+  )
+  return(lca)
+}
```
Moohan, line 47: This is currently using CA2019 codes, but could it please be expanded to use CA2011 and CA2018 too? This will make it more generally useful in some other places.

All required changes made in the suggestion:

```suggestion
ca_to_lca <- function(ca) {
  lca <- dplyr::case_when(
    ca == "S12000033" ~ "01", # Aberdeen City
    ca == "S12000034" ~ "02", # Aberdeenshire
    ca == "S12000041" ~ "03", # Angus
    ca == "S12000035" ~ "04", # Argyll and Bute
    ca == "S12000026" ~ "05", # Scottish Borders
    ca == "S12000005" ~ "06", # Clackmannanshire
    ca == "S12000039" ~ "07", # West Dun
    ca == "S12000006" ~ "08", # Dumfies and Galloway
    ca == "S12000042" ~ "09", # Dundee City
    ca == "S12000008" ~ "10", # East Ayrshire
    ca == "S12000045" ~ "11", # East Dun
    ca == "S12000010" ~ "12", # East Lothian
    ca == "S12000011" ~ "13", # East Ren
    ca == "S12000036" ~ "14", # City of Edinburgh
    ca == "S12000014" ~ "15", # Falkirk
    ca %in% c("S12000015", "S12000047") ~ "16", # Fife
    ca %in% c("S12000046", "S12000049") ~ "17", # Glasgow City
    ca == "S12000017" ~ "18", # Highland
    ca == "S12000018" ~ "19", # Inverclyde
    ca == "S12000019" ~ "20", # Midlothian
    ca == "S12000020" ~ "21", # Moray
    ca == "S12000021" ~ "22", # North Ayrshire
    ca %in% c("S12000044", "S12000050") ~ "23", # North Lan
    ca == "S12000023" ~ "24", # Orkney
    ca %in% c("S12000024", "S12000048") ~ "25", # P & K
    ca == "S12000038" ~ "26", # Renfrewshire
    ca == "S12000027" ~ "27", # Shetland
    ca == "S12000028" ~ "28", # South Ayrshire
    ca == "S12000029" ~ "29", # South Lan
    ca == "S12000030" ~ "30", # Stirling
    ca == "S12000040" ~ "31", # West Lothian
    ca == "S12000013" ~ "32" # Na h-Eileanan Siar
  )
  return(lca)
}
```

postcode_lookup.R
```diff
@@ -0,0 +1,149 @@
...
+read_simd_file <- function(file) {
+  simd_path <- fs::path(get_lookups_dir(), "Deprivation", file)
+  # If given a sav extension (or other), swap it for rds
+  simd_path <- fs::path_ext_set(simd_path, "rds")
+  # Check if the file exists and we can read it
+  if (!fs::file_access(simd_path, "read")) {
+    rlang::abort(message = "Couldn't read the simd file")
+  }
+  return(simd_path)
+}
```
Moohan, line 31: Move this function to its own script, please. I think called `get_simd_path`, since it actually returns the path, it doesn't read the file. I think the SPD function is badly named but we can deal with that separately!

postcode_lookup.R
```diff
@@ -0,0 +1,149 @@
...
+  locality_path <-
+    fs::path(get_lookups_dir(), "Geography", "HSCP Locality", file)
+  # If given a sav extension (or other), swap it for rds
+  locality_path <- fs::path_ext_set(locality_path, "rds")
+  # Check if the file exists and we can read it
+  if (!fs::file_access(locality_path, "read")) {
+    rlang::abort(message = "Couldn't read the locality file")
+  }
+  return(locality_path)
+}
```
Moohan, line 45: To its own script please, as above for naming, maybe: `get_locality_lookup_path`?

postcode_lookup.R
```diff
@@ -0,0 +1,149 @@
...
+
+# arrange pc lookups based on pc7
+pc_file <-
+  pc_file %>%
+  arrange(pc7)
+
+# arrange simd based on pc7
+simd_file <-
+  simd_file %>%
+  arrange(pc7)
```
Moohan, line 59: Sorting is not required for R (at least not to match). It's something we have to do in SPSS before matching files together.

postcode_lookup.R
```diff
@@ -0,0 +1,149 @@
...
+
+# join data together by pc7
+data <-
+  pc_file %>%
+  left_join(simd_file, by = "pc7") %>%
+  # sort by DataZone2011
+  rename(datazone2011 = "datazone2011.x") %>%
+  arrange(datazone2011)
```
Moohan, line 67: Assumedly this is because both files have a datazone2011 variable. I think a better way would be to have dropped it from one of the files earlier (probably makes the most sense to drop it from the SIMD file). Alternatively, you could match by pc7 and datazone2011 (they will be 1:1 so will do the same thing but not end up duplicating the column).
```suggestion
data <-
  pc_file %>%
  left_join(simd_file, by = c("pc7", "datazone2011")) 
```

postcode_lookup.R
```diff
@@ -0,0 +1,149 @@
...
+
+
+## save file ##
+
+data <-
+  data %>%
+  # rename pc7
+  rename(
+    postcode = "pc7",
+    simd2020v2_rank = "simd2020v2_rank.x"
```
Moohan, line 97: This is caused by the same issue as above.

I'm thinking it would be better to only keep the variables we want from each file before the match.

postcode_lookup.R
```diff
@@ -0,0 +1,149 @@
...
+    simd2020v2_sc_quintile,
+    simd2020v2_hb2019_decile,
+    simd2020v2_hb2019_quintile,
+    simd2020v2_hscp2019_decile,
+    simd2020v2_hscp2019_quintile,
+    ur8_2016,
+    ur6_2016,
+    ur3_2016,
+    ur2_2016
+  )
```
Moohan, line 114: Is this required as there is another select just below this?

tests/testthat/test-convert_ca_to_lca.R
```diff
@@ -0,0 +1,17 @@
...
+
+  expect_equal(
+    ca_to_lca(ca),
+    c(
+      "01",
+      "17",
+      "25"
+    )
+  )
+})
```
Moohan, line 17: Can you add an `NA` and a 'bad' (not matching) code to test too, please?

tests/testthat/test-convert_ca_to_lca.R
```diff
@@ -0,0 +1,17 @@
+library(testthat)
```
Moohan, line 1: Not needed.
```suggestion
```

postcode_lookup.R
```diff
@@ -0,0 +1,149 @@
...
```
Moohan, line 142: Examples of using matches:
```suggestion
  select(
    postcode,
    hb2018,
    hscp2018,
    ca2018,
    lca,
    locality,
    datazone2011,
    hb2019,
    ca2019,
    hscp2019,
    matches("simd\\d{4}.?.?_rank"),
    matches("simd\\d{4}.?.?_sc_decile"),
    matches("simd\\d{4}.?.?_sc_quintile"),
    matches("simd\\d{4}.?.?_hb\\d{4}_decile"),
    matches("simd\\d{4}.?.?_hb\\d{4}_quintile"),
    matches("simd\\d{4}.?.?_hscp\\d{4}_decile"),
    matches("simd\\d{4}.?.?_hscp\\d{4}_quintile"),
    ur8_2016,
    ur6_2016,
    ur3_2016,
    ur2_2016
  )
```
This should mean that we don't have to change the code when the variable names change. e.g. in future, they will be something like `simd2024_rank` and `simd2024_hscp2025_quintile` etc.

We can probably apply this to the other variables too, e.g. `hb2018` and `hb2019` can just be `matches("hb\\d{4}")` where `\\d` means any number (0-9) and `{4}` means the previous thing 4 times, so 'any 4 numbers'.

Moohan (changes requested):

postcode_lookup.R
catherineholland1: These functions exist but needed a change for it to work - @Jennit07 was on this edit - the functions were included until this was updated. Would you want this name changed on that function?

postcode_lookup.R
catherineholland1: edited code so when read is read in it only selects certain variables

postcode_lookup.R
catherineholland1: removed

postcode_lookup.R
catherineholland1: included for hb, hscp and ca too

Moohan (approved):
Think I'm happy now.

@catherineholland1 sorry I just made some changes myself as they didn't break down into nice chunks. Can you check and if you're still happy / don't spot a new issue pass to Jen please?

catherineholland1:
> Think I'm happy now.
> 
> @catherineholland1 sorry I just made some changes myself as they didn't break down into nice chunks. Can you check and if you're still happy / don't spot a new issue pass to Jen please?

catherineholland1 (commented):
All good from my side

Jennit07 (approved):
This looks really good. My only comment is that in the outfile there are a few more variables in the R output than the current SPSS output and they are not in the same order as the SPSS version. But i'm not too sure how important this is for going forward. The variables in question are a bit older and not sure they are needed which are picked up by the regex. Here is a list of them which appear in the R output and not the SPSS output:
- datazone2001
- hb2014
- hb2006
- hb1995
- hscp2016
- ca2011

My thoughts with this were if we do not need the variables it may reduce the file size.

Happy to approve this pending my query :)

Jennit07:
### File sizes 

R versions:
- Including old variables (1,662KB)
- Excluding old variables (1,448KB)

SPSS versions (This varies from existing lookup in the folder):
- Including old variables (1,356KB)
- Excluding old variables (1,141KB) 

Small differences to be expected but good to compare and think about these things going forward :)

Jennit07 (approved):
All Looks good. New regex checked against the file and now drops old variables. Happy to merge

PR #152, opened 2022-02-24
Process the homelessness extract

### LINKED ISSUE #86: Convert Homelessness extract processing

Moohan:
At this point, it matches the SPSS (as far as I've tested it). I now need to incorporate my other code for filtering duplicates...

Closes #86


Moohan:
Had to rush this out as we needed the filtering for this update, with hindsight I could have saved the completeness figures out and used them in the SPSS code. 

I checked the output against the SPSS output using waldo and it matches including SPSS types etc. so have used it in the update and it seems to work fine.

catherineholland1 (changes requested):
Looks good, just a couple of little things:

- [ ] - I needed `readr` to be loaded for col_date to work, unsure if this will be the case anyway or should include library(readr) somewhere? 

- [ ] - I get the error: _The file homelessness_for_source-201718.rds exists in /conf/sourcedev/Source_Linkage_File_Updates/1718 but is not writeable_ when _map(convert_year_to_fyyear(2017:2021), process_homelessness_extract)_ is run. This could just my access perhaps?

- [ ] should some other text be included after this: _\item{extracts_dir}{(optional) \code{TRUE}/\code{FALSE}, whether to}}_ in _man/get_year_dir.Rd_

Moohan:
> * [ ]  - I needed `readr` to be loaded for col_date to work, unsure if this will be the case anyway or should include library(readr) somewhere?

This should be solved if you load the package first as I've included them as an `import_from`. I've included the line `library(createslf)` in the script but for that to work, you need to have installed the package (in the build menu at the top, click install and restart, or just ctrl + shift + B). Alternatively, you could do load all (`devtools::load_all`) which does pretty much the same thing i.e. makes the package available for use.


> * [ ]  - I get the error: _The file homelessness_for_source-201718.rds exists in /conf/sourcedev/Source_Linkage_File_Updates/1718 but is not writeable_ when _map(convert_year_to_fyyear(2017:2021), process_homelessness_extract)_ is run. This could just my access perhaps?

I'll have a look, I think you should be able to overwrite it... Not sure if this is to do with code or permissions!

catherineholland1 (dismissed):

Jennit07:
> > * [ ]   - I needed `readr` to be loaded for col_date to work, unsure if this will be the case anyway or should include library(readr) somewhere?
> 
> This should be solved if you load the package first as I've included them as an `import_from`. I've included the line `library(createslf)` in the script but for that to work, you need to have installed the package (in the build menu at the top, click install and restart, or just ctrl + shift + B). Alternatively, you could do load all (`devtools::load_all`) which does pretty much the same thing i.e. makes the package available for use.
> 
> > * [ ]   - I get the error: _The file homelessness_for_source-201718.rds exists in /conf/sourcedev/Source_Linkage_File_Updates/1718 but is not writeable_ when _map(convert_year_to_fyyear(2017:2021), process_homelessness_extract)_ is run. This could just my access perhaps?
> 
> I'll have a look, I think you should be able to overwrite it... Not sure if this is to do with code or permissions!

I tested the code against 1819 and because we are in the middle of an update there is a file in the folder. I didnt want to overwrite this incase so tried a test file which worked okay for me. I got the message 'The file homelessness_for_source-201819test.rds did not exist in /conf/sourcedev/Source_Linkage_File_Updates/1819, it has now been created.' as expected

Jennit07 (approved):
This looks really good, i spent some time going through testing/understanding the code and looks like it is all working as expected for me. Outputs match and variables look in the right place etc. I know @Moohan you wanted to make some further changes to the filtering. Happy to check this over again once it's ready but for now i will approve :)

PR #156, opened 2022-03-01
Amend test scripts

### LINKED ISSUE #153: Amend test scripts
Just realised i didnt put an option to save an output for test comparison scripts. A line of code will need to be added for the following scripts in the production folder : 

- [x] Lookup-01b-Postcode-tests
- [x] Lookup-02b-GP-practice-tests
- [x] Lookup-04-CHI-Deceased-tests
- [x] Lookup-05-Social_Care_Demographics-tests

Small bug but will be useful for looking at the outputs in the same format as existing tests

Jennit07:
Minor change to add an option to save an output. This will eventually be put in an excel workbook with tabs for each test. This also includes a count for each row in the sc demographic tests. Close #153


Production_scripts/Lookup-01b-Postcode-tests.R
```diff
@@ -19,5 +19,12 @@ old_tests <- produce_slf_postcode_tests(haven::read_sav(get_slf_postcode_path(up
 # create tests
 comparison <- produce_test_comparison(old_tests, new_tests)
 
+# Save output as zsav for now
+# Eventually change this to rds when we have more R scripts
+haven::write_sav(comparison,
+                 path = glue::glue("/conf/hscdiip/SLF_Extracts/Lookups/source_postcode_lookup_{latest_update()}_tests.zsav",
```
Moohan, line 25: Use `get_slf_dir` function here (and in others).

Moohan (changes requested):
Looks ok, other than resolving Catherine's issue.

We did discuss outputting the test data to sheets in an excel workbook for easier viewing. Could this be done on this PR? Can be done later if not.

Jennit07:
> In `Lookup-01b-Postcode-tests.R` and `Lookup-02b-GP-practice-test.R` getting an error:
> 
> _> new_tests <- produce_slf_postcode_tests(haven::read_sav(get_slf_postcode_path())) Error in `mutate_cols()`: ! Problem with `mutate()` column `NHS_Ayrshire_and_Arran`. ℹ `NHS_Ayrshire_and_Arran = if_else((~.data$HB2019) == "S08000015", 1, 0)`. x Column `HB2019` not found in `.data`. Caused by error in `.data$HB2019`: ! Column `HB2019` not found in `.data`. Run `rlang::last_error()` to see where the error occurred._
> 
> only for the new_tests for both. I had `load_all()` before.
> 
> I think the issue is in `produce_slf_postcode_tests.R` as when the data is run `HB2019` etc.. reads in as lower case `hb2019` etc whereas in the previous update the variables are upper case. This is the same in `produce_slf_gpprac_tests.R`
> 
> This issue does not occur in either of the other scripts.
> 
> Is there someway to incorporate both cases (upper&lower) in the script to account for this?
> 
> The rest of scripts all look great and running fine for me!

Thanks, good spot! This was working before but i have amended this to be lower case. I think this happened because of the new scripts produced in R having lower case and the SPSS old version had a mixture of upper and lower case. This may need amended if there are problems in the future. But hopefully we can make all variables in R lowercase going forward.

Production_scripts/Lookup-01b-Postcode-tests.R
Jennit07: I have updated the zsav file path but also added in the option to save tests as an xlsx workbook. This may need some changes but the idea is to load the workbook, add a new sheet, write data and save out.

catherineholland1 (approved):
all looks good Jen!

Production_scripts/Lookup-01b-Postcode-tests.R
```diff
@@ -11,13 +11,63 @@
...
+  comparison
+)
+# save output
+saveWorkbook(wb,
+  fs::path(
+    get_slf_dir(), "Tests",
+    paste0(latest_update(), "_tests.xlsx")
+  ),
+  overwrite = TRUE
+)
```
Moohan, line 57: We spoke about having this create the workbook if it doesn't yet exist. Also packaging it into a function e.g.

`write_tests_xlsx <- function(comparison_data, name)`

Jennit07:
I've wrapped the excel output code into a function called `write_tests_xlsx` and implemented this into the test scripts for: 
- pc lookup
- gp prac lookup 
- deaths lookup
- sc demographic
- sc_all_ch_episodes

This can also be used for B file testing as well. I think this is ready to merge if the changes look okay :)

Moohan:
@Jennit07 Can you resolve the merge conflicts with main-R before I check this please?

Jennit07:
@Moohan Merge conflicts dealt with. Looked like it was just the styling :)

Moohan (approved):
Looks good to merge

PR #165, opened 2022-03-21
Rename Extract function

catherineholland1:
rename of extract function 

added other extract file paths to the function to save it being edited when those are converted. hopefully have included all the ones with this file location.


catherineholland1:
added an edit to `get_source_extract_path`, this is less straightforward to what files/file names are required here so may need updating if another extract needs adding

Moohan (approved):

PR #170, opened 2022-03-22
Source Extract Test function

catherineholland1:
new function similar to the other extract functions which creates the file path and name to write the extract tests to


Jennit07:
This is good but as discussed, i don't think we need this now. I've updated the excel workbook function on the PR amend-tests-function so we can use that for tests going forward in the B files testing. Just going to close this PR, can you delete the branch if okay, thanks :)

PR #172, opened 2022-03-29
Create monthly beddays

Moohan:
I was getting a build error that I couldn't resolve, I assume it's from my new code...
```
> checking dependencies in R code ... WARNING
  '::' or ':::' import not declared from: ‘purrr’
```

I used a [snapshot](https://testthat.r-lib.org/articles/snapshotting.html) for the tests as that's simpler for the bigger results. Hopefully, the output is actually correct! You might want to add in some more tests / test dates if you're not convinced though.


Jennit07:
@Moohan I'm not too sure what that error is either. I had this before as well and i think it started to appear with the phs connection function but i'm not sure what resolved it as i haven't seen it in a while. Does this appear when you do a check on main-R?

Moohan:
> @Moohan I'm not too sure what that error is either. I had this before as well and i think it started to appear with the phs connection function but i'm not sure what resolved it as i haven't seen it in a while. Does this appear when you do a check on main-R?

Solved this, it was actually two separate issues 🙄 just about to push the fix now.

Jennit07 (approved):
Tested this against my maternity data and i think this is working as expected. Assigns the correct number of beddays to each month with the relevant abbreviations. Thanks very much for picking this up James, great work :)

PR #173, opened 2022-03-29
Update cost paths

Jennit07:
Now includes option update = NULL where update = latest_update() can be specified for saving the cost lookups


Moohan (approved):
Code looks fine to me

PR #175, opened 2022-03-30
Add package openxlsx

Jennit07:
Build check now has 0 errors, warning and notes. Minor change to write xlsx test script

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -25,6 +25,7 @@ Imports:
     lubridate (>= 1.8.0),
     magrittr (>= 2.0.1),
     odbc (>= 1.3.1),
+    openxlsx,
     phsmethods (>= 0.2.0),
     phsopendata (>= 0.0.0.900),
     readr (>= 2.1.0),
--- a/R/write_tests_xlsx.R
+++ b/R/write_tests_xlsx.R
@@ -4,6 +4,7 @@
 #' @param name of dataset tests
 #'
 #' @return a workbook containing comparison tests with each dataset with a new tab
+#'
 #' @export
 #'
 write_tests_xlsx <- function(comparison_data, name) {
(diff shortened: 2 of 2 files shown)


DESCRIPTION
```diff
@@ -25,6 +25,7 @@ Imports:
     lubridate (>= 1.8.0),
     magrittr (>= 2.0.1),
     odbc (>= 1.3.1),
+    openxlsx,
```
Moohan, line 28:
```suggestion
    openxlsx (>= 4.2.5),
```
Pick a reasonable version, I just picked the current latest. See the [openxlsx changelog](https://ycphs.github.io/openxlsx/news/index.html)

R/write_tests_xlsx.R
```diff
@@ -4,29 +4,30 @@
 #' @param name of dataset tests
 #'
 #' @return a workbook containing comparison tests with each dataset with a new tab
+#'
 #' @export
 #'
```
Moohan, line 9:
```suggestion
```

R/write_tests_xlsx.R
```diff
@@ -4,29 +4,30 @@
 #' @param name of dataset tests
 #'
 #' @return a workbook containing comparison tests with each dataset with a new tab
```
Moohan, line 6: This function doesn't return the workbook. It writes the workbook to disk. I'm not actually sure what it returns i.e. if you did `object <- write_tests_xlsx(...)` what would `object` be?

Some options are to return nothing `return(NULL)`, return TRUE to show that it worked. If this would be used in a pipe it might make more sense to either return the data `return(comparison_data)` (this is usually what the readr and other type functions do) or to return the file path where the xlsx was saved.

Moohan (changes requested):

R/write_tests_xlsx.R
Jennit07: This makes sense/I agree, the function writes the workbook to disk so i dont think we would use this applied to an object. We would use this in a pipe, so i think return the file path where the xlsx was saved makes sense to me. I've made this change and think this is ready now :)

PR #181, opened 2022-04-01
Rename production scripts

Moohan:
I'm not tied to the new names, I just thought it was getting messy and inconsistent. There's no real reason to stick to the SPSS format which had issues itself, with sub-scripts (a, b, c) and numbers that are missing from some years.


Jennit07:
@Moohan Split scripts into folders as discussed today. I was thinking having sub-folders for tests within each one as applicable. Also put others e.g. episode/individual processing for later down the line. Let me know what you think about the format, thanks :)

Jennit07 (approved):

PR #184, opened 2022-04-05
Create monthly costs fn

catherineholland1:
taking this from within a outpatient PR as a quick one to allow to work within A&E


Moohan:
Is this not the same as #183?

catherineholland1:
> Is this not the same as #183?

Yes @Moohan just checked sorry it was still on the outpatient branch and I got confused with the name change etc I didn't realise both the same - will delete this PR and branch

catherineholland1:
not needed

PR #187, opened 2022-04-06
add function to check the year format

Moohan:
New function to check the year type. The intention was to be used in a bit Jen has in Acute (fix C3 costs) but can be used elsewhere too. I had to modify some tests as it can't be properly used with a vector.


R/check_year_format.R
```diff
@@ -0,0 +1,44 @@
...
+#' @param format the format that year should be using. Default is "fyyear" for
+#' example `1718`, the other format available is "alternate" e.g. `2017`
+#'
+#' @return The year if the check passes, it will be converted to a character if
+#' not already.
+#' @export
+#'
+#' @examples
+#' year <- "1920"
+#' year <- check_year_format(year)
```
Moohan, line 13:
```suggestion
#' year <- check_year_format("1920")
```

R/check_year_format.R
```diff
@@ -0,0 +1,44 @@
...
+  } else if (format == "alternate") {
+    if (as.integer(first_part) != 20) {
+      rlang::abort(c(
+        "The year has been entered in the wrong format.",
+        "Try again using the alternate form, e.g. `2017`",
+        "Or use the function `convert_fyyear_to_year`."
+      ))
+    }
+  }
+
```
Moohan, line 42:
```suggestion
  if (format == "fyyear") {
    if (any(as.integer(first_part) + 1L != as.integer(second_part))) {
      rlang::abort(c(
        "The year has been entered in the wrong format.",
        "Try again using the standard form, e.g. `1718`",
        "Or use the function `convert_year_to_fyyear`."
      ))
    }
  } else if (format == "alternate") {
    if (any(as.integer(first_part) != 20)) {
      rlang::abort(c(
        "The year has been entered in the wrong format.",
        "Try again using the alternate form, e.g. `2017`",
        "Or use the function `convert_fyyear_to_year`."
      ))
    }
  }

```

PR #188, opened 2022-04-06
Amend monthly beddays

Jennit07:
Now includes option to `include_costs = TRUE` which will work out costs and assign them to each month. A few tweaks to the original syntax as this was breaking during testing. Specifically against the str_replace for renaming beddays to cost. This code is now working against maternity to calculate monthly beddays and costs.

### DIFF
--- a/R/create_monthly_beddays.R
+++ b/R/create_monthly_beddays.R
@@ -60,12 +60,13 @@ create_monthly_beddays <- function(data, year, admission_date, discharge_date, i
   data <- dplyr::bind_cols(data, beddays) %>%
     dplyr::select(-.data$stay_interval)
 
-  if (include_costs == TRUE) {
+  # Work out costs for each month if applicable
+  if (include_costs) {
     costs <- beddays %>%
-      rename_with(~ str_replace(., "_beddays", "_costs"))
+      dplyr::rename_with(~ str_replace(., "_beddays", "_costs"))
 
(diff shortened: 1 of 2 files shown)


R/create_monthly_beddays.R
```diff
@@ -59,5 +60,14 @@ create_monthly_beddays <- function(data, year, admission_date, discharge_date, c
   data <- dplyr::bind_cols(data, beddays) %>%
     dplyr::select(-.data$stay_interval)
 
+  # Work out costs for each month if applicable
+  if (include_costs) {
+    costs <- beddays %>%
+      dplyr::rename_with(~ str_replace(., "_beddays", "_costs"))
+
+    data <- dplyr::bind_cols(data, costs) %>%
+      dplyr::mutate(dplyr::across(dplyr::ends_with("_costs"), ~ dplyr::if_else(yearstay != 0, .x / yearstay * cost_total_net, 0)))
```
Moohan, line 69: Sorry, realised that I suggested this, but should this not be:
```suggestion
    data <- dplyr::bind_cols(data, costs) %>%
      dplyr::mutate(dplyr::across(dplyr::ends_with("_costs"), ~ dplyr::if_else(.x!= 0, .x / yearstay * cost_total_net, 0)))
```
i.e. if beddays are zero then set cost to be zero, otherwise `beddays / yearstay * cost`

Moohan (approved):

PR #169, opened 2022-03-22
Acute r

Jennit07:
I think this is ready to be merged into main-R, although still working through the tests on another branch which can then be applied to other B file extracts. One change i am aware of will be the `eng_gpprac_to_dummy` function which is currently work in progress on another PR. This will need brought up to date with the branch and changed to `convert_eng_gpprac_to_dummy` in the script

PR #193, opened 2022-04-12
Extract tests/Acute tests

### LINKED ISSUE #178: Convert Acute tests script
New branch for Acute tests. Go back and amend this

Jennit07:
Tests developed for processed extracts. This follows the same testing format as the SPSS syntax, e.g. measure, value, percentage difference. The layout for these tests are split into various functions and this is working specific to acute which can be tested on this branch. However the format can be applied to other extract tests. 

This starts with the `extract_acute_incl_gls_tests` script which will run a new 'processed' data frame and compare this with an 'old' data frame taken from the SLF episode file. This will link to the function `produce_source_acute_tests` which links to other `test flags functions` and create a summary of each measure. The new `calculate_measures` function will then create a summary dataframe specific to variables that are passed to the function. This includes: Sum, Mean, min/max and an option for this joined together. More measures can be added to this as required, but for start up this will produce a dataframe which works for the variables needed for acute tests. These summary dataframes are then joined together to create a whole summary of tests comparing 'new' processed data to 'current/old' data in the SLF episode file. 

I had some issues with this due to the data showing in scientific notation. However this is now set up that it is okay if this is in scientific notation in R because the data is outputted in excel which allows you to view the figures okay. Hope this bit of background helps/makes sense. If you have any questions, let me know. Thanks :) 

Closes #178


R/produce_source_acute_tests.R
```diff
@@ -0,0 +1,39 @@
...
+    # keep variables for comparison
+    select(c(.data$valid_chi:.data$NHS_Lanarkshire_cost)) %>%
+    # use function to sum new test flags
+    calculate_measures(measure = "sum")
+
+  all_measures <- data %>%
+    calculate_measures(vars = c("beddays", "cost", "yearstay"), measure = "all")
+
+  min_max <- data %>%
+    calculate_measures(c("record_keydate1", "record_keydate2", "cost_total_net", "yearstay"), "min-max")
```
Moohan, line 29:
```suggestion
  all_measures <- data %>%
    calculate_measures(vars = c("beddays", "cost", "yearstay"), measure = "all")

  min_max <- data %>%
    calculate_measures(vars = c("record_keydate1", "record_keydate2", "cost_total_net", "yearstay"), measure = "min-max")
```

R/write_tests_xlsx.R
```diff
@@ -18,11 +18,11 @@ write_tests_xlsx <- function(comparison_data, name) {
...
   # add a new sheet for tests
-  openxlsx::addWorksheet(wb, name)
+  openxlsx::addWorksheet(wb, paste0(name, "-", format(Sys.Date(), "%d-%m-%Y")))
   # write comparison output to new sheet
   openxlsx::writeData(
     wb,
-    name,
+    paste0(name, "-", format(Sys.Date(), "%d-%m-%Y")),
     comparison_data
   )
```
Moohan, line 27: `paste0(name, "-", format(Sys.Date(), "%d %b"))` will create a shorter / nicer name. Also store the sheet_name as a variable. Possible rename the parameter from `name` to `sheet_name` then call the new variable `sheet_name_dated`

Moohan (changes requested):
Just did a quick review. Once you've had a look and made any changes can you re-request, please?

If it's not too tricky could you also remove the acute specific files e.g. `Production_scripts/Process_extracts/extract_acute_incl_gls.R`, `R/convert_monthly_rows_to_vars.R`, `R/fix_c3_costs.R`...etc? If it's a faff then we can just prioritise merging #169 before I re-check this!

Production_scripts/Process_extracts/Produce_extract_tests/extract_acute_incl_gls_tests.R
```diff
@@ -0,0 +1,49 @@
...
+year <- "1920"
+
+# Read current SLF episode file and filter for 01B and GLS records
+slf_ep_1920 <- read_slf_episode("1920", recid = c("01B", "GLS")) %>%
+  get_chi()
+
+
+# Create new and old dataframes with measures for testing
+# To compare new slf extract to current file in hscdiip
+new_tests <- produce_source_acute_tests(readr::read_rds(get_source_extract_path(year, "Acute", ext = "rds")))
```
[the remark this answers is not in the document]
Jennit07: Going to leave this for now as the column chi is not available in the existing data and breaking the code. The get_chi() function here changes anon_chi to match chi in the new file. I also tried using the tolower function for selecting lowercase names but couldnt exclude chi unless there was another way to do this?

R/produce_sc_ch_episodes_tests.R
```diff
@@ -25,5 +25,5 @@ produce_sc_ch_episodes_tests <- function(data) {
       .data$ch_nursing, .data$ch_adm_reason, .data$sc_latest_submission
     )) %>%
     # use function to sum new test flags
-    sum_test_flags()
+    calculate_measures(measure = "sum")
```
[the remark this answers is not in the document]
Jennit07: Tested this and it works okay. If the vars are NULL and the measure is sum then this will summarise for test flags. If we want the sum and mean then this is specified as all but will need to specify variables which will match e.g. 'beddays', 'cost'

Production_scripts/Process_extracts/Produce_extract_tests/extract_acute_incl_gls_tests.R
```diff
@@ -18,26 +14,24 @@ library(openxlsx)
...
 
 year <- "1920"
 
+# Read new data file
+new_data <- readr::read_rds(get_source_extract_path(year, "Acute", ext = "rds"))
+
 # Read current SLF episode file and filter for 01B and GLS records
-slf_ep_1920 <- read_slf_episode("1920", recid = c("01B", "GLS")) %>%
+existing_data <- read_slf_episode("1920", recid = c("01B", "GLS")) %>%
   get_chi()
```
Moohan, line 26: Just had a thought that we probably don't need the chi number. We just use chi or anon_chi to count individual people so we could just rename anon_chi to chi and it would serve the same purpose.

```suggestion
  rename(chi = anon_chi)
```

Alternatively, you could alter the functions so that they can take chi or anon_chi.

Moohan (approved):
Looks good. I think there will probably be changes when we start applying it to other data sets but let's get it merged and start using it!

PR #201, opened 2022-05-04
Maternity tests

### LINKED ISSUE #200: Convert Maternity tests script
Testing comparing processed SOURCE maternity extract to existing data in the SLFs

Jennit07:
Testing for processed SOURCE maternity script

Closes #200


Production_scripts/Process_extracts/Produce_extract_tests/extract_maternity_tests.R
```diff
@@ -0,0 +1,43 @@
...
+library(openxlsx)
+library(slfhelper)
+
+
+# Read in Data-----------------------------------------
+
+year <- "1920"
+
+# Read new data file
+new_data <- readr::read_rds(get_source_extract_path(year, "Maternity", ext = "rds"))
```
catherineholland1, line 22: issue with reading this in - should be fixed when functions updated (check_mode read/write)

catherineholland1 (approved):
This looks good and runs well - only issue was reading in the Maternity data but that should be fixed when the file paths are updated. 

Only other comment would be if it is worth adding in any plots to identify any differences?

Jennit07:
This one is ready to be merged. May need to be reviewed again as mentioned in #239 that the same acute function could be applied to create the tests.

catherineholland1:
> This one is ready to be merged. May need to be reviewed again as mentioned in #239 that the same acute function could be applied to create the tests.

linked to PR #242 which renames the test function

catherineholland1:
> > This one is ready to be merged. May need to be reviewed again as mentioned in #239 that the same acute function could be applied to create the tests.
> 
> linked to PR #242 which renames the test function

PR merged - get branch up to date then `produce_source_extract_tests()` will be available to use

Production_scripts/Process_extracts/Produce_extract_tests/extract_maternity_tests.R
Moohan: @Jennit07 / @catherineholland1 can we check this is fixed now, please? If Jen deletes the maternity rds file (which might already not exist) then re-create it with the latest functions/versions of the package. Hopefully, then everything will work for Catherine.

catherineholland1: I just re ran the maternity extract and saved out the .rds file and now it reads in error free

PR #203, opened 2022-05-04
Amend sc scripts

### LINKED ISSUE #202: Use SC snapshot data for June 22 update
Currently we read in data that has not been validated by DM in the 'snapshot' but we ask the social care team for the most up to date validated period and implement this in the code. Using the snapshot data will skip this step and we will not need to ask the social care team to confirm the validated data. This will also reduce the filtering when reading in the data from the platform

Jennit07:
I think this only affects care homes and home care all episodes for now. If i have missed anything else, please let me know. I have also looked at the data before and after using the snapshot which i can show you. Thanks

Closes #202


Moohan:
I think there are some extra bits to add 😉

- [ ] Demographics file
- [ ] Client lookup

Jennit07:
Good spot! changes made

Moohan (approved):

PR #209, opened 2022-05-10
Function to extract relevant data for tests from the SLFs

Moohan:
This function should extract the relevant data for the tests from the SLFs. It should mean that the tests run a bit quicker as we'll only be reading the least amount of data necessary to run them.


Jennit07 (approved):
I've made one minor change in the tests script where the function calls the parameter `year`. This was breaking the code but now works as expected. Happy to merge

PR #210, opened 2022-05-11
File permissions fix

Moohan:
I've made two new functions which basically wrap `readr::write_rds()` and `haven::write_sav` but also correct the file permissions on the newly written file, i.e. make it writable to anyone in the group.

I think the functions work as expected (there are also some tests which pass), but there could still be some issues with the implementation.

- [x] Have I missed any instances which could be replaced?
- [ ] Did I replace everything correctly?
- [x] The LTC file looks broken so I didn't touch that.
- [x] Are there any instances where we load haven or readr at the top of the script? - In theses cases they should be removed or `library(createslf)` must come after them in the list.


Jennit07:
LTC fixed on amend_check_mode PR

tests/testthat/test-write_data.R
```diff
@@ -0,0 +1,39 @@
...
+
+  # Check data is the same
+  expect_equal(haven::zap_formats(read_data), dummy_data)
+
+  # Check the permissions are as expected
+  expect_match(
+    as.character(file_info$permissions),
+    "rw\\-rw\\-\\-\\-\\-"
+  )
+})
```
catherineholland1, line 22: sav test failed for me? unsure if you got this when ran. 

`── Error (Line 3): write_sav creates a file with the correct permissions ─────────────────────
Error: Expected single logical value
Backtrace:
 1. global write_sav(dummy_data, temp_path)
 2. haven::write_sav(data = data, path = path, compress = compress)
 3. haven:::write_sav_(...)`

catherineholland1 (commented):
one little issue in the sav test but apart from that all instances seem to have been changed and all run as expected :)

tests/testthat/test-write_data.R
Jennit07: I don't get this error here, the tests seem to be working okay for me

Jennit07 (changes requested):

Jennit07:
Just one thing to check from me!

Jennit07:
This one might be finished? I think we were going to merge Amend check mode #214 first and then bring this up to date but this could probably be merged first and i can deal with the conflicts

tests/testthat/test-write_data.R
Moohan: Can you confirm you still get an error, Catherine? Is it possible that you're using an old version of haven? I'm using the `compress` argument `"zsav"` which is only valid from [version 2.5.0](https://haven.tidyverse.org/news/index.html#haven-250), `TRUE` and `FALSE` still work for backwards compatibility but using `"zsav"` is clearer.

I suspect your error is that the old version expects a logical so doesn't like the `"zsav"`  bit.

catherineholland1: updated `haven` and test works now :)

PR #212, opened 2022-05-12
Tidy up the imports

Moohan:
I removed all the imports from dplyr. I was in two minds about removing the imports for `if_else` and `case_when`.

I added some new imports for the 'column selection helpers' from readr, and moved them to a central place rather than on the homelessness function. If we add the `if_else` and `case_when` back in they should also be added to `createslf-package.R`


catherineholland1 (approved):
happy to merge :)

Issue #205, opened 2022-05-06
Update CIJ marker SPSS type


Moohan:
Some years now have a length 4 CIJ marker i.e. > 999. This means that in older files CIJ marker is a length 3 string and newer it's a length 4.

We could add a line to alter the type to string 4 or even 5 `Alter type cij_marker (A5)` but actually we would be much better storing it as a number. @Jennit07 can you think of any reason switching `cij_marker` from a string to a number could break our or anyone else's code? I guess if you did something like `cij_marker == "1"` but I'm not sure why anyone would do that... I think we only use the `cij_marker` in C01 and the DD matching code but don't think either of those would be impacted by the change in type?

Jennit07:
PR complete #213

PR #213, opened 2022-05-12
Change the `cij_marker` type to a numeric

Moohan:
I did a search for all instances of `cij_marker` in the SPSS code and then checked them all for types.

I've changed the type from String to numeric on the initial read.

I also changed any instances where the `cij_marker` was used. None of this has been tested.

@Jennit07 if you can check and test this (to some extent) that would be great. You can approve but please don't merge until we decide if we're actually using it! The alternative to this would be to set the `cij_marker` to be `(A5)` which would be much fewer changes, but leave it as a string.

Closes #213

### DIFF
--- a/2014-15/B01 Process Acute (SMR01 and GLS) extract.sps
+++ b/2014-15/B01 Process Acute (SMR01 and GLS) extract.sps
@@ -1,4 +1,4 @@
-﻿* Encoding: UTF-8.
+* Encoding: UTF-8.
 
 ********************************************************************************************************.
  * Run 01-Set up Macros first!.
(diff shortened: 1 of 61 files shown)


Moohan:
I just did a force-push of my rebase: 

- Commits are split out so the changes can be seen separately.
- C01 is fixed so that only the correct changes are applied!
- Some files/changes were missing for 2016/17, they're now added too.

Jennit07 (approved):
I've tested this code using 19/20 data by running up to the end of C10. There were a few bugs which i have now resolved and pushed to the branch. This now keeps the cij_marker as a numeric variable and i have updated the news section. Ready to merge. Closes #205

PR #214, opened 2022-05-16
Amend check mode

Jennit07:
I think I have captured everything here, however, if I have missed any, feel free to correct me. This now includes `...` as a parameter and will read the default file path as `.rds`. We can specify `ext = "zsav"` if we want to save or read the file in SPSS format and also specify `check_mode = "write"` if we want to write the file to disk. These file path functions are set to `read` by default as this is passed through the `get_file_path` function.

Scripts should now be in this format for writing to disk: 

SPSS: 
```r
# Reading
haven::read_sav(get_source_extract_path(year, "Outpatient", ext = "zsav"))
# Writing
haven::write_sav(get_source_extract_path(year, "Outpatient", ext = "zsav", check_mode = "write"), compress = "zsav")
```

R: 
```r
# Reading
readr::read_rds(get_source_extract_path(year, "Outpatients"))
# Writing
readr::write_rds(get_source_extract_path(year, "Outpatients", check_mode = "write"), compress = "xz")
```

I have also amended the LTCs script on this branch as I was changing file paths and came across an error. This now includes a new function `get_ltcs_path()` in the same format as others here. There was a part of the code where this was breaking as an older object was specified while trying to do a save. This has been amended and should now work.

PR #216, opened 2022-05-16
Fix ltc script

Moohan:
I extracted the changes that Jen made on #214 so that they can be checked and merged separately, which will make that PR simpler!

I also made a few minor tweaks and additions.


catherineholland1 (approved):
happy to merge

PR #217, opened 2022-05-16
Add home care costs

Moohan:
This PR adds costs to Home Care. There are quite a few parts to it!

There is a spreadsheet from the Improvement Service which contains the costs, these are copied into our spreadsheet "hc_costs.xlsx", from there an R script picks that up and processes them into a lookup - this follows the pattern of the other R costs scripts but it was done out of the context of the `createslf` R package so doesn't use any functions. That's so that it can be run in isolation for this update without relying on anything else in the package. The next bit is amendments to the R script which processes the raw home care data into an episodic format. The penultimate changes are to the yearly SPSS scripts to bring the cost variables into those. Finally, cost variables are added to the episode and individual files.


All_years/03-Costs/Costs-04-home_care.R
```diff
@@ -0,0 +1,138 @@
...
+# Packages #
+
+library(dplyr)
+library(ggplot2)
+library(tidyr)
+library(purrr)
+
+costs_dir <- fs::path("/conf/hscdiip/SLF_Extracts/Costs/")
+hc_costs_path <- fs::path(costs_dir, "costs_hc_lookup.rds")
+hc_costs_path_old <- fs::path(costs_dir, "costs_hc_lookup_pre-Jun_2022.rds")
```
Jennit07, line 24: I wonder if it is worth specifying latest_update or previous_update as an object here? then when it comes to running this we know to update the 'old' path? Although i realise we will want to convert this into the r conversion project anyway so this might not be necessary

All_years/03-Costs/Costs-04-home_care.R
```diff
@@ -0,0 +1,138 @@
...
+  labs(fill = "Local Authority", x = "Year", y = "% difference")
+
+
+## Plot to check for obviously wrong looking costs ##
+
+ggplot(data = matched_costs_data, aes(x = year, y = hourly_cost)) +
+  geom_line(aes(group = ca_name, color = ca_name)) +
+  scale_y_continuous(labels = scales::label_dollar(prefix = "£")) +
+  labs(y = "Hourly costs", color = "Local Authority") +
+  facet_wrap(~health_board)
```
Jennit07, line 129: The plots here aren't working right for me? They are returning the x and y axis but no line/bar graph

All_years/03-Costs/Costs-04-home_care.R
Moohan: Yep, good point the June_2022 bit should be in a `latest_update` variable.

All_years/03-Costs/Costs-04-home_care.R
Moohan: I think that's fine, they compare current to previous but at the moment that's the same file so the `filter(x > 0)` removes all the data.

The last plot should be working though?

Jennit07: Good point, that makes sense! 

The last plot wasn't working for me. Just ran again and its because the `matched_costs_data` object contains two `health_board` variables. I've just specified `health_board.x` in the facet_wrap and this is now working

Jennit07 (changes requested):
Looks good, the code itself seems to be working correctly and i think this looks okay. I've made a few small changes..

I've applied the suggested changes where this now includes an object for `latest_update`. One small change for the last chart which is now working okay. I've also declared hc variables for years before 2017/18. Let me know if you are happy with the changes and i'll approve the PR. Thanks :)

PR #221, opened 2022-05-17
Include SC Alarms-Telecare data in SLFs

### LINKED ISSUE #75: Add Alarms & Telecare data

Jennit07:
Alarms and Telecare data for the source linkage files. We have set up a meeting with the social care team to discuss outstanding data cleaning and how to deal with some aspects. I've tested this script with trying to change missing start dates to the start of the FY. This poses the question of 2017 records and whether this should be the start of FY or start of Q4. Once running this script in R i have ran the output in SPSS which seems to be working fine. If you have any suggestions or if I've missed anything, please let me know. Thanks! 

Outstanding:

- [ ] Records assigned to 2017 (possibly change to 2017Q4 in line with home care data) and 2020 with no quarter
- [ ] Missing service start dates (~11,000, large number of records in Aberdeen City) 
- [ ] Copy SPSS syntax to 2017/18 onwards
- [ ] Declare variables in old files <2016/17
- [ ] Add AT extract/tests to C01 List

Some suggestions for dealing with missing start dates to ask sc team: 
- Drop the records
- Assign to the start of the FY
- Assign to the start of Quarter

Closes #75


Jennit07:
Now not including Alarms Telecare in June-22-Update. Had a call with social care team and our questions above were answered, however there is now the question as to why we do not merge records. We need to spend some more time looking into this as there may be multiple submissions across a quarter with the same start date.

Jennit07:
This is ongoing/work still needing done to the logic

PR #225, opened 2022-05-18
Make fst tweaks

Moohan:
Made the changes as discussed to create a copy before starting the write out.

Also made some other QoL changes


Jennit07:
This needs testing when we get to the end of the SLF update but i think this should work as we discussed

Jennit07 (approved):
Tested this on one year which worked fine and also ran for other years. Happy to merge this

PR #226, opened 2022-05-18
Replace Homelessness SPSS code with an error

Moohan:
This doesn't delete the file, so it will still be there as a reminder to use the R code but it can't accidentally be used anymore.

If it is used e.g. in existing production jobs it will create an error and print the messages to the output.

We still need the SPSS test file.

It's also worth noting that all the SPSS code will be kept in the git history.


Jennit07 (approved):
This is nice, as the error is clear and we have a reminder to run this syntax in R. Ready to merge

PR #229, opened 2022-05-20
Fix the CH costs year

Moohan:
The year in the spread sheet (e.g. 2018) refers to the end of the year, therefore the 2018 column contains costs for 2017/18, not 2019/20 as we were using.


Jennit07 (approved):
This is now working as expected and 2018 refers to 2017/18 costs. Happy to merge

PR #230, opened 2022-05-20
BUG - Fix overcounting of preventable beddays

Moohan:
We were previously counting the beddays for the whole year on every episode. So if a CIJ had 2 episodes it would have 2X the correct number of beddays etc.


Jennit07 (approved):
Tested this on 2019/20 and i think the correct number of preventable beddays is now pulling through. I thought there was a bug for a missed string to sysmis change but this must have been on another branch. Happy to merge.

PR #235, opened 2022-05-23
Add more clusters to the GP practice lookup

Moohan:
The main change here is to take more historic data from the GP Practice open data platform which will bring through more clusters for practices.

The other changes are to add in some extra variables which aren't used in the SLFs but might be used in other projects e.g. Matrix Dashboard


Moohan:
This test shows the new code against the previous update but we can see that there is an increase in the number of practices which now have clusters.
image

Jennit07 (approved):

PR #238, opened 2022-05-24
# 96 Mental Health Extract

### LINKED ISSUE #96: Convert Mental Health extract processing

catherineholland1:
Closes #96 

conversion of mental health extract 

slight issue with summarising costs/beddays and then the rest of the variables as so many to type out so have done separately to use 'summarise_all` on the rest then join together. any other suggestion of how to do this efficiently? 

changed the file path type on BOXI and source extract functions too.


Moohan:
See Jen's Acute code, the data is in a similar structure to that. I think she just does a pivot, rather than all the aggregation stuff we did in SPSS

catherineholland1:
> See Jen's Acute code, the data is in a similar structure to that. I think she just does a pivot, rather than all the aggregation stuff we did in SPSS

I can't see any pivot within acute code. perhaps can discuss/find on Monday ?

PR #239, opened 2022-05-24
#177 A&E Extract Tests

### LINKED ISSUE #177: Convert A&E tests scripts
New branch for A&E tests. Go back and amend this script

catherineholland1:
Closes #177 

A&E extract tests

1. The functions created for acute work with no change for A&E extract too. Shall the `produce_source_acute_tests` function be re-named if it is applicable to more than one? Saves the exact code being re-written and saved with another name. 

2. I got a 'Permission denied' error when I tried to save out the output, is it possible for this to be changed? I can't see anything within the `write_tests_xlsx` function to show why


Jennit07:
hmm, i think this is the same for maternity, maybe this could be further simplified and put into another function? Maybe some restructuring for the tests! I suspect this might be the same for MH tests as well so something to think about!

The tests are saved in a new folder in HSCDIIP called 'Tests' i'm not sure why it won't let you as it says the group has access?

Production_scripts/Process_extracts/Produce_extract_tests/extract_ae_tests.R
```diff
@@ -0,0 +1,45 @@
...
+
+# Read in Data-----------------------------------------
+
+year <- "1920"
+
+# Read new data file
+new_data <- readr::read_rds(get_source_extract_path(year, "AE", ext = "rds"))
+# zsav until rds file avaliable
+new_data <- haven::read_sav(get_source_extract_path(year, "AE", ext = "zsav")) %>%
+  rename(postcode = "Postcode")
```
Jennit07, line 25: I think you can run the rds file and remove this line, i just deleted the rds files in the folder for space but if you need it then this can be ran again. Let me know if you need others kept

Production_scripts/Process_extracts/Produce_extract_tests/extract_ae_tests.R
```diff
@@ -0,0 +1,45 @@
...
+# Read current SLF episode file and filter for 01B and GLS records
+existing_data <- get_existing_data_for_tests(new_data = new_data)
+
+
+# Produce comparison-------------------------------------
+# Compare new file with existing slf data
+comparison <- produce_test_comparison(
+  produce_source_acute_tests(existing_data),
+  produce_source_acute_tests(new_data)
+)
```
Jennit07, line 36: If we are using this function for others as well we can re name this to something more generic

Moohan:
> 2. I got a 'Permission denied' error when I tried to save out the output, is it possible for this to be changed? I can't see anything within the `write_tests_xlsx` function to show why

We said that `write_tests_xlsx` should be modified so that after the file has been written it modifies the permissions, as we do in `write_sav` and `write_rds`.

Jennit07:
> > 2. I got a 'Permission denied' error when I tried to save out the output, is it possible for this to be changed? I can't see anything within the `write_tests_xlsx` function to show why
> 
> We said that `write_tests_xlsx` should be modified so that after the file has been written it modifies the permissions, as we do in `write_sav` and `write_rds`.

Now created a PR (#252) to implement this change in `write_tests_xlsx` function

catherineholland1:
merged to bring up with new test permissions fix

Jennit07 (approved):
Looks good for A&E tests. There is a small bug where the dates are not converting into numeric properly and causing the tests to show -100%. I will approve/merge this for now and create a separate issue as this will also affect other tests

PR #241, opened 2022-05-25
Update monthly cost variables for all changed costs

Moohan:
The main issue this change fixes is that previously only the monthly variables for A&A were being updated after cost_total_net was for both A&A and FV. This change fixes that.

It also adds some extra selections to make extremely clear what is being affected.

With the new year selector we could in theory run this code on every year as it would only actually do anything for `year >= "1819"`, this is like the R version.


Jennit07 (approved):
I see how FV was missed in this code! looks good to me!

PR #242, opened 2022-05-30
Rename extract test function

catherineholland1:
quick one: rename of extract test function as it is applicable to more than one extract. 

named `produce_source_extract_tests()` and wait to see if it is applicable to all extracts, if it can not be used by all extracts can easily rename at that point once all extract tests written. unless any other better suggestions?

changed it on merged acute script. 

this will affect current open PR #239 and #201, it will need merged before those.


R/produce_source_acute_tests.R
```diff
@@ -1,4 +1,4 @@
-#' Produce the Acute tests
+#' Produce the Source Extract tests
 #'
 #' @param data new or old data for testing summary flags
 #' (data is from \code{\link{get_source_extract_path}})
```
Moohan, line 4: Add a description to the function documentation.

```suggestion
#' Produce the Source Extract tests
#'
#' @description Produce a set of tests which can be used by most
#' of the extracts. 
#' This will produce counts of various demographics
#' using [create_demog_test_flags()] counts of episodes for every `hbtreatcode`
#' using [create_hb_test_flags()], a total cost for each `hbtreatcode` using
#' [create_hb_cost_test_flags()]. 
#' It will also produce various summary statistics for bedday, cost and
#' episode date variables.
#'
#' @param data new or old data for testing summary flags
#' (data is from \code{\link{get_source_extract_path}})
```

Moohan (approved):

PR #247, opened 2022-06-02
Fix some build errors/warnings.

Moohan:
The `surpressWarnings` is needed as `{slfhelper}` loads `{fst}` which produces the warning:
```
fst package v0.9.8
Warning message:
package ‘fst’ was built under R version 3.6.3
```
I could/should probably suppress this more cleverly in slfhelper itself!

The `get_practice_details` test failure is just because we now also produce an rds version. These errors will become more common as we make rds versions and eventually stop producing zsav versions, so we just need to stay on top of them.


catherineholland1 (approved):
happy to merge :)

PR #266, opened 2022-06-14
Amend tests function

Jennit07:
I've updated the function `produce_source_extract_tests` to now include two extra parameters for selecting the variables used in the `calculate_measures` function. The idea behind this is to match the variables we need for tests. For example, outpatients does not have beddays so the existing code would not work for this instance. By specifying the variables in this parameter it allows us to select 'cost' only or 'beddays' only if needed. It would not work for changing the variables specified in `calculate_measures` function as this needs to be set as NULL so this will summarise the test flags. I'll provide an example below of what i think this would look like in the outpatients test script. 

Let me know if the new changes don't make sense or if you think the parameter names could be changed as i wasn't too sure what to call these for them to make the most sense! I went with all_vars and max_min_vars as this is where this would change and they need to be separate. Thanks


Jennit07:
Compare new file with existing slf data
  produce_source_extract_tests(existing_data,
                               all_vars = "cost",
                               max_min_vars = c("record_keydate1", "record_keydate2", "cost_total_net")),
  produce_source_extract_tests(new_data,
                               all_vars = "cost",
                               max_min_vars = c("record_keydate1", "record_keydate2", "cost_total_net")))

R/produce_source_extract_tests.R
```diff
@@ -20,7 +22,12 @@
 #' @seealso [create_hb_test_flags()],
 #' [create_hscp_test_flags()] and [create_hb_cost_test_flags()]
 #' for creating test flags
-produce_source_extract_tests <- function(data) {
+produce_source_extract_tests <- function(data,
+                                         all_vars = c("beddays", "cost", "yearstay"),
```
catherineholland1, line 26: perhaps change to something more informative? is all sum and mean? maybe

```suggestion
                                         sum_mean_vars = c("beddays", "cost", "yearstay"),
```

catherineholland1 (changes requested):
just a suggestion on the name but apart from that looks good and makes sense

Will you be changing this on acute and maternity (think the only ones merged so far?)

Jennit07:
I think on maternity and acute it might be okay as the default variables are cost and beddays. Whereas outpatients only called upon cost variables, therefore this is specified. Will test this on Maternity, Acute and A&E tests once merged

catherineholland1 (approved):

PR #268, opened 2022-06-15
Remove unnecessary code

Jennit07:
Quick one - spotted a bug in the tests which was preventing the max and min dates being calculated. This was converting the variables to numeric and not producing the correct format. This was then saying the 'new' data has dropped by 100% when this was not the case.


Jennit07:
Not working further down, Will look at this again

PR #271, opened 2022-06-20
Add two functions for dealing with financial year intervals

Moohan:
I've added and am using these as part of GP OoH's, but thought it would be simple / quicker for me to separate them out for review now whilst I finish the rest of that code.

One function creates a lubridate interval from a year. The other function takes a date and a year and checks if the date falls within the year (using the interval function and lubridate).


catherineholland1 (approved):
all looks good

PR #274, opened 2022-06-22
Pull out functions used in CH and HC 

catherineholland1:
pulling out these functions from year specific CH and HC to get documentation in line with my review


Moohan (approved):
Looks good to me

PR #277, opened 2022-06-24
Deal with gender when we have no CHI number.

### LINKED ISSUE #276: Issue with gender being 1.5 in individual file
From @Ryan-Harris93

>Just a quick point on the SLFs as Gaby was working with the files earlier this morning.

>When Gaby was completing the translation for the population & pathways matrix earlier today she spotted an issue with the gender classification in the older source individual files. I have pasted in a frequency example below for 2016/17
 
image

>From what Gaby explained to me earlier, she mentioned that there were some gender codes listed as 1.5, which SPSS rounds to 2, shown in the frequency table above, but not defined as male or female as the actual value isn’t 1 or 2, hence the separate frequency. However in R, these numbers are not rounded, so they remained as 1.5 during the frequency check, which is where she spotted the difference when comparing against each other after an aggregate.

>We know this is the case for 2016/17 and 2015/16 so maybe worth checking for years prior to this, granted it’s a small number but I assume it’s still worthwhile changing. Everything checks out fine from 2017/18 onwards.

Moohan:
Fixes #276 

There is a very small chance that we have someone who was legitimately coded with `0` who we now switch to `9`. We'd need a more substantial change to identify people where **all** of their codes were `0` before the aggregate, then we could retrieve this later.

See the data dictionary for definitions: [https://www.ndc.scot.nhs.uk/Dictionary-A-Z/Definitions/index.asp?Search=S&ID=1277&Title=Sex]()

### DIFF
--- a/2014-15/D01 Make Individual File.sps
+++ b/2014-15/D01 Make Individual File.sps
@@ -575,15 +575,19 @@ save outfile = !Year_dir + "temp-source-individual-file-1-20" + !FY + ".zsav"
     /zcompressed.
 get file = !Year_dir + "temp-source-individual-file-1-20" + !FY + ".zsav".
 
-* Clean up the gender, use the most common (by rounding the mean), if the mean is 1.5 (i.e. no gender known or equal male and females) then take it from the CHI).
+* Clean up the gender, use the most common (by rounding the mean).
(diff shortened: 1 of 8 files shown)


Moohan:
@Jennit07 I realised that this is probably wrong as the individual file shouldn't have any missing CHI numbers. It will need looking into in more detail!

PR #199, opened 2022-04-28
GPOoh r conversion

Jennit07:
I think this is almost ready. Definitely needs more testing/formatting but i thought i would submit the PR as i can't figure out some parts and wonder if you had any suggestions on how to write some of the code. I started to join the scripts together but then got a bit confused while testing so i've kept these separate for now and will put them into one script once finalised.

One part i am not sure of is line 103 in `extract_GPOoH_Diagnosis` in SPSS we take the first position of `.` and assign this to `diag1` to `diag6` but when i use `stringr::str_locate` i think this returns a vector which then doesn't let me do a pivot wider to create diag1 to diag6. Any thoughts on how to do this differently/improvements? This is the error returned when trying to do pivot wider:

`Error: Can't recycle ..1 (size 10) to match ..2 (size 5).
Run rlang::last_error() to see where the error occurred.`

 Still to do on this PR: 
- [ ] Create diag1 to diag6 variables with the read codes assigned accordingly by position of `.` 
- [ ] Any improvements to code where this can be simplified/run faster
- [ ] Join the 3 scripts together into one large script (kept this separate while testing)
- [ ] remove code where this writes to disk as we will store temp files as an object 
- [ ] check if there is a simpler way of doing things in R compared to SPSS code 

Closes #91

PR #278, opened 2022-06-24
Add the GP Out of Hours extract processing scripts

Moohan:
@Jennit07 did most of the initial work. I've gone through and corrected some bits, I'm up to the bit flagged as WIP (Work in Progress). Feel free to pick this up and add to / finish it, otherwise, I'll finish it when I'm back. 

Closes #199


Moohan:
Need to add changes from #302 / #358

PR #280, opened 2022-07-04
Remove typo

Jennit07:
Fixes a bug in the documentation where this would fail. There was a typo with an extra '#' which was appearing in the namespace. Check now passes. One note - however no binding for where() function. Not sure how to stop this as tidyselect::where() does not work.


catherineholland1 (approved):
looks good, happy to merge

PR #283, opened 2022-07-05
Homelessness tests

shintoLamp:
Created `extract_homelessness_tests.R`, incorporating a new function `produce_slf_homelessness_tests.R`

The tests create a table that compares the number of records between the existing data and the new extract. The reason the regular tests have not been used is that the extract does not have measures like bed days or cost.


Production_scripts/Process_extracts/Produce_extract_tests/extract_homelessness_tests.R
```diff
@@ -0,0 +1,43 @@
...
+year <- "1920"
+
+# Read new data file
+new_data <- readr::read_rds(get_source_extract_path(year, "Homelessness", ext = "rds")) %>%
+  mutate(smrtype = stringr::str_trim(smrtype, "both"))
+
+# Read current SLF episode file and filter for homelessness records (recid HL1)
+existing_data <- read_slf_episode(year, recid = "HL1") %>%
+  dplyr::rename(chi = anon_chi) %>%
+  dplyr::select(colnames(new_data))
```
Moohan, line 27: Use [`get_existing_data_for_tests()`](https://public-health-scotland.github.io/source-linkage-files/reference/get_existing_data_for_tests.html) here.

Jennit07, line 27:
```suggestion
existing_data <- get_existing_data_for_tests(new_data = new_data)
```

Production_scripts/Process_extracts/Produce_extract_tests/extract_homelessness_tests.R
```diff
@@ -0,0 +1,43 @@
...
+library(openxlsx)
+library(slfhelper)
+
+# Read in Data-----------------------------------------
+
+year <- "1920"
+
+# Read new data file
+new_data <- readr::read_rds(get_source_extract_path(year, "Homelessness", ext = "rds")) %>%
+  mutate(smrtype = stringr::str_trim(smrtype, "both"))
```
Jennit07, line 22:
```suggestion
new_data <- readr::read_rds(get_source_extract_path(year, "Homelessness", ext = "rds"))
```
Not needed for new version of tests to match SPSS output.

R/produce_slf_homelessness_tests.R
```diff
@@ -0,0 +1,45 @@
...
+    )) %>%
+    dplyr::group_by(hl1_sending_lca, smrtype, age) %>%
+    dplyr::summarise(records_old = dplyr::n())
+
+  comparison <- dplyr::full_join(new_data, existing_data) %>%
+    dplyr::mutate(
+      record_change = records_new - records_old,
+      record_proportion = abs((records_new - records_old) / records_old * 100)
+    ) %>%
+    dplyr::arrange(desc(record_proportion))
```
Jennit07, line 42:
```suggestion

```
Can remove this from the function and wrap the old and new dataframes in the function `produce_test_comparison`

Production_scripts/Process_extracts/Produce_extract_tests/extract_homelessness_tests.R
```diff
@@ -0,0 +1,43 @@
...
+  dplyr::rename(chi = anon_chi) %>%
+  dplyr::select(colnames(new_data))
+
+# Produce comparison-------------------------------------
+# Compare new file with existing slf data
+
+comparison <- produce_slf_homelessness_tests(
+  new_data = new_data,
+  existing_data = existing_data
+)
```
Jennit07, line 35:
```suggestion
comparison <- produce_test_comparison(
  produce_source_extract_tests(existing_data),
  produce_source_extract_tests(new_data)
)
Can wrap the old and new dataframes in the `produce_test_comparison` function. This example is from Acute tests. Will need tweaking to reflect homelessness variables and use updated function with `sum_mean_vars` parameters
```

R/produce_slf_homelessness_tests.R
```diff
@@ -0,0 +1,45 @@
...
+    dplyr::mutate(age = dplyr::case_when(
+      age %in% c(-1:17) ~ "<18",
+      age %in% c(18:44) ~ "18-45",
+      age %in% c(45:64) ~ "45-64",
+      age %in% c(65:84) ~ "65-84",
+      age >= 85 ~ "85+",
+      TRUE ~ "Missing"
+    )) %>%
+    dplyr::group_by(hl1_sending_lca, smrtype, age) %>%
+    dplyr::summarise(records_old = dplyr::n())
```
Jennit07, line 35:
```suggestion

```
We have talked through the other tests logic and you were going to go back and put it in here. You also were going to create a new test flag function called `create_lca_test_flags` for capturing `HL1_sending_lca`

R/produce_slf_homelessness_tests.R
```diff
@@ -0,0 +1,45 @@
...
+#' @param new_data The new data for testing
+#' @param existing_data The existing data for testing
+#'
+#' @description Produce the tests for the SLF Homelessness Extract
+#'
+#' @return A dataframe with counts of records, their differences, and proportional differences
+#'
+#' @export
+#' @family slf test functions
+produce_slf_homelessness_tests <- function(new_data, existing_data) {
```
Jennit07, line 12:
```suggestion
produce_slf_homelessness_tests <- function(data,
                                         sum_mean_vars = c("beddays", "cost", "yearstay"),
                                         max_min_vars = c(
                                           "record_keydate1", "record_keydate2",
                                           "cost_total_net", "yearstay"
                                         )) {
  test_flags <- data %>%
    # use functions to create HB and partnership flags
    create_demog_test_flags() %>%
    create_hb_test_flags(.data$hbtreatcode) %>%
    create_hb_cost_test_flags(.data$hbtreatcode, .data$cost_total_net) %>%
    # keep variables for comparison
    dplyr::select(c(.data$valid_chi:.data$NHS_Lanarkshire_cost)) %>%
    # use function to sum new test flags
    calculate_measures(measure = "sum")

  all_measures <- data %>%
    calculate_measures(vars = {{ sum_mean_vars }}, measure = "all")

  min_max <- data %>%
    calculate_measures(vars = {{ max_min_vars }}, measure = "min-max")

  join_output <- list(
    test_flags,
    all_measures,
    min_max
  ) %>%
    purrr::reduce(dplyr::full_join, by = c("measure", "value"))

  return(join_output)
}

```
This type of format for the new homelessness tests function

Jennit07 (changes requested):
Spoke through the changes and need to go back and follow the same format as other tests

Production_scripts/Process_extracts/Produce_extract_tests/extract_homelessness_tests.R
```diff
@@ -0,0 +1,41 @@
...
+# Original Date - July 2022
+# Written/run on - RStudio Server
+# Version of R - 3.6.1
+# Description - Produce tests for source linkage files:
+#               homelessness processed file.
+#####################################################
+
+# Load packages
+library(createslf)
+library(openxlsx)
```
lizihao-anu, line 13:
```suggestion

```

Production_scripts/Process_extracts/Produce_extract_tests/extract_homelessness_tests.R
```diff
@@ -0,0 +1,41 @@
...
+library(slfhelper)
```
lizihao-anu, line 14:
```suggestion

```

Production_scripts/Process_extracts/Produce_extract_tests/extract_homelessness_tests.R
shintoLamp: Changed to fit this

Production_scripts/Process_extracts/Produce_extract_tests/extract_homelessness_tests.R
shintoLamp: Changed this line

Production_scripts/Process_extracts/Produce_extract_tests/extract_homelessness_tests.R
shintoLamp: Done, other things changed as mentioned with the `produce_test_comparison` function

R/produce_slf_homelessness_tests.R
shintoLamp: Changed all of this to work as other tests do, and created `create_lca_test_flags`

R/create_lca_test_flags.R
```diff
@@ -0,0 +1,47 @@
...
+#' @param data the data containing an LCA variable
+#' @param lca_var LCA variable e.g. CA2019, CA2011
+#'
+#' @return a dataframe with flag (1 or 0) for each LCA
+#' @export
+#' @family flag functions
+create_lca_test_flags <- function(data, lca_var) {
+  data <- data %>%
+    dplyr::mutate(
+      Aberdeen_City = dplyr::if_else({{ lca_var }} == "S12000033", 1, 0),
```
Jennit07, line 14:
```suggestion
      Aberdeen_City = dplyr::if_else({{ lca_var }} == "S12000033" | {{ lca_var }} == "01", 1, 0),
```
This function looks great! 

I've had a thought where this may not work/apply to other lca variables. I'm not too sure if this is applicable to any dataset tests but these codes would only apply to the variable `HL1_sending_lca`. In the suggested output LCA codes in the SLFs are coded "01", "02" etc. Perhaps its worth adding in an `OR` statement so this could be picked up? I'm not sure if its worth doing this so we would need to check if any other tests use LCA. If no other tests use LCA flags, then i think the function is perfect the way it is. From the top of my head maybe in some SC tests we use LCA but that may also be a different code? e.g. `sc_sending_lca`?

Production_scripts/Process_extracts/Produce_extract_tests/extract_homelessness_tests.R
```diff
@@ -0,0 +1,39 @@
...
+# Load packages
+library(createslf)
+
+# Read in Data-----------------------------------------
+
+year <- "1920"
+
+# Read new data file
+new_data <- readr::read_rds(get_source_extract_path(year, "Homelessness", ext = "rds")) %>%
+  dplyr::mutate(smrtype = stringr::str_trim(smrtype, "both"))
```
Jennit07, line 20:
```suggestion
new_data <- readr::read_rds(get_source_extract_path(year, "Homelessness")) 
```
Suggested to remove `ext = "rds"`

Do we need the mutate for smrtype here? I think this was applied to your version of the tests, but i dont think we use smrtype flags for counting in the function

R/produce_slf_homelessness_tests.R
```diff
@@ -0,0 +1,43 @@
...
+                                           max_min_vars = c("record_keydate1", "record_keydate2")) {
+  test_flags <- data %>%
+    # Would usually use create_demog_test_flags() here but needs to be different for Homelessness
+    dplyr::arrange(.data$chi) %>%
+    # create test flags
+    dplyr::mutate(
+      has_chi = dplyr::if_else(!is_missing(.data$chi), 1, 0),
+      n_males = dplyr::if_else(.data$gender == 1, 1, 0),
+      n_females = dplyr::if_else(.data$gender == 2, 1, 0),
+      missing_dob = dplyr::if_else(is.na(.data$dob), 1, 0),
```
Jennit07, line 22: Suggestion to use `create_demog_test_flags` here. We spoke about this at the R conversion meeting. Just highlighting as a reminder when you have the time to go back to this :)

Jennit07 (changes requested):
Looks great/working as expected! Just a few minor suggestions

R/create_lca_test_flags.R
shintoLamp: I had a look at the SC tests but they don't do the same thing, they just look for a missing or not missing sending location as far as I can tell. We could update the sc tests to break down by lca, in which case I can change this code to account for other lca code types, or we can leave both as is

Production_scripts/Process_extracts/Produce_extract_tests/extract_homelessness_tests.R
shintoLamp: Suggestion implemented

R/produce_slf_homelessness_tests.R
shintoLamp: Used `create_demog_test_flags` here instead

R/create_lca_test_flags.R
Moohan: > I think the SC tests should probably break down by sending LA (can do that next/separately , no need to include it in this PR).
> 
> We can use `%in%` to get more streamlined code: `Aberdeen_City = dplyr::if_else({{ lca_var }} %in% c("S12000033", "01")`
> 
> Also (maybe a wider point) but is the `if_else` actually needed here? If we're just summing the `1/0` then that's the same as summing `TRUE/FALSE` and that can be achieved without the `if_else` which again cuts down some code:
> 
> ```r
> dplyr::mutate(
>       Aberdeen_City = {{ lca_var }} %in% c("S12000033", "01"),
>       ...
>       )
> ```
> 
> Don't forget to add any 'old' codes to the list too: [Boundary changes lookup.xlsx](https://github.com/Public-Health-Scotland/source-linkage-files/files/9208708/Boundary.changes.lookup.xlsx) that just protects us against some datasets using older codes - for the test we're not interested in which code is being used, we're just interested in properly grouping things into LAs so we can see that breakdown.

Just quoting this here! Also I made an issue for the SC test changes (#320)

Moohan:
> I think the SC tests should probably break down by sending LA (can do that next/separately , no need to include it in this PR).
> 
> We can use `%in%` to get more streamlined code: `Aberdeen_City = dplyr::if_else({{ lca_var }} %in% c("S12000033", "01")`
> 
> Also (maybe a wider point) but is the `if_else` actually needed here? If we're just summing the `1/0` then that's the same as summing `TRUE/FALSE` and that can be achieved without the `if_else` which again cuts down some code:
> 
> ```r
> dplyr::mutate(
>       Aberdeen_City = {{ lca_var }} %in% c("S12000033", "01"),
>       ...
>       )
> ```
> 
> Don't forget to add any 'old' codes to the list too: [Boundary changes lookup.xlsx](https://github.com/Public-Health-Scotland/source-linkage-files/files/9208708/Boundary.changes.lookup.xlsx) that just protects us against some datasets using older codes - for the test we're not interested in which code is being used, we're just interested in properly grouping things into LAs so we can see that breakdown.

Just quoting this here! Also I made an issue for the SC test changes (#320)

Jennit07:
Looks good! Happy to merge :)

Jennit07 (approved):

PR #286, opened 2022-07-08
Check scripts

### LINKED ISSUE #284: Check scripts specify check_mode = "write"
Found an issue where the MH script was not creating the rds file. This was due to the script missing the check_mode = "write" parameter. I think this may have occurred with two previous PRs which were merging at the same time - #214 and #210 which has overwrote one of the changes to include check_mode as parameter

Jennit07:
Relates to the issue where the MH extract was not writing to file. There must have been some overwrite to the code but now i think this captures everything available and now specifies `check_mode = "write"`  and `compress = "zsav" or "xz" (for RDS)` when saving extracts. Closes #284


Moohan (changes requested):
`createslf::write_sav` and `createslf::write_rds` compress by default, so there's no need to specify the compression parameter, on any of the function calls.

Moohan (approved):
Looks good!

PR #287, opened 2022-07-11
PIS extract tests

### LINKED ISSUE #179: Convert prescribing tests script
New branch for prescribing tests. Go back and amend this

Jennit07:
I've tested this and think it is working as expected. Draws upon existing tests logic but creates a new function called `produce_source_pis_tests` for selecting out functions specific to SPSS PIS tests. There are less variables to measure on here and we may develop another way of producing tests as this is very repetitive so the logic could be further developed. For example, we can use `produce_source_extract_tests` for processing tests for Acute, MH, Maternity, A&E and Outpatients but this won't work for PIS as there are more test flag functions used for the other datasets and only one of them used for the pis tests. For now this is working and producing an output we would expect. Closes #179


Production_scripts/Process_extracts/Produce_extract_tests/extract_PIS_tests.R
```diff
@@ -0,0 +1,46 @@
...
+# Produce comparison-------------------------------------
+# Compare new file with existing slf data
+comparison <- produce_test_comparison(
+  produce_source_pis_tests(existing_data,
+    sum_mean_vars = c("cost", "no_dispensed_items")
+  ),
+  produce_source_pis_tests(new_data,
+    sum_mean_vars = c("cost", "no_dispensed_items")
+  )
+)
```
Moohan, line 37: Since this is a custom function for PIS, I think we don't need to specify the variables like this, and can just 'hard-code' them into the `produce_source_pis_tests` function. In fact these are already the default, so this will work as is.

```suggestion
comparison <- produce_test_comparison(
  produce_source_pis_tests(existing_data),
  produce_source_pis_tests(new_data)
)
```

R/produce_source_pis_tests.R
```diff
@@ -0,0 +1,42 @@
...
+    calculate_measures(vars = {{ sum_mean_vars }}, measure = "all")
+
+  join_output <- list(
+    test_flags,
+    all_measures
+  ) %>%
+    purrr::reduce(dplyr::full_join, by = c("measure", "value"))
+
+  return(join_output)
+}
```
Moohan, line 42: See the comment re: 
```r
  produce_source_pis_tests(existing_data),
  produce_source_pis_tests(new_data)
)
```
```suggestion
#' Source PIS Tests
#'
#' @description Produce a set of tests which can be used by most
#' of the extracts.
#' This will produce counts of various demographics
#' using [create_demog_test_flags()] counts of episodes for every `hbtreatcode`
#' It will also produce various summary statistics for `cost` and `no_dispensed_items`
#'
#' @param data new or old data for testing summary flags
#' (data is from [get_source_extract_path()])
#'
#' @return a dataframe with a count of each flag
#' from [calculate_measures()]
#' @export
#'
#' @family extract test functions
#' for creating test flags
#' @seealso calculate_measures
produce_source_pis_tests <- function(data) {
  test_flags <- data %>%
    # use functions to create HB and partnership flags
    create_demog_test_flags() %>%
    dplyr::mutate(n_episodes = 1) %>%
    # keep variables for comparison
    dplyr::select(c(.data$valid_chi:.data$n_episodes)) %>%
    # use function to sum new test flags
    calculate_measures(measure = "sum")

  all_measures <- data %>%
    calculate_measures(vars = c("cost", "no_dispensed_items"), measure = "all")

  join_output <- list(
    test_flags,
    all_measures
  ) %>%
    purrr::reduce(dplyr::full_join, by = c("measure", "value"))

  return(join_output)
}
```

Moohan (changes requested):

Moohan (approved):

Moohan (changes requested):
See the R-CMD check note: `sum_mean_vars` needs to be documented in the `produce_source_pis_tests` function.

Jennit07:
> See the R-CMD check note: `sum_mean_vars` needs to be documented in the `produce_source_pis_tests` function.

Removed this from documentation as we do not need this as a parameter for PIS tests now - we hardcode the variables instead for this function

Moohan (approved):

PR #289, opened 2022-07-12
Fix write tests function

### LINKED ISSUE #288: Write tests function fails
Issue spotted by @lizihao-anu with `write_tests_xlsx`. Basically, it tries to change the permissions on the file *every* time, but we only need to change them once (when it's created). This actually fails with an error as only the person who created the file (the owner, currently @Jennit07) is allowed to change the permissions.

Moohan:
Fixes #288 

The main change is to only try and change the permissions when creating the file for the first time (i.e. if it doesn't already exist). Other minor changes are to improve comments and documentation.


lizihao-anu:
Looks good to me.

Jennit07 (approved):
This looks good to me. Improved comments are much clearer and makes sense to have the permissions set nearer the start than after the save! Ready to merge

PR #249, opened 2022-06-06
#211 Year Specific Source Home Care R

catherineholland1:
Closes #231 

Conversion of the Year Specific Home Care Extract

Includes a couple of functions created in the Care Home PR #223 - I have copied and pasted into this branch to run. 

I have also changed the latest update to `Jun_2022` to get the new costs variables read in and updated into the code

Package reference index updated also

PR #290, opened 2022-07-13
Fix some bugs in the main home care code

Moohan:
The main fix was to the write out bit, I needed the rds version for #249.

I also removed some other code which isn't needed.


Jennit07 (approved):
This looks good to me! Happy to merge :)

PR #292, opened 2022-07-14
Bug/ch costs

Moohan:
Main issue resolved here is the final save out.


Jennit07 (changes requested):
One query, but other than that i think this looks fine!

PR #293, opened 2022-07-14
Bug fix - cost variables should be '_cost' not '_costs'

Moohan:
Fix a bug / typo where the new cost variables being created are named as '_costs' (with an s) whereas they should be just '_cost'.


Jennit07 (approved):
Checked through some process scripts to make sure it was "_cost" when saving the outfile as i remember having this problem in the code. Everything looks like it checks out okay. Happy to merge

PR #294, opened 2022-07-14
Update convert_year_to_fyyear to handle vectors of years

Moohan:
I also added an equivalent test for this.


R/check_year_format.R
```diff
@@ -27,18 +27,18 @@ check_year_format <- function(year, format = "fyyear") {
 
   if (format == "fyyear") {
     if (any(as.integer(first_part) + 1L != as.integer(second_part))) {
```
shintoLamp, line 29: I noticed that if you give a vector like `c("1819", NA, "1920")` the function will abort but won't print a specific error message acknowledging that

shintoLamp (commented):
Just worth considering a situation where the vector of years given (in either function) has an NA value

R/check_year_format.R
Moohan: I think in this situation it should just throw an error?

Could modify `check_year_type` to do that?

Moohan: I think I've addressed this now.

I also noticed some tests were missing from one of the functions, so I added those and then had to write code to pass them properly!

shintoLamp (commented):
This all looks good to go now

PR #297, opened 2022-07-15
Format tests

Moohan:
This PR adds formatting to the test outputs which should make them easier to read and check.


Jennit07 (approved):
Looks good. We talked through this and found a bug where the permissions were being overwritten when we had a line of code to change this near the top. This causes files to reset back to permissions where this is read only for the group but read/write access for the owner. Now resolved and ready to merge.

PR #305, opened 2022-07-19
Expand is date in year

Moohan:
The original function checked a single date was in the financial year, it can now also check if a date range falls in a financial year.


R/is_date_in_fyyear.R
```diff
@@ -1,17 +1,37 @@
...
+#' is_date_in_year(
+#'   date = as.Date("2020-01-01"),
+#'   date_end = as.Date("2023-01-01"),
+#'   year = "2122"
+#' )
 #'
 #' @family date functions
-is_date_in_year <- function(date, year) {
-  lubridate::`%within%`(date, fy_interval(year))
+is_date_in_year <- function(date, year, date_end = NULL) {
```
Jennit07, line 25:
```suggestion
is_date_in_fyear <- function(date, year, date_end = NULL) {
```
Name suggestion to match the script

Moohan: I've updated the function name and parameter name as you suggested.

I also re-ordered the parameters as I didn't like having the `year` param in between the two dates but I also wanted the optional param to be last so that you can use the function without having to specify parameter names.

Moohan:
@Jennit07 and I had a chat about this (and other things) we tested this function as it is to be used on the Care Home extract. The last few commits were a result of the chat.

R/is_date_in_fyyear.R
```diff
@@ -1,17 +1,73 @@
...
+      cli::cli_abort(c("{.var date_end} must not be earlier than {.var date}",
+        "i" = "See case {first_error} where
+                         {.var date} = '{date[first_error]}'and
+                         {.var date_end} = '{date_end[first_error]}'",
+        "There {?is/are} {sum(date > date_end, na.rm = TRUE)} error{?s} in total."
+      ))
+    }
+  }
+
+  if (is.null(date_end)) {
```
Moohan, line 59:
```suggestion
  if (is.missing(date_end)) {
```

R/is_date_in_fyyear.R
```diff
@@ -1,17 +1,73 @@
...
+      cli::cli_abort(c("{.var date_end} must not be earlier than {.var date}",
+        "i" = "See case {first_error} where
+                         {.var date} = '{date[first_error]}'and
+                         {.var date_end} = '{date_end[first_error]}'",
+        "There {?is/are} {sum(date > date_end, na.rm = TRUE)} error{?s} in total."
+      ))
+    }
+  }
+
+  if (is.missing(date_end)) {
```
shintoLamp, line 59:
```suggestion
  if (missing(date_end)) {
```

shintoLamp (commented):
Ran all the tests and everything comes out fine, tests seem to cover all scenarios so I would say this is finished pending that small change I noted

PR #310, opened 2022-07-20
308 update create monthly beddays

### LINKED ISSUE #308: Update `create_monthly_beddays`
We were discussing something else when we came across this possible bug in the `create_monthly_beddays` function. 
Agreed that James would look at this to update function to account for where the end date is `NA`

Moohan:
Closes #308 

Expands `create_monthly_beddays` to handle missing (`NA`) values for the discharge date.


lizihao-anu (approved):
It looks good and has passed all the test. So happy to merge

PR #315, opened 2022-07-26
313 problem with simd lookup

### LINKED ISSUE #313: Problem with SIMD lookup
The `read_simd_file()` function doesn't currently work, as it is not generalised to pick up the latest version. Use the `read_spd_file()` function as a reference.

Should also be renamed with `get_` as a prefix

Moohan:
Closes #313 

- Functions renamed to be of the form `get_x_path` to be consistent with others and be more descriptive of what they return.
- Tidied up the function `find_latest_file`.
- Expanded `get_file_path` so now it can take a regexp instead of a file name, if so it should use `find_latest_file` to look for the latest file.
- Made all the functions more consistent and tidied up the documentation.
- Renamed the file where the functions live to be more consistent

TODO (not necessarily to complete this PR):
- [ ] Add some tests for the lookup file-path functions (#318).
- [ ] Expand (and rename) `get_datazone_pop_path` so it can read any 'type' of pop estimate (#319).


lizihao-anu (approved):
Looks good

PR #316, opened 2022-07-26
Fix permissions on `write_*` functions

Moohan:
I was getting an error again here, by the same logic we used in `write_xlsx_tests` we should only try and change the permissions if we're the owner i.e. we just create the file.

I think the error message was being confused by the way we pipe the data to `write_zsav` followed by `write_rds`, and how R is lazily executing that pipe (it goes inside out). If you run the `write_*` bits seperately the error message is clearer.

This will need 2 people to test it properly!

### DIFF
--- a/R/write_data.R
+++ b/R/write_data.R
@@ -28,7 +28,10 @@ write_sav <- function(data, path, compress = "zsav") {
     compress = compress
   )
 
-  fs::file_chmod(path = path, mode = "660")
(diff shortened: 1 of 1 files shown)


shintoLamp:
Checked with Jen, I wrote two files and Jen was able to overwrite them. I was then able to open Jen's files and subsequently overwrite them, so I think we can be confident this works as expected

shintoLamp (approved):
Tested and approved

PR #323, opened 2022-08-03
LTC IT extract tests

shintoLamp:
Wrote up tests for the LTC extract provided by IT. SPSS tests just returned if there were duplicates, but I have expanded this to also look for situations where a person has an LTC but doesn't have a valid date associated with it. 

Code can definitely be tidied up but I wanted to upload this as a proof of concept sort of thing, if the output is considered good then I can trim it down.


R/get_ltcs_path.R
```diff
@@ -12,7 +12,7 @@
 get_ltcs_path <- function(year, ...) {
   ltcs_file_path <- get_file_path(
     directory = fs::path(get_slf_dir(), "LTCs"),
-    file_name = glue::glue("LTC_patient_reference_file-20{year}.rds"),
+    file_name = glue::glue("LTCs_patient_reference_file-20{year}.zsav"),
```
Jennit07, line 15: I think this needs to stay as `.rds`. However, if there is no `.rds` file available in the folder, you can either run this (im sure the ltc extract has been converted) or alternatively when reading the file you can specify `ext = .zsav`.

Production_scripts/Process_extracts/Produce_extract_tests/extract_chi_ltc_tests.R
```diff
@@ -0,0 +1,93 @@
+####################################################
+# Name of file - extract_chi_ltc_tests.R
+# Original Authors - Bateman McBride
+# Original Date - August 2022
+# Written/run on - RStudio Server
+# Version of R - 3.6.1
+# Description - Produce tests for source linkage files:
+#               LTC file received from IT.
+#####################################################
+
```
Jennit07, line 10:
```suggestion
# Load packages 

library(dplyr)
```
Because you have used dplyr a lot here, i think we can load this at the beginning to trim down the code

Production_scripts/Process_extracts/Produce_extract_tests/extract_chi_ltc_tests.R
```diff
@@ -0,0 +1,93 @@
...
+    congen_invalid = congen - congen_valid,
+    bloodbfo_invalid = bloodbfo - bloodbfo_valid,
+    endomet_invalid = endomet - endomet_valid,
+    digestive_invalid = digestive - digestive_valid
+  ) %>%
+  tidyr::pivot_longer(
+    cols = tidyselect::everything(),
+    names_to = "measure",
+    values_to = "value"
+  )
```
Jennit07, line 85: While what you have here is good/working as we would expect to return a meaningful output i'm not sure we will need to include this as a test. Perhaps just the invalid dates would be sufficient as a check for our code. From looking at the processing for the LTC script, we actually derive the LTC flags from the dates and they are provided by IT so in theory we should never have a situation where a person has an LTC flag but does not have a date associated with it.

Production_scripts/Process_extracts/Produce_extract_tests/extract_chi_ltc_tests.R
```diff
@@ -0,0 +1,93 @@
...
+  dplyr::mutate(duplicate_chi_postcode = dplyr::n() > 1) %>%
+  dplyr::ungroup() %>%
+  dplyr::summarise(dplyr::across(c("duplicate_chi", "duplicate_chi_postcode"), sum)) %>%
+  dplyr::ungroup() %>%
+  tibble::as_tibble() %>%
+  tidyr::pivot_longer(
+    cols = tidyselect::everything(),
+    names_to = "measure",
+    values_to = "value"
+  )
```
Jennit07, line 31: This is good and working as expected. Thanks for showing us how to use `dtplyr::lazy_dt()` we have been able to implement the use of this function in the social care datasets and this will also be very useful going forward when the code gets more complicated :)

Only comment - is there a way to trim this down without having to do multiple group() and ungroup()? Just for improving the readability of the code

Jennit07 (dismissed):
A few comments

Production_scripts/Process_extracts/Produce_extract_tests/extract_chi_ltc_tests.R
```diff
@@ -0,0 +1,93 @@
...
+# Name of file - extract_chi_ltc_tests.R
+# Original Authors - Bateman McBride
+# Original Date - August 2022
+# Written/run on - RStudio Server
+# Version of R - 3.6.1
+# Description - Produce tests for source linkage files:
+#               LTC file received from IT.
+#####################################################
+
+year <- "1920"
```
Moohan, line 11:
```suggestion
year <- check_year_format("1920")
```

Production_scripts/Process_extracts/Produce_extract_tests/extract_chi_ltc_tests.R
```diff
@@ -0,0 +1,93 @@
...
+
+new_data <- haven::read_sav(get_ltcs_path(year))
+
+# Find and flag any duplicate chis and chi/postcode combinations
+duplicates <- new_data %>%
+  dtplyr::lazy_dt() %>%
```
Moohan, line 17: You need to run `usethis::use_package("dtplyr")` to make sure `dtplyr` becomes an import.

Production_scripts/Process_extracts/Produce_extract_tests/extract_chi_ltc_tests.R
```diff
@@ -0,0 +1,93 @@
...
+  dplyr::group_by(chi) %>%
+  dplyr::mutate(duplicate_chi = dplyr::n() > 1) %>%
+  dplyr::ungroup() %>%
+  dplyr::group_by(chi, postcode) %>%
+  dplyr::mutate(duplicate_chi_postcode = dplyr::n() > 1) %>%
```
Moohan, line 22: I'm not sure what `duplicate_chi_postcode` is / would tell us? If there's a duplicate CHI, that's an issue, I'm not sure if we care if those duplicate CHIs have the same postcode though?

Production_scripts/Process_extracts/Produce_extract_tests/extract_chi_ltc_tests.R
```diff
@@ -0,0 +1,93 @@
...
+  dplyr::ungroup() %>%
+  dplyr::summarise(dplyr::across(c("duplicate_chi", "duplicate_chi_postcode"), sum)) %>%
+  dplyr::ungroup() %>%
```
Moohan, line 25: Not needed as we were already ungrouped before the `summarise`
```suggestion
```

Production_scripts/Process_extracts/Produce_extract_tests/extract_chi_ltc_tests.R
```diff
@@ -0,0 +1,93 @@
...
+# Find and flag any duplicate chis and chi/postcode combinations
+duplicates <- new_data %>%
+  dtplyr::lazy_dt() %>%
+  dplyr::group_by(chi) %>%
+  dplyr::mutate(duplicate_chi = dplyr::n() > 1) %>%
+  dplyr::ungroup() %>%
+  dplyr::group_by(chi, postcode) %>%
+  dplyr::mutate(duplicate_chi_postcode = dplyr::n() > 1) %>%
+  dplyr::ungroup() %>%
+  dplyr::summarise(dplyr::across(c("duplicate_chi", "duplicate_chi_postcode"), sum)) %>%
```
Moohan, line 24: I think we can do this whole thing without any grouping (assuming we drop the postcode bit).
```suggestion
  dplyr::summarise(duplicate_chi = nrow(new_data) - dplyr::n_distinct(chi)) %>%
```

If we keep the postcode thing then some combination of a single grouping `n()` and `n_distinct()` should also work.

Production_scripts/Process_extracts/Produce_extract_tests/extract_chi_ltc_tests.R
Moohan: I don't think we need to load `dplyr` as we're using `dplyr::` throughout.

Moohan (changes requested):

Production_scripts/Process_extracts/Produce_extract_tests/extract_chi_ltc_tests.R
shintoLamp: I think with the changes James suggested, there's a lot fewer lines starting `dplyr::` so it's not as messy

Production_scripts/Process_extracts/Produce_extract_tests/extract_chi_ltc_tests.R
shintoLamp: Done

Production_scripts/Process_extracts/Produce_extract_tests/extract_chi_ltc_tests.R
shintoLamp: Changed the code to just check for duplicate CHI using this suggestion.

Production_scripts/Process_extracts/Produce_extract_tests/extract_chi_ltc_tests.R
shintoLamp: Removed this section entirely

R/get_ltcs_path.R
shintoLamp: Changed this back to what it was originally

Production_scripts/Process_extracts/Produce_extract_tests/extract_chi_ltc_tests.R
shintoLamp: I think there would have been, but the need to group repeatedly has been superseded by other suggestions

Production_scripts/Process_extracts/Produce_extract_tests/extract_chi_ltc_tests.R
```diff
@@ -0,0 +1,29 @@
...
+# Original Date - August 2022
+# Written/run on - RStudio Server
+# Version of R - 3.6.1
+# Description - Produce tests for source linkage files:
+#               LTC file received from IT.
+#####################################################
+
+year <- check_year_format("1920")
+
+new_data <- haven::read_sav(get_ltcs_path(year))
```
Moohan, line 13: Should this not be rds?

shintoLamp: Yes, it should. Can I change that here?

Moohan:
```suggestion
new_data <- readr::read_rds(get_ltcs_path(year))
```

Moohan (approved):
I ran through the code on an LTC extract, the read-in was super slow but not sure what we can do about that!

Issue #326, opened 2022-08-08
Fix documentation in `produce_source_nrs_tests`


Moohan:
There is a build warning that was caused by #306, where a parameter in the new test function wasn't properly documented. This is now causing the build check to fail on every PR which is trying to merge to `main-R` e.g. [The workflow notes for #278 ](https://github.com/Public-Health-Scotland/source-linkage-files/runs/7727970846?check_suite_focus=true#:~:text=%E2%9D%AF%20checking%20Rd%20%5Cusage,%E2%80%98max_min_vars%E2%80%99)

shintoLamp:
Merged fix for this issue in #327

PR #327, opened 2022-08-09
Removed documentation for parameter that no longer exists

shintoLamp:
Had roxygen reference to parameter that we hard-coded instead, in `produce_source_nrs_tests.R`. Removed this.


Moohan (approved):

Moohan:
Fixes #326

PR #328, opened 2022-08-09
CMH extract tests

shintoLamp:
Nothing special with this one, functionally the same as the NRS tests but with the variable `n_deaths` removed


Production_scripts/Process_extracts/extract_cmh.R
```diff
@@ -98,6 +98,7 @@ cmh_clean <- cmh_extract %>%
 
 outfile <- cmh_clean %>%
   select(
+    year,
```
shintoLamp, line 101: The script didn't select `year` before and I assumed this was an oversight as `year` was in the existing data

R/produce_source_cmh_tests.R
```diff
@@ -0,0 +1,37 @@
...
+#' from [calculate_measures()]
+#' @export
+#'
+#' @family extract test functions
+#' @seealso [calculate_measures()]
+produce_source_cmh_tests <- function(data) {
+  test_flags <- data %>%
+    # create test flags
+    create_demog_test_flags() %>%
+    create_hb_test_flags(hb_var = hbrescode) %>%
```
Jennit07, line 21:
```suggestion
    create_hb_test_flags(hb_var = hbrescode) %>%
    dplyr::mutate(n_episodes = 1) %>%
```
Only thing not captured by the tests here is n_episodes which is in the SPSS code

Jennit07 (changes requested):
Looks good! Just one minor comment, im not sure if the other tests have n_episodes but i noticed we had this in the nrs deaths and checked the cmh SPSS code here which had it. We could maybe even add this flag into the demographic test flags? or just remember to hard code in the scripts

Jennit07 (approved):

PR #349, opened 2022-08-11
Remove the duplicate GP cluster script

Moohan:
For some reason we had two scripts which were downloading, cleaning and saving the GP cluster data, they did things slightly differently so I've removed the stand-alone one and incorporated the differences into the other one.


Jennit07:
This is looking good to me and i agree that `lookup_gp_clusters.R` should be the one that is deleted/incorporated into `lookup_gp_practice.R`

Jennit07 (approved):
Ready to merge

PR #350, opened 2022-08-11
322 add new quarter date functions

### LINKED ISSUE #322: Add new quarter date functions
`yq` is working it out as the cal year. So 2017 Q1 is Jan-Mar 2017, whereas we want the FY year, so need to add 3 months (1 quarter) to it to make it correct. We then need to add another 3 months to get the end of the quarter.

Possibly worth making some functions for SC period dates? - Take SC `period` and return date e.g. `start_sc_quarter(period)` `end_sc_quarter(period)`. So in this case we'd use `end_sc_quarter(period) + lubridate::days(1)` (or equivalent)

_Originally posted by @Moohan in https://github.com/Public-Health-Scotland/source-linkage-files/pull/312#discussion_r934379053_

Moohan:
Closes #322 

Adds functions to get the dates from a character vector of quarters. Most useful for Social Care stuff as the quarter variable has to be in the form `2017Q1`.

It makes sense to split them out as then they can be checked and tested separately.

I've tried to make them as efficient as possible: to that end, they find the unique values, work out the dates for those then apply those 'answers' back to the original (potentially large) vector.


Jennit07:
I've had a look through this one and it's looking great to me. Tested this on the various different scenarios and happy it's doing as expected, all tests also pass

lizihao-anu (approved):

PR #314, opened 2022-07-26
Convert main (all data) Care Home extract 

Moohan:
Closed #85 and is a replacement PR for #208 which Catherine did most of the work on. Has been reviewed by me and Jen in part, then we did a group review of the code and made lots more fixes and changes.

I need to finalise the usage of the Care Home name lookup to tidy up the care home names and postcodes.

PR #361, opened 2022-08-19
359 change care home type of admission variable

Moohan:
Changes to Care Home to use the newer `type_of_admission` instead of the older `reason_for_admission` variable. Where an episode only has the older variable it does a rough re-coding.

I also removed some old / defunct code.


Moohan:
We will also need to make these changes in #314

Jennit07 (approved):
I've checked through the logic and think this makes sense to me. I've also ran this in SPSS and it looks like it is working with no issues. Happy to approve/merge

PR #362, opened 2022-08-19
Update 2021/22 sparra and HHG

Jennit07:
Hopefully i have made the correct changes, 22/23 has identical code to 21/22 previously and i have copied the code from 20/21 to 21/22. I tried testing this however get the error saying there are duplicates. I have tried taking the extracts and sorting them by UPI/CHI but not sure what is going wrong. I'm thinking it may be due to the code i have copied if i am missing something else to change? Could you please check this to ensure i have ran the correct code? we have the temp files in 2021/22 so i can easily test this again once you're happy with the code


Moohan (approved):
The code looks good to me, the issue was with the new SPARRA file. There were some (302) blank CHI numbers and a handful of duplicates.

We could remove the blank CHI rows ourselves and then pick the first row to deal with the duplicates but hopefully Katie / SPARRA team will get back with a fixed file early on Monday.

PR #365, opened 2022-08-23
Add smrtype variable to extracts

shintoLamp:
Created a function, `add_smr_type` that uses recid and mpat to figure out the smr type. I also created tests for this function. Then I added the function to the extract processing scripts for A&E, PIS, NRS Deaths, Maternity, Mental Health, and Outpatients.


Production_scripts/Process_extracts/extract_a&e.R
```diff
@@ -134,7 +134,9 @@ ae_clean <- ae_file %>%
   # Create month variable
   mutate(month = strftime(record_keydate1, "%m")) %>%
   # Allocate the costs to the correct month
-  create_day_episode_costs(record_keydate1, cost_total_net)
+  create_day_episode_costs(record_keydate1, cost_total_net) %>%
+  # Add SMR type
+  mutate(smrtype = add_smr_type(recid))
```
Jennit07, line 139:
```suggestion
  create_day_episode_costs(record_keydate1, cost_total_net),
  # Add SMR type
  smrtype = add_smr_type(recid)
  )
```
Since there is a mutate above, we can wrap this together.

Production_scripts/Process_extracts/extract_maternity.R
```diff
@@ -166,7 +166,9 @@ maternity_clean <- maternity_file %>%
     discondition = factor(discondition,
       levels = c(1:5, 8)
     )
-  )
+  ) %>%
+  # Add SMR type
+  mutate(smrtype = add_smr_type(recid, mpat))
```
Jennit07, line 171:
```suggestion
  ),
  # Add SMR type
  smrtype = add_smr_type(recid, mpat)
  )
```
Wrap mutate together

Production_scripts/Process_extracts/extract_mental_health.R
```diff
@@ -178,7 +178,9 @@ mh_clean <- mh_extract %>%
     cost_total_net = rowSums(across(ends_with("_cost"))) %>%
   # total length of stay
     stay = calculate_stay(year, record_keydate1, record_keydate2)
-  )
+  ) %>%
+  # Add SMR type
+  mutate(smrtype = add_smr_type(recid))
```
Jennit07, line 183: wrap mutate

Production_scripts/Process_extracts/extract_prescribing.R
```diff
@@ -60,7 +60,9 @@ pis_clean <- pis_file %>%
   mutate(
     record_keydate1 = end_fy(year),
     record_keydate2 = record_keydate1
-  )
+  ) %>%
+  # Add SMR type
+  mutate(smrtype = add_smr_type(recid))
```
Jennit07, line 65:
```suggestion
  ) ,
  # Add SMR type
 smrtype = add_smr_type(recid)
 )
```
wrap mutate

tests/testthat/test-add_smr_type.R
```diff
@@ -0,0 +1,53 @@
+# Single character input
```
Jennit07, line 1: Tests look great! :)

R/add_smr_type.R
```diff
@@ -0,0 +1,75 @@
...
+      recid == "02B" & mpat %in% c("2", "4", "6") ~ "Matern-DC",
+      recid == "02B" & mpat == "0" ~ "Matern-HB",
+      recid == "04B" ~ "Psych-IP",
+      recid == "00B" ~ "Outpatient",
+      recid == "AE2" ~ "A & E",
+      recid == "PIS" ~ "PIS",
+      recid == "NRS" ~ "NRS Deaths"
+    )
+  }
+
```
Jennit07, line 72: Function looks great!

R/add_smr_type.R
```diff
@@ -0,0 +1,75 @@
...
+#' @family Codes
+#'
+#' @examples
+#' x <- tibble::tibble(
+#'   recid = c("02B", "02B", "02B", "04B", "00B", "AE2", "PIS", "NRS"),
+#'   mpat = c("1", "2", "0", "", "", "", "", "")
+#' )
+#' add_smr_type(x)
+add_smr_type <- function(recid, mpat = NULL) {
+
```
Jennit07, line 18: Nice idea to put in some checks!

Jennit07 (changes requested):
Hey @shintoLamp function looks good, tests look good! Just one minor comment throughout to wrap the mutates since there is one above where you are adding in smrtype. Could you also please do a check using the build tab? There are some errors pulling through which i suspect are due to changes to the IT extract task number changing since we have the new extracts/any changes you made to this. Thanks :)

R/add_smr_type.R
```diff
@@ -0,0 +1,75 @@
...
+  # Situation where some recids are missing
+  if (any(is.na(recid)) & !all(is.na(recid))) {
+    cli::cli_inform(c("i" = "Some values of {.var recid} are {.val NA},
```
Moohan, line 21: Should this not be an error?

Would the function be able to continue (in a meaningful way) if there were missing `recid`s? In the SLFs we'd never expect to have a missing recid. I'd be tempted to change the check to: `if (any(is.na(recid)))` and then use `cli::cli_abort`

R/add_smr_type.R
```diff
@@ -0,0 +1,75 @@
...
+                    please check this is populated throughout the data"))
+  }
+
+  # Situation where some recids are not in the accepted values
+  if (any(!(recid %in% c("02B", "04B", "00B", "AE2", "PIS", "NRS"))) &
+    !any(is.na(recid))) {
+    cli::cli_inform(c("i" = "One or more values of {.var recid} do not have an
+                   assignable {.var smrtype}"))
```
Moohan, line 29: Should this be a warning, rather than just a message? What would be the scenario where some non-handled recids were submitted and what should we do in that case? I'd almost be tempted to make it an error even.

```suggestion
    cli::cli_warn(c("i" = "One or more values of {.var recid} do not have an
                   assignable {.var smrtype}"))
```

R/add_smr_type.R
```diff
@@ -0,0 +1,75 @@
...
+      recid == "AE2" ~ "A & E",
+      recid == "PIS" ~ "PIS",
+      recid == "NRS" ~ "NRS Deaths"
+    )
+  } else {
+    # Recode all recids
+    smrtype <- dplyr::case_when(
+      recid == "02B" & mpat %in% c("1", "3", "5", "7", "A") ~ "Matern-IP",
+      recid == "02B" & mpat %in% c("2", "4", "6") ~ "Matern-DC",
+      recid == "02B" & mpat == "0" ~ "Matern-HB",
```
Moohan, line 64: I think we need a check, either here or earlier for `any(recid) == "02B"` -> `mpat` can't be missing.

Moohan (changes requested):
Looking good, could you also roll this out to all the other process extract scripts, please? So we have this function as the single point of call for setting smrtype.

Note that the [code in acute](https://github.com/Public-Health-Scotland/source-linkage-files/blob/37bdc1f58a0d7550a1255a74da46da57196eee2a/Production_scripts/Process_extracts/extract_acute_incl_gls.R#L200) - Is no longer correct as we removed that complicated bit in the last update. This is [how the new logic](https://github.com/Public-Health-Scotland/source-linkage-files/blob/37bdc1f58a0d7550a1255a74da46da57196eee2a/2021-22/B01%20Process%20Acute%20(SMR01%20and%20GLS)%20extract.sps#L242) needs to be.

Production_scripts/Process_extracts/extract_a&e.R
shintoLamp: Wrapped

R/add_smr_type.R
shintoLamp: Changed this to use `cli_abort` when any `recid` is NA now

R/add_smr_type.R
shintoLamp: Changed this to `cli_warn`

R/add_smr_type.R
shintoLamp: There's a check here in line 32

` # Situation where maternity records are present without an mpat
  if (any(recid == "02B") & any(is.na(mpat))) {
    cli::cli_warn(c("i" = "In maternity records, {.var mpat} is required to assign
                      an smrtype, and there are some {.val NA} values. Please check the data."))
  }`

that covers this scenario. Could change it to `cli_abort` if you think that'd be more appropriate

R/add_smr_type.R
```diff
@@ -0,0 +1,104 @@
...
+      recid == "GLS" & ipdc == "I" ~ "GLS-IP"
+    )
+  }
+  # Home care
+  else if (recid == "HC" & !is.null(hc_service)) {
+    smrtype <- dplyr::case_when(
+      recid == "HC" & hc_service == "1" ~ "HC-Non-Per",
+      recid == "HC" & hc_service == "2" ~ "HC-Per",
+      recid == "HC" & !(hc_service %in% c("1", "2")) ~ "HC-Unknown"
+    )
```
Jennit07, line 99:
```suggestion
    smrtype <- dplyr::case_when(
      recid == "HC" & hc_service == "1" ~ "HC-Non-Per",
      recid == "HC" & hc_service == "2" ~ "HC-Per",
      TRUE ~ "HC-Unknown"
    )
```
This line can be re-written like this which means 'everything else assign HC-Unknown'. Other than this, i have tested these lines and the code works for HC. Something in the filter is preventing this from returning the correct values and returns NA instead, but to me the filter looks fine. I have also tried this with other scenarios e.g. when mpat and ipdc is null which also doesn't work

Jennit07, line 99:
```suggestion
  # Recids that can be recoded with no identifier
  if (is.null(mpat) & is.null(ipdc) & is.null(hc_service)) {
    smrtype <- dplyr::case_when(
      recid == "04B" ~ "Psych-IP",
      recid == "00B" ~ "Outpatient",
      recid == "AE2" ~ "A & E",
      recid == "PIS" ~ "PIS",
      recid == "NRS" ~ "NRS Deaths",
      recid == "CMH" ~ "Comm-MH",
      recid == "DN" ~ "DN"
    )
  }
  # Maternity recids, identifier is `mpat`
  else if (all(recid == "02B") & !is.null(mpat) & is.null(ipdc)) {
    smrtype <- dplyr::case_when(
      recid == "02B" & mpat %in% c("1", "3", "5", "7", "A") ~ "Matern-IP",
      recid == "02B" & mpat %in% c("2", "4", "6") ~ "Matern-DC",
      recid == "02B" & mpat == "0" ~ "Matern-HB"
    )
  }
  # Acute recids, identifier is `ipdc`
  else if (is.null(mpat) & !is.null(ipdc)) {
    smrtype <- dplyr::case_when(
      recid == "01B" & ipdc == "I" ~ "Acute-IP",
      recid == "01B" & ipdc == "D" ~ "Acute-DC",
      recid == "GLS" & ipdc == "I" ~ "GLS-IP"
    )
  }
  # Home care
  else if (!is.null(hc_service)) {
    smrtype <- dplyr::case_when(
      recid == "HC" & hc_service == "1" ~ "HC-Non-Per",
      recid == "HC" & hc_service == "2" ~ "HC-Per",
      TRUE ~ "HC-Unknown"
    )
```
I think i have found the solution. The first if statement did not specify hc_service = NULL so the function was getting confused with when hc_service is not NULL. These changes now allow me to return the correct smrtype when i run `add_smr_type(recid = "HC", hc_service = "1")` 

Could you please check this works for you? Thanks

shintoLamp: Aye, I has the `TRUE ~ "HC-Unknown` in there originally but thought it might be a problem. I've changed it back now!

shintoLamp: It does! Thanks for looking into this!

PR #366, opened 2022-08-23
Update the cost uplift macro

Moohan:
New numbers from Paul Leak for uplifting 2020/21, 2021/22 and 2022/23. Also uplift any future years beyond that by 1%.

I created a dummy dataset with a `recid` variable and tested with this code:
```
Define !latest_cost_year()
    2022
!EndDefine.

Define !fy()
    "2324"
!EndDefine.

!create_uplift_var.
execute.
```
Then tested changing the `fy` and the `latest_cost_year`


Jennit07 (approved):
I have tested this with dummy data as well and i think this is working as expected. Great work - happy to merge

PR #367, opened 2022-08-23
a quick test for get_slf_lookup_paths.R

lizihao-anu:
James has already written the most of the tests before, so I just finish the rest.


tests/testthat/test-get_slf_lookup_paths.R
```diff
@@ -1,5 +1,3 @@
-skip_on_ci()
-
```
Moohan, line 2: This needs to stay as this test can't be run on GitHub.

The `skip_on_ci()` means that this test file won't be run when in a CI (Continuous Integration) environment like GitHub actions.

Moohan (changes requested):

Jennit07 (approved):
looks good

PR #369, opened 2022-08-23
Remove the variable `sds_option` as it is recoded into `smrtype` 

Moohan:
The variable `sds_option` isn't needed any more as it is entirely encoded into `smrtype`.

SDS 4 was missed from `smrtype` so this has now been added and all references and uses of `sds_option` have been removed.


Jennit07 (approved):
Thanks for looking into this @Moohan, a bit of an oversight on my part but identified that option 4 was missed entirely! The code changes look good, only thing i have changed is removing SDS from 22/23 as this was copied over. Other than that this is good to merge

PR #370, opened 2022-08-24
Update SC function codes and names

Jennit07:
Problem with North lanarkshire not being coded properly for sc_send_lca variable. This fixes this and also adds in names for each code/ordering


Moohan (approved):
Looks good to me, all the LCAs are there now 😆

We do need to remember to copy these changes to the main-r branch.

Issue #373, opened 2022-08-26
Remove zsav from extract scripts


Jennit07:
We included zsav when saving the output as we intended to use these for updating the SLFs in SPSS. However this is no longer needed as we are progressing onto phase 2 and the structures will be reshaped for running the SLFS. We only need to keep rds versions now

PR #374, opened 2022-08-30
373 remove zsav

Jennit07:
Linked to #373. Need to remove the line of code which writes a zsav file in each of the extract processing scripts. I think i have captured all apart from homelessness where we do need to keep the zsav file as this is now removed from the SPSS process.


shintoLamp (approved):
Looked through and all the removals look good, I also used Edit -> Find in Files... within rStudio to see if there were any missed onces but couldn't see any, so I think this is fine to be merged.

PR #376, opened 2022-09-07
375 add year parameter

### LINKED ISSUE #375: Add year parameter to `write_tests_xlsx` function
An oversight where this works fine if you are running one year but if you run multiple years it is not attached to a year

Jennit07:
I've tested this running some test extracts and this will now put the year in the sheet name for each test. e.g. "2019_04B_07_Sep". I've also shortened the sheet names for all tests to use recid as i was getting an error saying this was too long. Closes #375

### DIFF
--- a/R/write_tests_xlsx.R
+++ b/R/write_tests_xlsx.R
@@ -14,7 +14,7 @@
 #'
 #' @family test functions
 #' @seealso produce_test_comparison
-write_tests_xlsx <- function(comparison_data, sheet_name) {
+write_tests_xlsx <- function(comparison_data, sheet_name, year) {
 
 
   # Set up the workbook -----------------------------------------------------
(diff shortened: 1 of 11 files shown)


R/write_tests_xlsx.R
```diff
@@ -34,7 +34,7 @@ write_tests_xlsx <- function(comparison_data, sheet_name) {
   }
 
   # add a new sheet for tests
-  sheet_name_dated <- paste0(sheet_name, format(Sys.Date(), "_%d_%b"))
+  sheet_name_dated <- paste0(year, "_", sheet_name, format(Sys.Date(), "_%d_%b"))
```
Moohan, line 37: I think glue makes this more readable.
```suggestion
  sheet_name_dated <- glue::glue("{year}_{sheet_name}_{format(Sys.Date(), '%d_%b')}")
```

Moohan (approved):

Moohan (changes requested):
I realised that you're not passing through the year parameter in any of the calls e.g. `write_tests_xlsx(comparison, "04B")` needs to be `write_tests_xlsx(comparison, "04B", year)`.

Moohan (approved):

PR #377, opened 2022-09-07
331 ppa in c01

shintoLamp:
First pass at this function, but I can confirm it is adding the PPA flag as expected. I tested it using the Source Episode file and compared my output to the final version, and the situations where my `ppa` variable is `TRUE` are the same situations where `cij_ppa` is 1. 

I'm not sure how I would go about writing effective `testthat` cases for this though, up to the team if that's required.


R/add_ppa_flag.R
```diff
@@ -0,0 +1,149 @@
...
+    dplyr::mutate(
+
+      # Excluding operations are op1a codes from K01 to K50, K56, K60, and K61 (dental)
+      excluding_operation = stringr::str_sub(.data$op1a, 1, 3) %in%
+        purrr::reduce(list(
+          glue::glue("K{stringr::str_c(\"0\", c(1:9))}"),
+          glue::glue("K{as.character(c(10:50))}"),
+          c("K56", "K60", "K61")
+        ), union),
+      ppa = dplyr::case_when(
```
Moohan, line 33:
```suggestion
        ), union),
        
      ppa = dplyr::case_when(
```

R/add_ppa_flag.R
```diff
@@ -0,0 +1,149 @@
...
+    dplyr::group_by(.data$anon_chi, .data$cij_marker) %>%
+    dplyr::filter(dplyr::row_number() == 1 & !is.na(.data$cij_marker) & !is.na(.data$anon_chi)) %>%
+    dplyr::ungroup() %>%
+    dplyr::mutate(
+
+      # Excluding operations are op1a codes from K01 to K50, K56, K60, and K61 (dental)
+      excluding_operation = stringr::str_sub(.data$op1a, 1, 3) %in%
+        purrr::reduce(list(
+          glue::glue("K{stringr::str_c(\"0\", c(1:9))}"),
+          glue::glue("K{as.character(c(10:50))}"),
```
Moohan, line 30: `paste0("K", 10:50)` or `glue::glue("K{10:50}")` are equivalent to `glue::glue("K{as.character(c(10:50))}")`

R/add_ppa_flag.R
```diff
@@ -0,0 +1,149 @@
...
+          c("K56", "K60", "K61")
+        ), union),
```
Moohan, line 32: This is equivalent but much simpler.
```suggestion
c(glue::glue("K{stringr::str_pad(1:50, 2, 'left', '0')}"), "K56", "K60", "K61"),
```

The `purr::reduce` -> `union` wasn't doing anything since all the items were already unique, so could have just been:
```r
c(
  glue::glue("K{stringr::str_c(\"0\", c(1:9))}"),
  glue::glue("K{as.character(c(10:50))}"),
  c("K56", "K60", "K61")
)
```

R/add_ppa_flag.R
```diff
@@ -0,0 +1,149 @@
...
+          # Congestive HF
+          "I50",
+          "J81"
+        ) &
+          excluding_operation == FALSE ~ TRUE,
+        stringr::str_sub(.data$diag1, 1, 4) %in% c(
+          # Hypertension
+          "I119", "I110"
+        ) &
+          excluding_operation == FALSE ~ TRUE,
```
Moohan, line 133:
```suggestion
          !excluding_operation ~ TRUE,
        stringr::str_sub(.data$diag1, 1, 4) %in% c(
          # Hypertension
          "I119", "I110"
        ) &
          !excluding_operation ~ TRUE,
```

R/add_ppa_flag.R
```diff
@@ -0,0 +1,149 @@
...
+        # Reliant on diag2 only
+        # Bronchitis
+        stringr::str_sub(.data$diag2, 1, 3) == "J47" ~ TRUE,
+
+        # All other values
+        TRUE ~ FALSE
+      )
+    )
+  return(matching_data)
+}
```
Moohan, line 149:
```suggestion
    ) %>% 
      dplyr::select(.data$anon_chi, .data$cij_marker, .data$ppa)
    
  return(matching_data)
}
```

R/add_ppa_flag.R
```diff
@@ -0,0 +1,149 @@
...
+
+        # Reliant on diag2 only
+        # Bronchitis
+        stringr::str_sub(.data$diag2, 1, 3) == "J47" ~ TRUE,
+
+        # All other values
+        TRUE ~ FALSE
+      )
+    )
+  return(matching_data)
```
Moohan, line 148: I think instead of returning this 'lookup style' data frame, it would be better to just do the join and return the full data.

At the moment the usage would be: `slf_data %>% left_join(add_ppa_flag(slf_data), by = c("anon_chi", "cij_marker"))` which is a bit ugly but if we did the join in the function.

e.g. 
```r
ppa_cij_data <- dplyr::left_join(data, matching_data, by = c("anon_chi", "cij_marker")) %>% 
   dplyr::rename(cij_ppa = .data$ppa)

return(ppa_cij_data)
```

R/add_ppa_flag.R
```diff
@@ -0,0 +1,149 @@
...
+add_ppa_flag <- function(data) {
+  matching_data <- data %>%
+    # Select out only the columns we need
+    dplyr::select(
+      .data$anon_chi, .data$cij_marker, .data$cij_pattype, .data$recid,
+      .data$op1a, .data$diag1, .data$diag2, .data$diag3, .data$diag4,
+      .data$diag5, .data$diag6
+    ) %>%
+    # Filter only recids and patient type where admission was preventable
+    dplyr::filter(.data$recid %in% c("01B", "02B", "04B", "GLS") & .data$cij_pattype == "Non-Elective") %>%
```
Moohan, line 19: Can you write some checks so we get a nice error message if any of the required variables are not submitted, please?

```r
if (!(all(vars %in% names(data)) {
   # Figure out which variables(s) were missing
   missing_vars <- ...

   cli::cli_abort("{missing_vars} were missing and are needed.")
}
```

Also, a check with a message/warning if none of those recids are provided (it will still work if we don't have those recids, it just won't flag any as PPA so it is a bit pointless).

Moohan (changes requested):
Some changes, mostly to simplify the code a bit.

I think it would probably be worth extracting `stringr::str_sub(.data$diag1, 1, 3)` (and maybe some others) as a new variable before the `case_when` e.g. `diag1_3char = stringr::str_sub(.data$diag1, 1, 3),`

For testing I think we would want to make a tibble of sample data:

```
Row1 = Single ep PPA
Row2 = Single ep PPA
Row3 = Single ep PPA
Row4 = Single ep, not PPA (maybe elective admission)
Row5 = Single ep, not PPA (maybe wrong diag code)
Row6 = Single ep, not PPA (maybe wrong op code, but correct diag code)
Row 7-10 = Multi ep CIJ, where the first ep is a PPA (so it should be a PPA)
Row 11-13 = Multi ep CIJ, where the first ep is not a PPA (so it should not be a PPA) - maybe ep 2 or 3 are PPAs though.
```

Then test with a snapshot (do a select so that only the vars of interest show up):
```r
expect_snapshot(add_ppa_flag(test_tibble) %>%
    dplyr::select(
      anon_chi, cij_marker, cij_pattype, recid,
      op1a, diag1, diag2, diag3, diag4, diag5, diag6,
      cij_ppa
    ))
```

R/add_ppa_flag.R
shintoLamp: Added extra line here

R/add_ppa_flag.R
shintoLamp: Always learning! I have replaced the whole `purrr::reduce` command with the suggestion given

R/add_ppa_flag.R
shintoLamp: Ah yes, the boolean thing again. Changed this

R/add_ppa_flag.R
shintoLamp: Added the join into the function and tested, works as expected

R/add_ppa_flag.R
shintoLamp: Changed this as well, I assume this is so we didn't bring diag codes into the join

R/add_ppa_flag.R
shintoLamp: Good idea, done the suggested one and will add the other when I'm back

shintoLamp: Done, but will amend with use of #387 when merged

R/add_ppa_flag.R
```diff
@@ -0,0 +1,171 @@
...
+    "anon_chi", "cij_marker", "cij_pattype", "recid",
+    "op1a", "diag1", "diag2", "diag3", "diag4",
+    "diag5", "diag6"
+  )
+  if (!all(expected_vars %in% names(data))) {
+    missing_vars <- setdiff(expected_vars, names(data))
+
+    cli::cli_abort("{missing_vars} were missing from the data and are needed
+                     to assign preventable admissions")
+  }
```
Moohan, line 21: We'll use `check_variables_exist` from #387 here when that has been merged

tests/testthat/test-add_ppa_flag.R
```diff
@@ -0,0 +1,62 @@
...
+  ))
+})
+
+test_that("Errors are handled as expected", {
+  # Wrong columns
+  error_data <- tibble::tribble(
+    ~anon_chi, ~recid, ~diag1, ~something_silly,
+    1, "01B", "A01", "Foo"
+  )
+  expect_error(add_ppa_flag(error_data))
```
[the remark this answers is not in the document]
shintoLamp: Done

R/add_ppa_flag.R
shintoLamp: Added use of this function and altered tests to reflect this

Moohan (changes requested):
Looks good.

The only issue I found was with the `anon_chi` variable name. At this point in the SLF process it would still be called `chi`.

R/add_ppa_flag.R
```diff
@@ -0,0 +1,166 @@
...
+#' @description This function takes a data frame input and determines, based on
+#' a combination of diagnostic codes and operation codes, whether an admission was
+#' preventable or not
+#' @param data A data frame
+#'
+#' @return A data frame to use as a lookup of PPAs
+#' @export
+add_ppa_flag <- function(data) {
+  check_variables_exist(data, variables = c(
+    "anon_chi", "cij_marker", "cij_pattype", "recid",
```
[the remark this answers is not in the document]
shintoLamp: Changed all references to anon_chi with chi

Jennit07 (approved):
Looks good - changes made and checks done. ready to merge

Issue #335, opened 2022-08-10
C04 - Add on NSU cohort


Jennit07:
C04 - Add on NSU cohort conversion

This script is not very long and we imagine this to be a top level script which calls upon functions. This is because we don't always have a NSU extract for latest years.

For example, something like this: 
- if year is <= latest year then run some code. else skip over this.

We discussed that we would prioritise including this in the process

shintoLamp:
#381 now merged, closing issue

Issue #331, opened 2022-08-10
Deal with PPA in C01 


Jennit07:
C01 Make episode file - Lines 212-362

This could probably be put into a function. But deal with records if they have been admitted in an emergency and match the criteria

shintoLamp:
Ongoing in #377

PR #387, opened 2022-09-27
A new function to check if variables are in the data

Moohan:
Function provided with some tests. Applicable to #335 and #331 and probably most other functions where we're taking (a part of) the SLF as input.

It should hopefully give a nice informative error message so it's super obvious which variables are missing.


shintoLamp (approved):
Looked into this and ran some test cases myself, everything works as expected. Especially good idea to check we're not trying to pass a matrix or something. I think this is fine to merge

PR #389, opened 2022-09-28
Convert MH extract script into a function

Jennit07:
There is one note when doing the check where this says ` process_extract_mental_health: no visible binding for global variable
    ‘gpprac’` which i cannot get rid of. I have tried doing `.data$gpprac` but when this is applied the function stops working and returns the error `Error in local_error_context(dots = dots, .index = i, mask = mask) : 
  promise already under evaluation: recursive default argument reference or earlier problems? `

This also needs checking for parsing issues in `read_extract_mental_health.R` when you run this, it has a parsing warning but i haven't been able to find what is causing this. I think it is to do with the coltypes but not sure which one is incorrectly parsing through. When reviewing can this be looked into please? Many thanks 

read mh extract:
- read_extract_mental_health.R
Process mh extract:
- process_extract_mental_health.R
mh tests:
- process_tests_mental_health.R
Then i think all that needs removed are production_scripts/extracts/Extract_mental_health and production_scripts/extracts/extract_tests/extract_mental_health_tests.R


Jennit07:
The binding issue of using `.data$gpprac` is now resolved by changes in #392

lizihao-anu (approved):
It looks good

PR #391, opened 2022-09-28
Convert A&E extract into a function

Jennit07:
There is one note when doing the check where this says `process_extract_ae: no visible binding for global variable ‘gpprac’` which i cannot get rid of. I have tried doing `.data$gpprac` but when this is applied the function stops working and returns the error `Error in local_error_context(dots = dots, .index = i, mask = mask) : promise already under evaluation: recursive default argument reference or earlier problems?`

Can this be checked when reviewed please? - Same issue as other scripts being converted into a function 

read ae extract:
- read_extract_ae.R

Process ae extract:
- process_extract_ae.R

ae tests:
- process_tests_ae.R

Then i think all that needs removed are production_scripts/extracts/Extract_aeand production_scripts/extracts/extract_tests/extract_ae_tests.R


Jennit07:
The binding issue of using `.data$gpprac` is now resolved by changes in #392

Jennit07:
Changes up to date with #392 Ready for review

lizihao-anu (approved):
The test run looks good

PR #392, opened 2022-09-28
Change `convert_eng_gp_prac...` to take and return a vector

Moohan:
I think this change to the function is good, as it simplifies it without changing its functionality.

Its usage would need to be changed from:
```r
data %>%
  convert_eng_gpprac_to_dummy(gpprac)
```
to
```r
data %>%
  mutate(gpprac = convert_eng_gpprac_to_dummy(gpprac))
```

This change would need making on most of your new PRs / branches...


Jennit07 (approved):
Hi @Moohan This looks great, thanks for pushing this through. One check - did you mean to merge this into mainR instead of `function_mh`? Just checking as this affects quite a few PR's then i can bring all up to date with MainR and apply the changes

Moohan:
> Hi @Moohan This looks great, thanks for pushing this through. One check - did you mean to merge this into mainR instead of `function_mh`? Just checking as this affects quite a few PR's then I can bring all up to date with MainR and apply the changes

I meant to merge it into `function_mh` so you could check it. I didn't want to merge it straight to `main-R` as if it's accepted and merged it would mean needing to change the code on all those other branches.

Happy for you to merge it all in what ever order you want once your happy that it does resolve the issue!

PR #393, opened 2022-09-28
Convert CMH into a function for processing in phase 1b

Jennit07:
There is one note when doing the check where this says `process_extract_cmh: no visible binding for global variable ‘keyTime1’
  process_extract_cmh: no visible binding for global variable ‘duration’ Undefined global functions or variables:duration keyTime1` which i cannot get rid of. I have tried including `.data$` but when this is applied the function returns parsing issues returns the wranings: `Warning messages:
1: Problem while computing keyTime1 = lubridate::hms(.data$keyTime1).
Some strings failed to parse, or all strings are NAs 
2: Problem while computing duration = lubridate::hms(.data$duration).
 Some strings failed to parse, or all strings are NAs  error Error in local_error_context(dots = dots, .index = i, mask = mask) : promise already under evaluation: recursive default argument reference or earlier problems?`

Can this be checked when reviewed please? - Same issue as other scripts being converted into a function

read cmh extract:
- read_extract_cmh.R
Process cmh extract:
- process_extract_cmh.R
cmh tests:
- process_tests_cmh.R
Then i think all that needs removed are production_scripts/extracts/Extract_cmh and production_scripts/extracts/extract_tests/extract_cmh_tests.R


R/read_extract_cmh.R
```diff
@@ -0,0 +1,55 @@
...
+      "Patient DoB Date [C]" = col_date(format = "%Y/%m/%d %T"),
+      "Gender" = col_double(),
+      "Patient Postcode [C]" = col_character(),
+      "NHS Board of Residence Code 9" = col_character(),
+      "Patient HSCP Code - current" = col_character(),
+      "Practice Code" = col_character(),
+      "Treatment NHS Board Code 9" = col_character(),
+      "Contact Date" = col_date(format = "%Y/%m/%d %T"),
+      "Contact Start Time" = col_time(format = "%T"),
+      "Duration of Contact" = col_time(format = "%M"),
```
shintoLamp, line 24:
```suggestion
      "Duration of Contact" = col_integer(),
```
Adding this as I think `col_time()` was trying to parse this as a time of day rather than a duration of minutes

R/process_extract_cmh.R
```diff
@@ -0,0 +1,77 @@
...
+      recid = "CMH",
+      smrtype = add_smr_type(recid = .data$recid),
+      year = year
+    ) %>%
+    # contact end time
+    dplyr::mutate(
+      keyTime1 = lubridate::hms(keyTime1),
+      duration = lubridate::hms(duration)
+    ) %>%
+    dplyr::mutate(keyTime2 = .data$keyTime1 + .data$duration) %>%
```
shintoLamp, line 36:
```suggestion
    dplyr::mutate(keyTime2 = hms::as.hms(.data$keyTime1 + lubridate::dminutes(.data$duration))) %>%
```
This solves the parsing issue by making sure `keyTime1` is a time of day, and the duration of `duration` is added on to it in minutes. Then it makes sure `keyTime2` is of `hms` class

shintoLamp (commented):
Just a couple of changes which I believe solves the time parsing issues. There might be a better way to do this but I tried a lot of different things and this one worked!

R/read_extract_cmh.R
Jennit07: Amazing! thanks for looking into this! Just tested this and also working fine for me! :)

R/process_extract_cmh.R
Jennit07: Great work, thanks for the fix! Tested and works perfectly. Will commit and then merge this into MainR. Thanks! :)

Jennit07:
Changes made. @shintoLamp Can you please re-review and approve/merge. Thanks

shintoLamp (approved):
Works perfectly for me, happy to merge

PR #395, opened 2022-09-28
Convert DN as a function

Jennit07:
read dn extract:
- read_extract_district_nursing.R
Process DN extract:
- process_extract_district_nursing.R

Then i think all that needs removed are production_scripts/extracts/Extract_district_nursing

DN tests are still to be done and missing from this list


lizihao-anu (approved):
Looks sensible after a test run

PR #396, opened 2022-09-28
Convert maternity as a function

Jennit07:
read maternity extract:
- read_extract_maternity.R
Process maternity extract:
- process_extract_maternity.R
maternity tests:
- process_tests_mental_health.R
Then i think all that needs removed are production_scripts/extracts/Extract_maternity and production_scripts/extracts/extract_tests/extract_maternity_tests.R


shintoLamp (approved):
I don't see any issues here, looks ready to merge

PR #398, opened 2022-09-28
Convert outpatients extract into a function

Jennit07:
read outpatients extract:
- read_extract_outpatients.R
Process outpatients extract:
- process_extract_outpatients.R
outpatients tests:
- process_tests_outpatients.R

Then i think all that needs removed are production_scripts/extracts/Extract_outpatients and production_scripts/extracts/extract_tests/extract_outpatients_tests.R


Moohan (approved):
Everything looks good to me

PR #399, opened 2022-09-30
Convert Delayed Discharges Extract into a Function for processing in phase 1b

Jennit07:
read delayed discharges extract:
- read_extract_delayed_discharges.R
Process Delayed discharges extract:
- process_extract_delayed_discharges.R
Then i think all that needs removed are production_scripts/extracts/Extract_delayed_discharges

DD tests are still to be done and missing from this list


Jennit07:
One change i have made to this function is removing smrtype from the select list before saving the output. The reason for this is that it is not applied during data processing and does not fit the function `add_smrtype`. SMRType is derived later in the process for delayed discharges in C02

Production_scripts/Process_extracts/extract_delayed_discharges.R
```diff
@@ -1,101 +0,0 @@
...
-    dd_responsible_lca,
-    monthflag,
-    cennum
-  )
-
-outfile %>%
-  # Save as zsav file
-  write_sav(get_source_extract_path(year, "DD", ext = "zsav", check_mode = "write")) %>%
-  # Save as rds file
-  write_rds(get_source_extract_path(year, "DD", check_mode = "write"))
```
shintoLamp, line 99: Covered in lines 63-87 of `process_extract_delayed_discharges.R`

R/read_extract_delayed_discharges.R
```diff
@@ -0,0 +1,18 @@
+#' Read Delayed Discharges extract
+#'
+#' @param year Year of BOXI extract
+#'
+#' @return zsav/rds data file for Delayed discharges
+#' @export
+#'
+read_extract_delayed_discharges <- function(year) {
+  extract_delayed_discharges <- haven::read_sav(get_dd_path(ext = "zsav")) %>%
```
shintoLamp, line 9: Is it sensible to hard-code a `.zsav` extension? Presumably in the future we won't be receiving the extract in this format

R/process_extract_delayed_discharges.R
```diff
@@ -0,0 +1,87 @@
...
+#' @param year The year to process, in FY format.
+#' @param data The extract to process
+#' @param write_to_disk (optional) Should the data be written to disk default is
+#' `TRUE` i.e. write the data to disk.
+#'
+#' @return the final data as a [tibble][tibble::tibble-package].
+#' @export
+#' @family process extracts
+
+process_extract_delayed_discharges <- function(year, data, write_to_disk = TRUE) {
```
shintoLamp, line 16:
```suggestion
process_extract_delayed_discharges <- function(data, year, write_to_disk = TRUE) {
```

I think it's just good practice to specify the data before the year

R/process_extract_delayed_discharges.R
```diff
@@ -0,0 +1,87 @@
+#' Process the delayed discharges extract
+#'
+#' @description This will read and process the
+#' delayed discharges extract, it will return the final data
+#' but also write this out as a zsav and rds.
+#'
+#' @param year The year to process, in FY format.
+#' @param data The extract to process
```
shintoLamp, line 8:
```suggestion
#' @param data The extract to process
#' @param year The year to process, in FY format.
```

Just to match parameter shuffling below

shintoLamp (commented):
All looking good apart from a couple of comments there

R/process_extract_delayed_discharges.R
```diff
@@ -0,0 +1,87 @@
...
+
+  # Specify MH specialties for dealing with correct DD dates
+  mh_spec <- c("CC", "G1", "G2", "G21", "G22", "G3", "G4", "G5", "G6", "G61", "G62", "G63")
+
+  dd_clean <- data %>%
+    # Use end of the month date for records with no end date (but we think have ended)
+    # Create a flag for these records
+    dplyr::mutate(
+      month_end = lubridate::ceiling_date(.data$keydate1_dateformat, "month") - 1,
+      ammended_dates = dplyr::case_when(
```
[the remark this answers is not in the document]
Jennit07: Great spot!

R/read_extract_delayed_discharges.R
Jennit07: The function `get_dd_path` is already hard-coded in `.rds` format, so eventually we will not need the `ext = .zsav` here. This is only specified as the delayed discharges team will be converting their script into R so until then, we will only have the `.zsav` versions available. Can you think of another way of doing this in the meantime? otherwise i will add an issue to review this when we have the `.rds` versions available from the team

R/process_extract_delayed_discharges.R
Jennit07: Good spot!

R/process_extract_delayed_discharges.R
```diff
@@ -0,0 +1,91 @@
...
+  dd_clean <- data %>%
+    # Use end of the month date for records with no end date (but we think have ended)
+    # Create a flag for these records
+    dplyr::mutate(
+      month_end = lubridate::ceiling_date(.data$keydate1_dateformat, "month") - 1,
+      amended_dates = dplyr::case_when(
+        .data$keydate2_dateformat == as.Date("1900-01-01") ~ TRUE,
+        TRUE ~ FALSE
+      ),
+      keydate2_dateformat = dplyr::if_else(.data$ammended_dates, .data$month_end, .data$keydate2_dateformat)
```
Moohan, line 37: Typo here

```suggestion
      amended_dates = dplyr::case_when(
        .data$keydate2_dateformat == as.Date("1900-01-01") ~ TRUE,
        TRUE ~ FALSE
      ),
      keydate2_dateformat = dplyr::if_else(.data$amended_dates, .data$month_end, .data$keydate2_dateformat)
```

Moohan, line 37: Much simpler?
```suggestion
      keydate2_dateformat = dplyr::if_else(.data$keydate2_dateformat == as.Date("1900-01-01"), .data$month_end, .data$keydate2_dateformat)
```

R/process_extract_delayed_discharges.R
```diff
@@ -0,0 +1,91 @@
...
+      (.data$keydate1_dateformat <= .data$keydate2_dateformat) | is.na(.data$keydate2_dateformat)
+    ) %>%
+    # set up variables
+    dplyr::mutate(
+      recid = "DD",
+      year = year
+    ) %>%
+    # recode blanks to NA
+    dplyr::mutate(
+      dplyr::across(tidyselect::ends_with("delay_reason"), dplyr::na_if, "")
```
Moohan, line 50: Slightly more specific
```suggestion
      dplyr::across(tidyselect::ends_with("_delay_reason"), dplyr::na_if, "")
```

R/read_extract_delayed_discharges.R
```diff
@@ -0,0 +1,18 @@
+#' Read Delayed Discharges extract
+#'
+#' @param year Year of BOXI extract
+#'
+#' @return zsav/rds data file for Delayed discharges
+#' @export
+#'
+read_extract_delayed_discharges <- function(year) {
```
Moohan, line 8: `year` is a redundant parameter here, did you mean to use `dd_period` which could be passed to `get_dd_path` ?
```suggestion
#' @return zsav/rds data file for Delayed discharges
#' @export
#'
read_extract_delayed_discharges <- function() {
```

Moohan (changes requested):
A few little things, only one actual issue.

R/read_extract_delayed_discharges.R
Jennit07: Good spot! Checked `dd_period` and we don't need this because `get_dd_path` comes from the raw data with all years. When the year is supplied in `process_extract_delayed_discharges` this will use the function `is_date_in_fyyear` to correctly filter out episodes within the year

Jennit07:
Changes made, ready for re-review @Moohan

Moohan (approved):
Looks good, no more comments.

PR #400, opened 2022-10-05
Changed typo in substance misuse cohort assignment

shintoLamp:
Changed "F11" to "F13" in each year's C07 file, to fix a typo.


Moohan (approved):
Looks good.

That's actually simpler than I was imagining, but as you say, it does just look like a typo!

PR #401, opened 2022-10-05
Improve get file path messages

Moohan:
I noticed some of the messages produced by the `get_file_path` function could be better / nicer so I updated them to use `cli`.

I also changed the `get_year_dir` function to create directories when needed. I added messaging and tests for this scenario.

### DIFF
--- a/R/get_file_paths.R
+++ b/R/get_file_paths.R
@@ -39,6 +39,11 @@ get_file_path <-
       cli::cli_abort("The directory {.path {directory}} does not exist.")
     }
 
+    check_mode <- match.arg(
+      arg = check_mode,
+      choices = c("exists", "read", "write", "execute")
+    )
+
     if (!is.null(file_name)) {
       file_path <- fs::path(directory, file_name)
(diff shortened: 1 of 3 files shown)


shintoLamp (commented):
Ran some test cases and happy with the messages displayed when I got things wrong on purpose. The only thing I'd potentially add is there's no custom error for when you add a `check_mode` that doesn't exist, like `check_mode = "silly"`.

Moohan:
> Ran some test cases and happy with the messages displayed when I got things wrong on purpose. The only thing I'd potentially add is there's no custom error for when you add a `check_mode` that doesn't exist, like `check_mode = "silly"`.

Done in [a0955315623e7c381fb4a5d58478a1fdb76a7f73](https://github.com/Public-Health-Scotland/source-linkage-files/pull/401/commits/a0955315623e7c381fb4a5d58478a1fdb76a7f73)

shintoLamp (approved):
Great, ready for merge

PR #403, opened 2022-10-11
Convert PIS into a function

Jennit07:
read PIS extract:
-read_extract_prescribing.R
Process PIS extract:
-process_extract_prescribing.R
PIS tests:
- process_tests_prescribing.R

Then i think all that needs removed are production_scripts/extracts/Extract_prescribing and production_scripts/extracts/extract_tests/extract_pis_tests.R


lizihao-anu (approved):
Testing running looks good. Happy to merge

Issue #406, opened 2022-10-11
Review lookup tests 


Jennit07:
The function `write_tests_xlsx` now includes a year parameter which does not fit how the lookup tests are set up. These need reviewed

lizihao-anu:
As for the issue "write_tests_xlsx now looks for a year parameter ", would it be ok to allow year to be a missing argument? We can use if else to separate the cases with or without year? Hope I make myself clear.

For write_tests_xlsx.R, argument year = NULL, and line 39 can be changed to

  if(missing(year)){
    sheet_name_dated <- glue::glue("{sheet_name}_{date_today}")
  }else{
    sheet_name_dated <- glue::glue("{year}_{sheet_name}_{date_today}")
  }

Moohan:
@lizihao-anu solution sounds like a good one to me, let's do it!

PR #408, opened 2022-10-11
Convert postcode lookup into a function for processing in phase 1b

Jennit07:
I've left pc lookup tests for now as the function `write_tests_xlsx` now looks for a year parameter which does not fit how this is set up - I will make a new issue to review lookup tests #406

The postcode lookup is now in one function called `process_lookup_postcode.R`


lizihao-anu:
As for the issue "`write_tests_xlsx` now looks for a year parameter ", would it be ok to allow `year` to be a missing argument? We can use if else to separate the cases with or without `year`? Hope I make myself clear.

If so, can I push a commit directly to function_lookup_pc branch?

lizihao-anu (approved):
Happy to merge if there is no intention to fix `write_tests_xlsx.R` here

Moohan:
> Happy to merge if there is no intention to fix `write_tests_xlsx.R` here

Dealt with in #376

Jennit07:
Simplified the code in the tests. Ready to merge

PR #409, opened 2022-10-11
Convert gpprac lookup into a function  for processing in phase 1b

Jennit07:
I've left gpprac lookup tests for now as the function write_tests_xlsx now looks for a year parameter which does not fit how this is set up - I will make a new issue to review lookup tests https://github.com/Public-Health-Scotland/source-linkage-files/issues/406

The gpprac lookup is now in one function called process_lookup_gpprac.R


R/process_lookup_gpprac.R
```diff
@@ -0,0 +1,123 @@
...
+    # Sort for SPSS matching
+    dplyr::arrange(.data$gpprac) %>%
+    # Write rds file
+    write_rds(get_practice_details_path(check_mode = "write"))
+
+
+  # Read Lookup files ---------------------------------------
+  # gp lookup
+  gpprac_ref_file <-
+    readr::read_rds(get_gpprac_ref_path()) %>%
```
lizihao-anu, line 58: The test run reports an error there. It is because the function `get_gpprac_ref_path()` use the extension '.sav'. I guess it will be changed to '.rds' some while later but for now, we use '.sav'. Am I right?

Jennit07: Good spot, i'll push the change and we can read the zsav version for now but this will change to .rds when this is available

Moohan: Can we check if there's tests for `get_gpprac_ref_path` so we can be alerted if/when it's changed from a sav to rds version.

lizihao-anu (approved):
The test run looks good.

R/process_lookup_gpprac.R
Moohan: This is now done in #446

PR #411, opened 2022-10-12
Convert LTC lookup into a function for processing in phase 1b

Jennit07:
The LTC lookup is now in two functions called `read_lookup_ltc.R` and `process_lookup_ltc.R`. The LTC lookup extract is now removed from production scripts


shintoLamp (approved):
Ran through, everything seems to be working fine, all the expected LTC variables showing up

PR #413, opened 2022-10-12
Convert SC client lookup into a function for processing in phase 1b

Jennit07:
The sc client lookup is now in two functions called `read_lookup_sc_client.R` and `process_lookupsc_client.R`. The sc client lookup extract is now removed from production scripts.


shintoLamp (approved):
All looking good to me, ran it with no issues

Issue #319, opened 2022-07-28
Expand (and rename) `get_datazone_pop_path` so it can read any 'type' of pop estimate 


Moohan:
Currently the function will search for the latest datazone (all ages) pop estimates. However you can also specify a literal file name and it will return that (e.g. in the DN costs).

It should be renamed to something like `get_pop_estimate_path` with new parameter `type` which could select different pop files e.g. datazone, hscp etc.

Usage should be updated for instance, in the DN costs, the usage should be `get_pop_estimate_path(type = "hscp")` so that it always returns the latest hscp population estimates.

PR #417, opened 2022-10-19
Convert District Nursing costs lookup into a function for processing in phase 1b

Jennit07:
Assigning @Moohan to check over this one as this set up differs from the other scripts being converted into a function. As discussed, we thought it would be good to have an RMarkdown document for processing the cost lookups as they include graphs which we would like to see during processing. 

I'm not sure i have captured this right but the function will process the Rmarkdown document and then read the saved data for the `data_list` used in the test-production branch. On teams you mentioned `A hack might be to save the data you need to the /tmp directory (this get's wiped on every new R session) in the Rmarkdown, then read it and return in the function.` but im not sure i have captured this properly. Any ideas/improvements on this one will be useful for other cost lookups which will follow the same kind of format


Production_scripts/Create_costs/costs_district_nursing.R
```diff
@@ -88,7 +88,7 @@ dn_raw_costs_contacts <- left_join(dn_raw_contacts,
 # Of the two HSCPs, Argyll and Bute provides the
 # District Nursing data which is 27% of the population.
 
-population_lookup <- readr::read_rds(get_datazone_pop_path("HSCP2019_pop_est_1981_2020.rds")) %>%
+population_lookup <- readr::read_rds(get_datazone_pop_path("HSCP2019_pop_est_1981_2021.rds")) %>%
```
Moohan, line 91: This still needs addressing in #319

Moohan (commented):
This looks fine, ignore what I was saying about temp files etc. it's irrelevant since we are writing out the costs data anyway!

I was hoping to do a proper review but have run out of time, so just left these comments and I'll check it properly next week!

Moohan (approved):
@Jennit07 I got a bit carried away with this one...

I had to do a reset back to `main-R` to resolve all the weird conflicts. I then tested everything and made some changes to how the costs stuff work.

- Deleted all the 'old' cost scripts
- There's now a 'top-level' function which does the `rmarkdown::render` with some consistent settings
   - The output dir is now the `tests` folder on `hscdiip`
   - The render should now only show error messages, not all the other messages
   - It should display the output in the active browser. Possibly we want this to be optional with a parameter?
- Fixed the copy bit in the cost scripts (needed a `check_mode = "write"` and an `overwrite = `TRUE`).
- Added a new action to the `style.yaml` workflow so that any files in `Rmarkdown/` will also get styled.

I've approved as I've checked everything and it seems to be working but can you give it a look over and test before you merge it please?

Jennit07:
> @Jennit07 I got a bit carried away with this one...
> 
> I had to do a reset back to `main-R` to resolve all the weird conflicts. I then tested everything and made some changes to how the costs stuff work.
> 
> * Deleted all the 'old' cost scripts
> * There's now a 'top-level' function which does the `rmarkdown::render` with some consistent settings
>   
>   * The output dir is now the `tests` folder on `hscdiip`
>   * The render should now only show error messages, not all the other messages
>   * It should display the output in the active browser. Possibly we want this to be optional with a parameter?
> * Fixed the copy bit in the cost scripts (needed a `check_mode = "write"` and an `overwrite = `TRUE`).
> * Added a new action to the `style.yaml` workflow so that any files in `Rmarkdown/` will also get styled.
> 
> I've approved as I've checked everything and it seems to be working but can you give it a look over and test before you merge it please?

Thanks @Moohan, great work tidying this up and all the comments- I prefer this format/having all the costs in the one script as it makes it more readable. Checked the three cost scripts were working as intended which im happy with and going to merge.

Issue #360, opened 2022-08-17
Create a function for computing changed sc_id.


Moohan:
We should create a function for computing changed sc_id.

_Originally posted by @Moohan in https://github.com/Public-Health-Scotland/source-linkage-files/pull/324#discussion_r947979959_

Take some social care data, with `sending_lca`, `social_care_id`, `period` and `chi` variables.

For each `chi` and `sending_lca` if there are multiple `sc_id`s, then replace them with the latests submitted one.

Jennit07:
Completed in #440

PR #418, opened 2022-10-19
Convert All Home care script into a function for processing in phase 1b

Jennit07:
Follows the same format as other functions. First read in data from the SC platform using `read_sc_all_home_care` function and process it using `process_sc_all_home_care`. I have also removed the home care script from the social care folder.


DESCRIPTION
```diff
@@ -24,6 +24,7 @@ Imports:
     fs (>= 1.5.2),
     glue (>= 1.6.2),
     haven (>= 2.5.0),
+    hms,
```
Moohan, line 27: Having looked at [the changelog](https://hms.tidyverse.org/news/index.html), I think we should require > 1.1.0 as that version had some breaking changes which might affect us.
```suggestion
    hms (>= 1.1.0),
```

R/process_sc_all_home_care.R
```diff
@@ -0,0 +1,183 @@
...
+#' all home care extract, it will return the final data
+#' but also write this out as a rds.
+#'
+#' @param data The extract to process
+#' @param write_to_disk (optional) Should the data be written to disk default is
+#' `TRUE` i.e. write the data to disk.
+#'
+#' @return the final data as a [tibble][tibble::tibble-package].
+#' @family process extracts
+#' @import lubridate
```
Moohan, line 13: We definitely don't want to import the whole lubridate package. I couldn't see where you used it but we should be able to use `lubridate::<fun>` or possibly have `@importFrom lubridate <fun>` in `createslf-package.R`.

R/process_sc_all_home_care.R
```diff
@@ -0,0 +1,183 @@
...
+
+  # Data Cleaning ---------------------------------------
+
+  # period start and end dates
+  period_dates <- matched_hc_data %>%
+    dplyr::distinct(.data$period) %>%
+    dplyr::mutate(
+      record_date = lubridate::yq(.data$period) %m+% lubridate::period(6, "months") %m-% lubridate::days(1),
+      qtr_start = lubridate::yq(.data$period) %m+% lubridate::period(3, "months")
+    )
```
Moohan, line 32: I think it's probably this that meant you needed to import `lubridate`?

We could use [`lubridate::add_with_rollback()`](https://lubridate.tidyverse.org/reference/mplus.html) here. e.g.
```suggestion
  period_dates <- matched_hc_data %>%
    dplyr::distinct(.data$period) %>%
    dplyr::mutate(
      record_date = lubridate::add_with_rollback(lubridate::yq(.data$period), lubridate::period(6, "months")) - lubridate::days(1),
      qtr_start = lubridate::add_with_rollback(lubridate::yq(.data$period), lubridate::period(3, "months"))
    )
```
Note we don't actually need to use `%m+%`, `%m-%` or `add_with_rollback()` when adding or subtracting days as the result isn't ambiguous. This function is to deal with the issue that a 'month' isn't a fixed number of days/hours/seconds, so there is ambiguity if you 'add (or subtract) a month', especially if you add a month to a date like 31-MM-YYYY (not relevant for this code)

The other option here is to use the function(s) in `get_fy_quarter_dates.R` which deal with almost this situation and are set up to be fast to use 'inline' i.e. we don't have to do this 'precalculation' step

R/read_sc_all_home_care.R
```diff
@@ -0,0 +1,39 @@
...
+#'
+read_sc_all_home_care <- function() {
+
+  # Read in data---------------------------------------
+
+  # set-up conection to platform
+  db_connection <- phs_db_connection(dsn = "DVPROD")
+
+  # read in data - social care 2 home care
+  home_care_data <- dplyr::tbl(db_connection, dbplyr::in_schema("social_care_2", "homecare_snapshot")) %>%
```
Moohan, line 13: What if we passed the connection as a parameter? That way we'd only have to open the connection once and could use it for all of the SC functions.

Have set a default, so that if you don't pass a connection it will still work by asking you to create one.
```suggestion
read_sc_all_home_care <- function(sc_dvprod_connection = phs_db_connection(dsn = "DVPROD")) {

  # Read in data---------------------------------------

  # read in data - social care 2 home care
  home_care_data <- dplyr::tbl(sc_dvprod_connection, dbplyr::in_schema("social_care_2", "homecare_snapshot")) %>%
```

R/process_sc_all_home_care.R
Jennit07: Yeah this is the reason why i added lubridate to import to remove the error message i was getting. But i forgot the fix for this was `add_with_rollback()`. I'll have a look at the other functions in `get_fy_quarter_dates` as that would make more sense to use them here

R/read_sc_all_home_care.R
Jennit07: I think this would be a good fix as this can also be implemented for other sc functions using the sc platform

Jennit07:

R/process_sc_all_home_care.R
```diff
@@ -0,0 +1,186 @@
...
+    dplyr::group_by(.data$sending_location, .data$social_care_id) %>%
+    dplyr::mutate(latest_sc_id = dplyr::last(.data$social_care_id)) %>%
+    # count changed social_care_id
+    dplyr::mutate(
+      changed_sc_id = dplyr::if_else(!is.na(.data$chi) & .data$social_care_id != .data$latest_sc_id, 1, 0),
+      social_care_id = dplyr::if_else(!is.na(.data$chi) & .data$social_care_id != .data$latest_sc_id,
+        .data$latest_sc_id, .data$social_care_id
+      )
+    ) %>%
+    dplyr::ungroup() %>%
```
Moohan, line 59: Can this block of code be extracted as its own function to be used elsewhere, please?

```R
replace_sc_id_with_latest <- function(data, ..?) {
# Check for required variables

# Filter to CHIs with multiple sc_id per sending loc

# Sort (by period?) to find the latest sc_id for each CHI / sending_loc

# match that back to the data overwriting the older one.

# Return the data (no new variables added, just overwrite the sc_id)

}
```

R/process_sc_all_home_care.R
```diff
@@ -0,0 +1,186 @@
...
+      start_after_end = .data$hc_service_start_date > .data$hc_service_end_date & !is.na(.data$hc_service_end_date)
+    ) %>%
+    dplyr::filter(
+      !.data$end_before_qtr,
+      !.data$start_after_quarter,
+      !.data$start_after_end
+    )
+
+  # count changed social_care_id
+  home_care_clean %>% dplyr::count(.data$changed_sc_id)
```
Moohan, line 78: This isn't doing anything in the context of the function.
```suggestion
```

R/process_sc_all_home_care.R
Moohan: Already an issue! #360

Jennit07: PR open #440

Jennit07:
ready for re-review

Moohan (approved):

PR #419, opened 2022-10-20
Changes due to `tidyselect` 1.2.0

Moohan:
In the latest version of `tidyselect` 1.2.0 they deprecated the use of `.data$` in 'selection contexts' this means things like `dplyr::select` and `across`, and probably some others too!

This was causing a lot of warnings to appear for our tests.

I've gone through and replaced all the instances I could find. The tests are now (still) passing with no warnings.

I'm aware that not all our code is covered by the tests though so please check to see if I've missed or broken anything.

See [the blog post](https://www.tidyverse.org/blog/2022/10/tidyselect-1-2-0/#using-data-inside-selections) for more info and background.


Moohan:
I've pushed a few other minor changes and fixes directly to `main-R` so this PR should now fix all of the remaining build issues, except for the `hms` import one which is being fixed elsewhere.

Jennit07 (commented):
I've checked through each function and think the changes are being picked up correctly. Could not find any more examples to change to this. There are some build warnings on the checks which includes the `hms` warning and something failing with one of the tests. See error below 

═════Failed tests══════
── Error (test-get_slf_lookup_paths.R:19:3): SLF Care Home names lookup path works ──
Error in `get_file_path(directory = fs::path(get_slf_dir(), "Lookups"), 
    file_name = glue::glue("Care_Home_Lookup_All.xlsx"), check_mode = "write", 
    ...)`: formal argument "check_mode" matched by multiple actual arguments

Jennit07:
I've fixed the checks on this branch, and everything now works/passes as expected. I'm happy with this but I think this could have one more glance over to ensure everything has been picked up on since there are quite a lot of scripts being changed/updated

Jennit07 (approved):
Approved, pending one more glance over since there are a lot of scripts being changed

shintoLamp (approved):
All looks good to me

PR #425, opened 2022-10-25
Fix sds_end_date is earlier than sds_start_date

### LINKED ISSUE #415: SDS bug - deal with end date before start
These are possibly being created when we set missing start dates to the start of the year (using the period) or could be coming from the data.

We should check with the Social Care team to see what to do about these - drop them, drop the start / end date, or set the start / end date to something sensible.

The change will need making on `master` and `main-R`

lizihao-anu:
To main-R, close #415 

Fix sds_end_date is earlier than sds_start_date by setting end_date to be the end of fyear


lizihao-anu:
Once it is merged, we can merge the pull request #405

Jennit07 (approved):
Looks good! Happy to merge

PR #427, opened 2022-10-25
Convert care home costs lookup into a function for processing in phase 1b 

Jennit07:
The cost lookups follow a different format to other scripts being converted into a function. Instead, this renders an Rmarkdown file containing the code and produces some graphs for any cost differences. The function will render this and then read the saved output and return the data. This will be the format for OOH cost lookup, DN cost lookup, Care home cost lookup and Home care cost lookup


shintoLamp (approved):
As with home care, all looks fine to me

PR #429, opened 2022-10-26
BUG - Fix sc client function

Jennit07:
When trying to run some social care year-specific extracts, I found the client file was returning 0 rows. This was due to the filtering which occurred in `read_lookup_sc_client` which made more sense to be in `process_lookup_sc_client`. I have also used the function `convert_fyyear_to_year` here as the client data is in cal year format instead of fyyear so this was also breaking the code. I think this now fixes both issues and I'm happy this is working


Moohan:
It seems to me better to do the filter in the dbplyr code, so that we're returning the minimal data needed.

I think this would require having a `year` parameter for `read_lookup_sc_client`. Please ignore if I've missed/misunderstood something.

All_years/04-Social_Care/04-SDS_data.R
```diff
@@ -68,8 +68,8 @@ sds_full_clean <- sds_full_data %>%
   )) %>%
   # Fix sds_end_date is earlier than sds_start_date by setting end_date to be the end of fyear
   mutate(sds_end_date = if_else(
-    sds_start_date>=sds_end_date,
-    end_fy(year = period, 'alternate'),
+    sds_start_date >= sds_end_date,
```
lizihao-anu, line 71: Just wondering of using `>=` here or `>` ? As in #430, we chose to use `>`, ie `service_start_date > service_end_date`,

Moohan: Yep, should be `>`, although I'm not really sure why this code change is on this branch at all 😂

lizihao-anu: Might be my fault

lizihao-anu, line 71: `sds_start_date > sds_end_date,`

lizihao-anu, line 71:
```suggestion
    sds_start_date > sds_end_date,
```

R/read_lookup_sc_client.R
```diff
@@ -3,17 +3,15 @@
...
 #' @param year The year to process
 #'
 #' @return the final data as a [tibble][tibble::tibble-package].
 #' @export
 #' @family process extracts
-read_lookup_sc_client <- function(year) {
-  # set-up conection to platform
-  db_connection <- phs_db_connection(dsn = "DVPROD")
-
+read_lookup_sc_client <- function(sc_dvprod_connection = phs_db_connection(dsn = "DVPROD"), year = convert_fyyear_to_year(year)) {
```
Moohan, line 12: This doesn't make sense to have the default value for a parameter also contain that parameter.

We can, however, just take a year in fyyear format (and check this is the case), then convert it to the calendar year format we need before using it in the `dbplyr` code

```suggestion
read_lookup_sc_client <- function(sc_dvprod_connection = phs_db_connection(dsn = "DVPROD"), fyyear) {

check_year_format(fyyear)
year <- convert_fyyear_to_year(fyyear)
```

R/read_lookup_sc_client.R
```diff
@@ -3,17 +3,15 @@
 #' @description This will read and process the
 #' social care client lookup
 #'
+#' @param sc_dvprod_connection The connection to the SC platform.
 #' @param year The year to process
```
Moohan, line 7:
```suggestion
#' @param fyyear The year to process, in the standard format '1718'
```

Jennit07:

Moohan (approved):

PR #430, opened 2022-10-28
BUG - fix alarms telecare end dates

Jennit07:
This fixes cases where the start date is earlier than the end date. Same logic from #425 applied.

PR #431, opened 2022-10-28
Convert (year specific) Alarms Telecare script into a function for processing in phase 1b

Jennit07:
I've tested this using the sc all alarms telecare extract and happy this is now working as expected. I ran into a few bugs which have now been corrected and open in PR's #429 #430. I also noticed that we were not creating a year variable. I thought this was to be done in the all alarms telecare script but it should actually be in the year specific which has been completed. You will see some commits in the history which reflect this as i made the changes and reverted them. I think these changes will also need to be made when converting SDS into a function which i will also do. Overall, happy this is working and produces an output that we would expect.


lizihao-anu (approved):
The test run looks good

PR #432, opened 2022-10-31
Convert (year specific) SDS script into a function for processing in phase 1b

Jennit07:
Conversion of the year specific SDS script. Tested this using the SDS all data script and happy this is working as expected. As mentioned in #431 there were a few bugs which i have corrected here


lizihao-anu (approved):
The test running looks good

PR #435, opened 2022-10-31
Add Bateman and Zihao as package authors

Moohan:
Please check and approve. Also, add an [ORCID](https://orcid.org/) if you want/have one.


lizihao-anu (approved):
just added my ORCID id. Happy to merge if Bateman has nothing to add

shintoLamp (approved):
Cool, |I don't think I have an ORCID so happy with it as it is

PR #436, opened 2022-11-01
Fixes for GPOOH smrtype

Moohan:
Missed at the last update, noticed as I was working on the GP OoH R code!


Jennit07 (approved):
Looks good!

PR #437, opened 2022-11-02
384 Add tests for District Nursing data

### LINKED ISSUE #384: Convert DN tests

lizihao-anu:
Close #384


Production_scripts/Process_extracts/Produce_extract_tests/extract_district_nursing_tests.R
```diff
@@ -0,0 +1,43 @@
...
+
+
+# Read in Data-----------------------------------------
+
+year <- check_year_format("1920")
+
+# Read new data file
+# new_data <- readr::read_rds(get_source_extract_path(year, "DN"))
+new_data <-
+  haven::read_spss("/conf/sourcedev/Source_Linkage_File_Updates/1920/DN_for_source-201920.zsav") %>%
```
Moohan, line 22: This should be reading from the `rds` file, using the inbuilt function.

If that file doesn't exist we will need to run the processing code to produce it so we can test this code properly.

Production_scripts/Process_extracts/Produce_extract_tests/extract_district_nursing_tests.R
```diff
@@ -0,0 +1,43 @@
...
+  dplyr::rename(year = Year)
```
Moohan, line 23: I think this rename is only required because you read in the SPSS file? It should be removed, everything in the `rds` files should be `snake_case` and if it isn't we need to fix it there.

@Jennit07 can we look at 'fixing' this in the SPSS too please? SPSS doesn't care about case so we can just swap `Year` for `year` without it affecting anything in the SPSS code.

Moohan (commented):
Just some minor comments. I still need to run/check the code in R but it looks good to me so far.

Jennit07 (approved):
Tested the code and it looks good to me. Happy to approve

PR #438, opened 2022-11-04
Fix sds option 4 bug - mainR

Jennit07:
Addresses the issue #416 where angus had cases of records with all sds options 1-3 missing. I have written the code so that this will catch any instances of this happening. The table below shows which areas (and how many) will change in the raw data. This branch will merge into mainR which is a copy cherry picked from the branch `bug_sds_opt4`

```
  sending_location     n
  <chr>            <int>
1 120                490 (Angus)
2 230                  3 (Edinburgh
3 235                  5 (Na_h_Eileanan_Siar)
4 260                  1 (Glasgow)
5 300                 26 (Moray)
6 355                  2 (Scottish Borders)
```

Issue #416, opened 2022-10-18
SDS bug - Angus option 4


Moohan:
[Peter SDS email.pdf](https://github.com/Public-Health-Scotland/source-linkage-files/files/9808432/Peter.SDS.email.pdf)

Per comment/email from @petermcclurg Angus should have any records with all SDS 1, 2 and 3 missing coded to SDS 4.

I'm not quite sure how this will affect our code... 

The change will need making in `master` and `main-R`

PR #439, opened 2022-11-04
Fix sds option4 bug - SPSS (December 2022 Update)

Jennit07:
Addresses issue #416 where Angus had cases of records with all SDS options 1-3 missing. I have written the code so that this will catch any instances of this happening. The table below shows which areas (and how many) will change in the raw data. This branch will merge into the `december_22_update` branch which is identical to the branch `bug_sds_opt4-mainR` open PR #438 

```
  sending_location     n
  <chr>            <int>
1 120                490 (Angus)
2 230                  3 (Edinburgh
3 235                  5 (Na_h_Eileanan_Siar)
4 260                  1 (Glasgow)
5 300                 26 (Moray)
6 355                  2 (Scottish Borders)

```


All_years/04-Social_Care/04-SDS_data.R
```diff
@@ -58,14 +58,24 @@ sds_full_clean <- sds_full_data %>%
   mutate(across(starts_with("sds_option_"), ~ .x == "1")) %>%
   # SDS option 4 is derived when a person receives more than one option.
   # e.g. if a person has options 1 and 2 then option 4 will be derived
-  mutate(sds_option_4 = rowSums(across(starts_with("sds_option_"))) > 1, .after = sds_option_3) %>%
+  mutate(
+    sds_option_4 = rowSums(across(starts_with("sds_option_"))) > 1, .after = sds_option_3,
+    # Fix sds option 4 cases where all 3 options are missing
+    sds_option_4 = if_else(sds_option_1 == FALSE & sds_option_2 == FALSE & sds_option_3 == FALSE, TRUE, sds_option_4)
```
Moohan, line 64: Simplification.
```suggestion
    sds_option_4 = if_else(!any(sds_option_1, sds_option_2, sds_option_3), TRUE, sds_option_4)
```

Moohan: I think we need to apply this to only Angus, i.e.
```suggestion
    sds_option_4 = if_else(!any(sds_option_1, sds_option_2, sds_option_3) & sending_location == "Angus", TRUE, sds_option_4)
```

Can you also double check that there are no `NA`s? I think this will be dealt with on line 58 but I'm not sure.

Moohan (commented):
Also applies to #438

All_years/04-Social_Care/04-SDS_data.R
Jennit07: In my latest commit, i have tried to deal with the NA's but i don't think this is working right as i tried to do this in a clever way. Any suggestions to improve this? code below

 `mutate(across(starts_with("sds_option_"), ~ if_else(is.na(.x), FALSE,.x)))`

Moohan (approved):
I'm happy with this now.

All_years/04-Social_Care/04-SDS_data.R
```diff
@@ -55,17 +55,31 @@ sds_full_data <- tbl(db_connection, dbplyr::in_schema("social_care_2", "sds_snap
...
   mutate(sds_start_date = if_else(is.na(sds_start_date),
     start_fy(year = period, format = "alternate"),
     sds_start_date
   )) %>%
+  # Fix sds_end_date is earlier than sds_start_date by setting end_date to be the end of fyear
+  mutate(sds_end_date = if_else(
+    sds_start_date >= sds_end_date,
+    end_fy(year = period, "alternate"),
+    sds_end_date
+  )) %>%
```
Moohan, line 82: This block of code keeps cropping up in PRs!
```suggestion
  # Fix sds_end_date is earlier than sds_start_date by setting end_date to be the end of fyear
  mutate(sds_end_date = if_else(
    sds_start_date > sds_end_date,
    end_fy(year = period, "alternate"),
    sds_end_date
  )) %>%
```

Jennit07: New issue open #447

PR #440, opened 2022-11-07
Replace sc id with the latest

Jennit07:
New function for #360 which can be applied to any instances where we want to replace sc id with the latest sc id because there are multiple sc id for sending location/chi


Jennit07:
I've added some tests to this PR. Ready for re-review

Moohan (approved):

PR #441, opened 2022-11-08
Translation of C05

shintoLamp:
As you can see from the title, this is a work in progress. The main stumbling block I came across is in the [SPSS code](https://github.com/Public-Health-Scotland/source-linkage-files/blob/master/2019-20/C05%20Match%20on%20LTCs.sps), from lines 101-111, which involves dealing with `lag`. 

I wrote some code as a literal translation from SPSS from lines 111-122, but I know that won't work as expected. Another idea I had was to use the `fill` function, but that's as far as I got.


lizihao-anu:
**A really long summary**
Thanks to Bateman's work, it actually can be finished in less than half a day. I have been thinking about using a more readable and intuitive way to rebuild the code and have done much research. I didn't find the best way that performs others in every aspect though. Bateman's way is good generally speaking in this case and I follow it.

However, I feel it would still be beneficial to discuss the topic (it's very likely that you guys already knew it. Then at least good for me >_<). I would discuss the essence of the difficulty and possible ways with its advantages and disadvantages.

So the real difficulty and inelegance are based on many conditions (ie many if-elses), to assign values to multiple variables (ie columns) in a data frame. In our case, we need to assign dob, age, changed_dob based on conditions.

Solutions:

1. `dplyr::case_when`, which Bateman and I choose this way
The problem is that `dplyr::case_when` can only assign values based on conditions to one variable/column. In our case, it is acceptable as we only need to replicate the `case_when` three times and `case_when` is efficient and fast. But for more columns, it is an unacceptable way as if we change one condition, we have change too many times. So in short,

- only good for a small number of columns to be assigned
- efficient and fast
- but not acceptable for more columns to be assigned.

2. `data.table::fcase`
This is actually the best the solution so far for most of case other than ours though. It is fast and can assign multiple different values to multiple differnt columns. Just the way of writing the script is a little bit counter intuitive. Normally, we write the conditions like in `case_when`

```
    condition1 ~ value1, 
    condition2 ~ value2, 
    ...

```
_**But `fcase` need to be written in reverse.**_ 
```
df1 <- df1[condition3,
               `:=`(var1 = value13,
                     var2 = value23,
                     ...)]
df1 <- df1[condition2,
               `:=`(var1 = value12,
                     var2 = value22,
                     ...)]
df1 <- df1[condition1,
               `:=`(var1 = value11,
                     var2 = value21,
                     ...)]
```
The reason I didn't choose this way is that I am worried about the potential impact of claiming `library(data.table)` on the whole project. But I am happy to use this method in this script if @Moohan are happy with claiming `library(data.table)` and I actually have finished the script in this way.
So in short,

- efficient and fast in code running
- efficient in developing the code
- just a little bit not intuitive
- unforeseen impact in our project mainly using `tidyverse`, `dplyr`


3. a function to enable `case_when` to assign multiple values to multiple columns
The function and the usage are
```
lod_casewhen <- function(...) {
  args <- list(...)
  do.call(function(...)
    mapply(data.frame, ..., SIMPLIFY = FALSE), args)
}

df1 <- df1 %>%
  mutate(
    case_when(
      condition1 ~ lod_casewhen(var1 = value11,
                                var2 = value21,
                                var3 = value31),
      condition2 ~ lod_casewhen(var1 = value12,
                                var2 = value22,
                                var3 = value32),
      condition3 ~ lod_casewhen(var1 = value13,
                                var2 = value23,
                                var3 = value33)
    ) %>%
      dplyr::bind_rows()
  )
```
In short, 

- This is really good for developing scripts for small data frames, 
- but ridiculously slow when running for large data frames.

4. **_(Not available currently!!!)_**`  dplyr::case_when` in the future release
It seems that the `dplyr` developers have been trying to add the multiple assignment feature to Package `dplyr`. See [Issue #6145](https://github.com/tidyverse/dplyr/pull/6145) in GitHub `dplyr` repository. But what confuses me is that as I understand, this feature should be included in the version 1.0.10 which has been released. But as I checked, the feature is not supported actually and I cannot see why. Would it be possible for @Moohan to have a look at [Issue #6145](https://github.com/tidyverse/dplyr/pull/6145) when you have free time? Completely fine if not. 

Again, you guys probably already knew it. Just want to write a short summary for myself. Thanks for reading!

R/match_on_ltcs.R
```diff
@@ -0,0 +1,171 @@
...
+    haven::read_sav(
+      "/conf/sourcedev/Source_Linkage_File_Updates/1920/temp-source-episode-file-4-1920-testing.zsav",
+      # "~/temp-source-episode-file-4-1920.zsav",
+      # n_max = n_maxx
+      n_max = 1e6
+    )
+
+  # Match on LTC lookup
+  matched <- dplyr::left_join(temp,
+    readr::read_rds(get_ltcs_path("1920")),
```
lizihao-anu, line 23: Maybe a year parameter to replace "1920"?

R/match_on_ltcs.R
```diff
@@ -0,0 +1,171 @@
...
+#' @examples
+match_on_ltcs <- function() {
+  # Get temp file
+  temp <-
+    haven::read_sav(
+      "/conf/sourcedev/Source_Linkage_File_Updates/1920/temp-source-episode-file-4-1920-testing.zsav",
+      # "~/temp-source-episode-file-4-1920.zsav",
+      # n_max = n_maxx
+      n_max = 1e6
+    )
```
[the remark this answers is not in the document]
Moohan: For reading the data, at the moment we have two options.

1. Read the SPSS file like you have done (@Jennit07 probably we should re-output these as `fst` files, which would make reading them much quicker?)
2. Read the finished SLF using [slfhelper](https://github.com/Public-Health-Scotland/slfhelper), and only selecting the required variables. This is much faster to read lots of data but the file might not be in the correct 'state' for testing. For example this code does lots of fixes on DoB but in the final SLF these have already been fixed.

R/match_on_ltcs.R
```diff
@@ -0,0 +1,171 @@
...
+      # Mutate age based on different scenarios
+      age = dplyr::case_when(
+        # Case when the dob is already valid
+        .data$dob == chi_dob_min ~ chi_age_max,
+        .data$dob == chi_dob_max ~ chi_age_min,
+        # If one option for dob isn't there, use the other
+        is.na(chi_dob_min) & !is.na(chi_dob_max) ~ chi_age_min,
+        !is.na(chi_dob_min) & is.na(chi_dob_max) ~ chi_age_max,
+        # If the age would be negative, use the older
+        chi_age_min < 0 ~ chi_age_max,
```
Moohan, line 59: I don't think this would ever be the case now we're using `phsmethods::dob_from_chi`. Could you double check please but I think this condition can be removed.
```suggestion
```

R/match_on_ltcs.R
```diff
@@ -0,0 +1,171 @@
...
+  # Checking and changing DOB and age
+  test <- matched %>%
+    dplyr::filter(!is_missing(chi)) %>%
+    dplyr::mutate(
+      # Create a dob in the previous century from the chi number
+      chi_dob_min = phsmethods::dob_from_chi(
+        chi_number = chi,
+        chi_check = FALSE,
+        min_date = lubridate::ymd("1900-01-01"),
+        max_date = lubridate::ymd("2000-01-01")
```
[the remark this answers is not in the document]
Moohan: We can shrink this range using the activity they have to have been born before the activity.

```suggestion
        max_date = pmin(keydate1_dateformat, lubridate::ymd("1999-12-31"))
```

R/match_on_ltcs.R
```diff
@@ -0,0 +1,171 @@
...
+      ),
+      # Create a dob in the current century from chi (will return NA if in the future)
+      chi_dob_max = phsmethods::dob_from_chi(
+        chi_number = chi,
+        chi_check = FALSE,
+        min_date = lubridate::ymd("2000-01-01")
```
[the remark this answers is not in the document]
Moohan:
```suggestion
        min_date = lubridate::ymd("2000-01-01"),
        max_date = max_date = pmin(keydate1_dateformat, end_fy(year))
```

R/match_on_ltcs.R
```diff
@@ -0,0 +1,171 @@
...
+        ((is.na(age) & chi == dplyr::lag(chi)) & (chi_age_max == dplyr::lag(age) | chi_dob_min == dplyr::lag(dob))) ~ chi_age_max,
+        ((is.na(age) & chi == dplyr::lag(chi)) & (chi_age_min == dplyr::lag(age) | chi_dob_max == dplyr::lag(dob))) ~ chi_age_min,
+        TRUE ~ age
+      ),
+      changed_dob = dplyr::case_when(
+        ((is.na(age) & chi == dplyr::lag(chi)) & (chi_age_max == dplyr::lag(age) | chi_dob_min == dplyr::lag(dob))) ~ 10,
+        ((is.na(age) & chi == dplyr::lag(chi)) & (chi_age_min == dplyr::lag(age) | chi_dob_max == dplyr::lag(dob))) ~ 10,
+        TRUE ~ changed_dob
+      )
+    ) %>%
```
Moohan, line 140: I'm sure we can do this better in R. e.g. [tidyr::fill](https://tidyr.tidyverse.org/reference/fill.html)
```suggestion
    # If we still don't have an age, try and fill it in from other records.
  dplyr::group_by(chi) %>%
  tidyr::fill(dob, .direction = "downup") %>%
  dplyr::ungroup()
```

R/match_on_ltcs.R
```diff
@@ -0,0 +1,171 @@
...
+    dplyr::mutate(age = dplyr::if_else(is.na(age), compute_mid_year_age(year, dob), age)) %>%
+    # If any gender codes are missing or 0 recode to CHI gender.
+    dplyr::mutate(chi_gender = phsmethods::sex_from_chi(chi)) %>%
+    # I cannot see why this is complicated in spss but I follow
+    dplyr::mutate(
+      gender = dplyr::case_when(
+        (is.na(gender) | gender == 0) & (chi_gender %% 2 == 1) ~ 1,
+        (is.na(gender) | gender == 0) & (chi_gender %% 2 == 0) ~ 2
+      )
+    ) %>%
```
Moohan, line 151: I'm pretty sure `phsmethods::sex_from_chi` already does this?

R/match_on_ltcs.R
```diff
@@ -0,0 +1,171 @@
...
+        changed_dob == 3 ~ "Younger dob gives negative age",
+        changed_dob == 4 ~ "Activity before birth",
+        changed_dob == 5 ~ "LTC before birth",
+        changed_dob == 6 ~ "Maternity record",
+        changed_dob == 7 ~ "GLS record",
+        changed_dob == 8 ~ "Congen at birth",
+        changed_dob == 9 ~ "Unrealistically old age",
+        changed_dob == 10 ~ "Copied from previous record"
+      )
+    )
```
Moohan, line 167: More redundant testing code.

R/match_on_ltcs.R
```diff
@@ -0,0 +1,171 @@
...
+
+  # save the data
+  # write_rds()
```
Moohan, line 170:
```suggestion
  # return the data
  
  return(data)
```

R/match_on_ltcs.R
```diff
@@ -0,0 +1,171 @@
...
+#'
+#' @description Match on LTC changed_dob and dates of LTC incidence (based on hospital incidence only)
+#'
+#' @param
+#'
+#' @return
+#' @export
+#'
+#' @examples
+match_on_ltcs <- function() {
```
Moohan, line 11:
```suggestion
match_on_ltcs <- function(data, year) {
```

R/match_on_ltcs.R
```diff
@@ -0,0 +1,171 @@
...
+    )
+
+  # Match on LTC lookup
+  matched <- dplyr::left_join(temp,
+    readr::read_rds(get_ltcs_path("1920")),
+    by = "chi",
+    suffix = c("", "_ltc")
+  ) %>%
+    # Replace any NA values with 0 for the LTC flags
+    dplyr::mutate(dplyr::across(arth:digestive, ~ tidyr::replace_na(., 0)))
```
Moohan, line 28:
```suggestion
    dplyr::mutate(dplyr::across("arth":"digestive", ~ tidyr::replace_na(., 0)))
```

Moohan (changes requested):
Looking good!

R/match_on_ltcs.R
lizihao-anu: Yes, `dob_from_chi` does.

R/match_on_ltcs.R
```diff
@@ -0,0 +1,100 @@
...
+#'
+#' @description Match on LTC changed_dob and dates of LTC incidence (based on hospital incidence only)
+#'
+#' @param
+#'
+#' @return
+#' @export
+#'
+#' @examples
+match_on_ltcs <- function(data, year) {
```
Moohan, line 11: These will need properly filling out
```suggestion
#' @param data ...
#' @param year ...
#'
#' @return ...
#' @export
match_on_ltcs <- function(data, year) {
```

R/match_on_ltcs.R
```diff
@@ -0,0 +1,100 @@
...
+    by = "chi",
+    suffix = c("", "_ltc")
+  ) %>%
+    # Replace any NA values with 0 for the LTC flags
+    dplyr::mutate(dplyr::across("arth":"digestive", ~ tidyr::replace_na(., 0)))
+  matched <- correct_demographics(matched, year)
+  return(matched)
+}
+
+correct_demographics <- function(data, year) {
```
Moohan, line 25: Can this be split into two separate scripts please.

R/match_on_ltcs.R
```diff
@@ -0,0 +1,100 @@
...
+  # Checking and changing DOB and age
+  # data <- data %>%
+  test <- data %>%
```
Moohan, line 28:
```suggestion
  test <- data %>%
```

Also rename the object from `test` to something meaningful.

R/match_on_ltcs.R
```diff
@@ -0,0 +1,100 @@
...
+        chi_check = FALSE,
+        min_date = lubridate::ymd("1900-01-01"),
+        max_date = lubridate::ymd("1999-12-31")
+      ),
+      # Create a dob in the current century from chi (will return NA if in the future)
+      chi_dob_max = phsmethods::dob_from_chi(
+        chi_number = chi,
+        chi_check = FALSE,
+        min_date = lubridate::ymd("2000-01-01"),
+        max_date = end_fy(year)
```
Moohan, line 43: `max_date` will always be the start of the activity as we know they have to have been born on or before the activity started.

```suggestion
        max_date = pmin(record_keydate1, lubridate::ymd("1999-12-31"))
      ),
      # Create a dob in the current century from chi (will return NA if in the future)
      chi_dob_max = phsmethods::dob_from_chi(
        chi_number = chi,
        chi_check = FALSE,
        min_date = lubridate::ymd("2000-01-01"),
        max_date = record_keydate1
```

R/match_on_ltcs.R
```diff
@@ -0,0 +1,100 @@
...
+        # If they have a GLS record and the age is broadly correct, assume older
+        dplyr::between(chi_age_max, 50, 130) &
+          recid == "GLS" ~ chi_age_max,
+        # If a congenital defect lines up with a dob, assume it is correct
+        chi_dob_max == congen_date ~ chi_age_min,
+        chi_dob_min == congen_date ~ chi_age_max,
+        # If being older makes them over 113, assume they are younger
+        chi_age_max > 113 ~ chi_age_min
+      ),
+      # For the same scenarios, mutate dob
```
Moohan, line 68: I don't think we need to do this twice? We use the 'possible' dates and ages to work out the most likely dob, then use `fill` to fill any remaining, and then just calculate the age from the dob.

```suggestion
      # Derive the most likely DoB based on different scenarios
```

R/match_on_ltcs.R
```diff
@@ -0,0 +1,100 @@
...
+      dob = dplyr::case_when(
+        # Case when the dob is already valid
+        .data$dob == chi_dob_min ~ dob,
+        .data$dob == chi_dob_max ~ dob,
+        # If one option for dob isn't there, use the other
+        is.na(chi_dob_min) &
+          !is.na(chi_dob_max) ~ chi_dob_max, !is.na(chi_dob_min) &
+          is.na(chi_dob_max) ~ chi_dob_min,
+        # If they have an LTC date before birth date, assume older
+        chi_dob_max > purrr::reduce(dplyr::select(., "arth_date":"digestive_date"), `min`) ~ chi_dob_min,
```
Moohan, line 78: I'm not sure if this will work! But if it does I think it reads better 😂
```suggestion
        chi_dob_max > pmin(dplyr::across("arth_date":"digestive_date") ~ chi_dob_min,
```

R/match_on_ltcs.R
```diff
@@ -0,0 +1,100 @@
...
+      )
+    ) %>%
+    # If we still don't have an age, try and fill it in from other records.
+    dplyr::group_by(chi) %>%
+    tidyr::fill(dob, .direction = "downup") %>%
+    dplyr::ungroup() %>%
+    # Fill in ages for any that are left.
+    dplyr::mutate(age = dplyr::if_else(is.na(age), compute_mid_year_age(year, dob), age)) %>%
+    # If any gender codes are missing or 0 recode to CHI gender.
+    dplyr::mutate(chi_gender = phsmethods::sex_from_chi(chi))
```
Moohan, line 96: We need to do something else with this e.g. 
```suggestion
    dplyr::mutate(chi_gender = phsmethods::sex_from_chi(chi),
    gender = dplyr::if_else(!is.na(chi_gender) & (is_missing(gender) | gender == 0L), chi_gender, gender)
```

Moohan (changes requested):
A few more changes, but looking almost done now.

R/match_on_ltcs.R
lizihao-anu: `phsmethods::dob_from_chi` requires `min_date <= max_date` and do not allow `NA`. The column `keydate1_dateformat` contains `NA` in some rows.

So, for line 36, I would write 
```
      max_date = pmin(
        dplyr::if_else(
          is.na(keydate1_dateformat),
          lubridate::ymd("1999-12-31"),
          keydate1_dateformat
        ),
        lubridate::ymd("1999-12-31")
      )
```

For line 43, `keydate1_dateformat`  ( which is essentially record_keydate1 but with an appropriate date format, is it?) can be earlier than `2000-01-01` which is specified as `min_date`. So `dob_from_chi` cannot allow this. Therefore, I guess it would better keep it as ` max_date = keydate1_dateformat`.

Happy to discuss. Thanks

R/match_on_ltcs.R
lizihao-anu: Yeah. It looks much better. Thanks for the great advice

R/match_on_ltcs.R
lizihao-anu: Yeah. Looks good. I also deleted the columns "chi_age_max", etc. 

Just another reflection here. `if_else` requires type stable, meaning values for condtion true or false should be of the same type. In this case, gender was imported as numeric which chi_gender is integer. So it reported an error in `if_else`. Subsequently, I do `mutate(gender = as.integer(gender))` at first. By the way, `base::ifelse` does not require type stable.

lizihao-anu:
I hope it should be really good to merge this time.

R/correct_demographics.R
```diff
@@ -0,0 +1,87 @@
...
+
+      # fix gender
+      chi_gender = phsmethods::sex_from_chi(chi),
+      gender = as.integer(gender),
+      gender = dplyr::if_else(!is.na(chi_gender) &
+        (is.na(gender) | gender == 0L),
+      chi_gender,
+      gender
+      )
+    ) %>%
```
Moohan, line 75:
```suggestion
    # Fill in the ages for any that are left.
    dplyr::mutate(
      age = compute_mid_year_age(year, dob),
) %>% 
      # Fill in gender from CHI if it's missing.
      dplyr::mutate(
      chi_gender = phsmethods::sex_from_chi(chi),
      gender = as.integer(gender),
      gender = dplyr::if_else(!is.na(chi_gender) &
        (is.na(gender) | gender == 0L),
      chi_gender,
      gender
      )
    ) %>%
```

Moohan (approved):

PR #444, opened 2022-11-09
Convert (year specific) Home care script into a function for processing in phase 1b

Jennit07:
Conversion of the year specific home care script. Tested this using the home care all data script and happy this is working as expected. I found one bug in the code which caused `cost_total_net` to return 0 for all rows. This has been corrected in this PR which was caused by a small typo (missing `s` on costs)


lizihao-anu (approved):
The test run looks good

PR #445, opened 2022-11-09
Changes to the PIS data

Moohan:
These changes are related to moving from using dispensed items to paid items.

I've made changes to the tests that will let us look at the changes in more detail.

- [x] Need to get the new extracts from BI with the Paid items and Paid cost in.
- [x] Replace the existing variables in the PIS processing code with the new ones.
- [x] Check what the difference is and decide which to use at this update.


Moohan:
I think this is ready to be checked and merged. We decided to use both Paid items (instead of dispensed) and to move to Gross cost (GIC) from Net cost (NIC), at this update.

Having run the tests, there is a very small decrease in the total number of items (this was expected) and a small (1-2%) increase in costs, both mean and total.

PR #446, opened 2022-11-09
318 add tests for the lookup file path functions

Moohan:
The main change here is to add tests for the lookup file path functions.

I have also added an `ext` parameter to `get_gpprac_ref_path`, and tidied up the documentation by using `@inheritParams` so things stay consistent and just need changing in one place.


tests/testthat/test-get_lookup_paths.R
```diff
@@ -0,0 +1,65 @@
...
+})
+
+test_that("gpprac reference file path returns as expected", {
+  expect_s3_class(get_gpprac_ref_path(), "fs_path")
+
+  expect_equal(fs::path_ext(get_gpprac_ref_path()), "sav")
+
+  expect_error(fs::path_ext(get_gpprac_ref_path(ext = "rds")))
+
+  expect_true(fs::file_exists(get_gpprac_ref_path(ext = "sav")))
```
Jennit07, line 63:
```suggestion

```
Looks good. Only thing i've spotted is that this line is causing the test to fail? It looks good to me so i'm not sure why it is failing. ext = "sav" is declared in the function and the file exists in the folder so i would expect this to pass? Error message below 

`Error in get_gpprac_ref_path(ext = "sav") : unused argument (ext = "sav")`

Jennit07 (changes requested):
One thing causing a test to fail, everything else looks good

tests/testthat/test-get_lookup_paths.R
Moohan: I got this error too, but it started working after I did `load_all()`. Does that resolve it for you?

Jennit07: Tested this again and looks good now!

Jennit07 (approved):

Issue #448, opened 2022-11-11
Add function for dealing with sc start dates 


Jennit07:
The following can be applied to multiple sc code: 

```
  # If sds start date is missing, assign start of FY
  mutate(sds_start_date = if_else(is.na(sds_start_date),
    start_fy(year = period, format = "alternate"),
    sds_start_date
  ))
```

Issue #447, opened 2022-11-11
Add function for dealing with sc end date is earlier than start date


Jennit07:
The following code can be applied to multiple datasets. Create a new function for this

```
 # Fix sds_end_date is earlier than sds_start_date by setting end_date to be the end of fyear
  mutate(sds_end_date = if_else(
    sds_start_date > sds_end_date,
    end_fy(year = period, "alternate"),
    sds_end_date
  )) 
```

Jennit07:
issue #448 in same branch called `fix_sc_dates`

PR #449, opened 2022-11-11
New functions for `fix_sc_dates`

Jennit07:
New functions `fix_sc_start_dates` (Issue #448) and `fix_sc_end_dates` (Issue #447). I've cherry picked these functions onto the `function_sc_alarms_telecare` branch and `function_sc_sds` branch for testing and bringing the code inline to use these. I checked home care and care home to see if this could be applied but there is slightly different logic in those scripts.


Moohan:
Could we add some tests please 😀

Jennit07:
> Could we add some tests please 😀

Tests added

Moohan (approved):

PR #422, opened 2022-10-24
Convert All Alarms Telecare script into a function for processing in phase 1b

Jennit07:
Follows the same format as other functions. First read in data from the SC platform using `read_sc_all_alarms_telecare.R` function and process it using `process_sc_all_alarms_telecare.R`. Some warnings/test failures due to #419

PR #450, opened 2022-11-14
Convert All SDS script into a function for processing in phase 1b

Jennit07:
I've implemented the following changes into the original script before converting this into a function for processing:

- Incorporated the SDS script with the R conversion project (e.g. use of correctly named functions which do not rely on 00-Social-care-functions)
- PR #438 has been applied to fix SDS option 4 bug 
- PR #449 has been applied to use functions for fixing start/end dates
- Use of `data.table::as.data.table()` for speeding up running of code which was also used in PR #422 

I've tested the function and i'm happy with how this is processing the SDS data from the SC platform.


Moohan (approved):
Looks good to me. I just made a tiny change to drop some variables which aren't needed at the end.

PR #451, opened 2022-11-15
cost uplift from C10 to C01

### LINKED ISSUE #345: Move cost uplift code from C10 into C01
Lines (4-8)

This could be put into a function and included in #332

lizihao-anu:
close #345 
Hope it makes sense.


R/cost_uplift.R
```diff
@@ -0,0 +1,27 @@
+#' Uplift costs
+#'
+#' @param data episode data
+#' @param uplift
+#'
+#' @return episode data with uplifted costs
+#' @export
+cost_uplift <- function(data, uplift) {
```
Moohan, line 8: I think we need to take the data and the year here, and then work out the uplift for that year within this function (or in another function.

lizihao-anu: Sorry, I thought of it simplistically. I dug out the uplift rate for each year and created a function to set the uplift. My way is to create a data frame with each year and each year's uplift rate, and then join the data frame with episode data. 

To my best knowledge, it should be the most efficient way to do this compared with manupilate with episode directly. But please let me know if there is a better way. Thanks!

Moohan: Cool, that looks good to me. Just having a proper check / review now, but the concept seems fine 👍

R/cost_uplift.R
```diff
@@ -0,0 +1,78 @@
...
+  return(data)
+}
+
+#' set uplift scale
+#'
+#' @param data episode data
+#' @param latest_cost_year default financial year '2223'
+#'
+#' @return episode data with a uplift scale
+#' @export
```
Moohan, line 39:
```suggestion
```

I don't think this needs exporting as it will only be used in the `cost_uplift` function?

R/cost_uplift.R
```diff
@@ -0,0 +1,78 @@
...
+  # We have set uplifts to use for 2020/21, 2021/22 and 2022/23, provided by Paul Leak.
+  # For older years, don't uplift.
+  # For years after 2022/23 uplift by an additional 1% per year after the latest cost year (2022/23)
+  # For non plics recids use uplift of 1 so we won't change anything.
+
+  # to accelerate, create a data frame of year and uplift for match-joining
+  start_year <- 10
+  end_year <- 70
+  year <- paste0(start_year:end_year, (start_year + 1):(end_year + 1)) %>% as.numeric()
+  uplift_df <- data.frame(year, uplift = 1) %>%
```
Moohan, line 50: Use tibble so we avoid strings turning into factors.
```suggestion
  uplift_df <- tibble::tibble(year, uplift = 1) %>%
```

R/cost_uplift.R
```diff
@@ -0,0 +1,78 @@
...
+#' @export
+uplift_set <- function(data, latest_cost_year = "2223") {
+  # We have set uplifts to use for 2020/21, 2021/22 and 2022/23, provided by Paul Leak.
+  # For older years, don't uplift.
+  # For years after 2022/23 uplift by an additional 1% per year after the latest cost year (2022/23)
+  # For non plics recids use uplift of 1 so we won't change anything.
+
+  # to accelerate, create a data frame of year and uplift for match-joining
+  start_year <- 10
+  end_year <- 70
```
Moohan, line 48: I think this is sufficient to always produce an uplift for required years (because we will never need an upflift for a year in the future).
```suggestion
  start_year <- 10L
  end_year <- as.integer(format(Sys.Date(), "%y"))
```

R/cost_uplift.R
```diff
@@ -0,0 +1,78 @@
...
+  year <- paste0(start_year:end_year, (start_year + 1):(end_year + 1)) %>% as.numeric()
```
Moohan, line 49: `year` can be an integer, and you are checking for it to be an integer below.
```suggestion
  year <- paste0(start_year:end_year, (start_year + 1):(end_year + 1)) %>% as.integer()
```

R/cost_uplift.R
```diff
@@ -0,0 +1,78 @@
...
+}
+
+#' set uplift scale
+#'
+#' @param data episode data
+#' @param latest_cost_year default financial year '2223'
+#'
+#' @return episode data with a uplift scale
+#' @export
+uplift_set <- function(data, latest_cost_year = "2223") {
```
Moohan, line 40: Could we have a funtion for `latest_cost_year` please. See `R/00-update_refs.R`
```suggestion
uplift_set <- function(data, latest_cost_year = latest_cost_year()) {
```
(We probably wouldn't even need it as an argument then.

Moohan (changes requested):
Looking good.

I think the functions need renaming to be a bit more illustrative of what they do. e.g. `apply_cost_uplift` and `lookup_uplift`. Or anything else! @Jennit07 or @shintoLamp might have some ideas for sensible names!

Moohan (approved):

PR #452, opened 2022-11-18
Make client variable as an integer 

Jennit07:
Opening a PR so we can see changes. Originally, this was supposed address #434 however, when i looked into this it looks like `sending_location` is being pulled as a character from the platform. I tested the code without this line in `process_extract_home_care` function

```
    # TODO change the client script to use sending location as an integer
    dplyr::mutate(sending_location = as.character(.data$sending_location)) %>%
``` 
and the code was working fine. If we make the change to have this as an integer, it will mean changing all SC scripts to reflect this. Even if we don't make this change, i found a few wee (minor) bugs which i corrected on this branch:

- moved ` dplyr::filter(.data$financial_year == year) %>%` code into `read_lookup_sc_client` function (which i thought i had already done, but it has slipped through)
- changed lca variable in home care script to `sc_send_lca` which i noticed was missing from the final output


Moohan (approved):
Great work finding and fixing those bugs!

I think we can just leave it as a character for now. Maybe we want to look at changing types later but it seems the simplest thing for now is to just remove that TODO note and the redundant line of code.

PR #454, opened 2022-11-18
Remove code which changes dates to numeric

Moohan:
We decided to just keep dates as dates, at least in the processing scripts/functions. I noticed that this had been overlooked on the MH function. I didn't spot any other places where this was the case.


Jennit07 (approved):
Looks good - happy to merge

PR #460, opened 2022-11-23
WIP add a function to clean up .csv or .gz files

lizihao-anu:
For safety's sake, comment out the deleting commands. I will uncomment them when the moment is right.

Feel free to make any comments or new require new features. Thanks!


R/clean_files_function.R
```diff
@@ -0,0 +1,95 @@
...
+    if (askYesNo("Would you like to delete the files listed above",
+      default = FALSE
+    )) {
+      # calculate the size of files deleted in IT_extracts
+      filesize_it <- round(sum(file.size(it_file_list)) / 1073741824, 2)
+      # sapply(it_file_list, unlink)
+    }
+  } else {
+    print("NOT clean up files in IT_extract")
+  }
```
Moohan, line 42: We generally don't delete the files in the IT extract directory. They should be zipped back up though (gz with maximum compression), so maybe search for any uncompressed csv files and gzip them?

R/clean_files_function.R
```diff
@@ -0,0 +1,95 @@
...
+
+  # delete files in year folder(s)
+  if (is.null(clean_years)) {
+    print("NOT clean up files in any year folder")
```
Moohan, line 46: Use `cli` for messages e.g. [`cli::cli_alert_info()`](https://cli.r-lib.org/reference/cli_alert.html) or [`cli::cli_inform()`](https://cli.r-lib.org/reference/cli_abort.html)

R/clean_files_function.R
```diff
@@ -0,0 +1,95 @@
...
+  } else {
+    if (tolower(clean_years) == "all") {
```
Moohan, line 48: Maybe use `match.arg` at the top, which can check that only correct values ("all" or a valid year) have been submitted, then at this point it will already match.

```r
if (!missing(clean_years)) {
  available_years <- fs::dir_ls("/conf/sourcedev/Source_Linkage_File_Updates/",
    regexp = "\\d{4}",
    type = "directory"
  ) %>%
    stringr::str_extract("\\d{4}$")

  clean_years <- match.arg(clean_years, choices = c("all", available_years), several.ok = TRUE)
}
```

```suggestion
    if (clean_years == "all") {
```

R/clean_files_function.R
```diff
@@ -0,0 +1,95 @@
...
+      year <- paste0(
+        11:as.integer(format(Sys.Date(), "%y")),
+        12:(as.integer(format(
+          Sys.Date(), "%y"
+        )) + 1)
+      )
```
Moohan, line 54:
```suggestion
      year <- fs::dir_ls("/conf/sourcedev/Source_Linkage_File_Updates/",
    regexp = "\\d{4}",
    type = "directory"
  ) %>%
    stringr::str_extract("\\d{4}$")
```

R/clean_files_function.R
```diff
@@ -0,0 +1,95 @@
...
+
+    print(year_file_list)
+    if (askYesNo("Would you like to delete the files listed above",
+      default = FALSE
+    )) {
+      # calculate file size deleted in year folders
+      filesize_year <- round(sum(file.size(year_file_list)) / 1073741824, 2)
+      # sapply(year_file_list, unlink)
+    }
+  }
```
Moohan, line 78: We would probably want to split this into a few chunks:

- Extracts - The csv / csv.gz files in the `/Extracts` folder.
- Production files - The zsav (or rds) files which should be zipped into `Activity*.zip`
- The Production test files (maybe grouped with the above), won't be needed after R conversion. These should all be zipped into `BXX Tests.zip`.
- The actual linkage files; `source-episode-*` and `source-individual-*`. They are produced here, then copied to hscdiip once they've been checked.

You could ask the user about each chunk in turn and create a list of files to delete, then delete them all in one go.

Moohan (commented):
Nice start, I've left a few comments. I think the command line output could be improved with [`cli`](https://cli.r-lib.org/index.html), or possibly using a package like [`ask`](https://github.com/gaborcsardi/ask).

Jennit07:
Agree with James' comments, nothing else to add

PR #462, opened 2022-11-23
Transform `anon_chi` from the episode file to `chi`

Moohan:
This small fix transforms the `anon_chi` back to the real `chi`. Previously we were just renaming `anon_chi` to `chi` in the tests but this meant that `create_demog_test_flags` wasn't able to make a proper comparison of `valid_chi` since all of the `anon_chi` from the episode file would fail `phsmethods::chi_check`.


Jennit07:
Not sure if it's just my session or the data (AE2) im using but the code doesnt seem to return an output for me. I've done a `load_all()` and restarted my session. Any other ideas?

Moohan:
> Not sure if it's just my session or the data (AE2) I'm using but the code doesn't seem to return an output for me. I've done a `load_all()` and restarted my session. Any other ideas?

What do you mean by 'the code' doesn't return an output? The test function should just write the tests out and return the path to the test files. You should see the change that this code has made in the xlsx test file.

Jennit07:
> > Not sure if it's just my session or the data (AE2) I'm using but the code doesn't seem to return an output for me. I've done a `load_all()` and restarted my session. Any other ideas?
> 
> What do you mean by 'the code' doesn't return an output? The test function should just write the tests out and return the path to the test files. You should see the change that this code has made in the xlsx test file.

When i run the `process_tests_ae' function to try to write the ae tests, my R session seems to lock up/run for a very long time and does not show me an output if i assign to a 'check dataframe' or write the tests to disk. Happy to have a call to discuss if it is easier

Moohan:
> > > Not sure if it's just my session or the data (AE2) I'm using but the code doesn't seem to return an output for me. I've done a `load_all()` and restarted my session. Any other ideas?
> > 
> > 
> > What do you mean by 'the code' doesn't return an output? The test function should just write the tests out and return the path to the test files. You should see the change that this code has made in the xlsx test file.
> 
> When i run the `process_tests_ae' function to try to write the ae tests, my R session seems to lock up/run for a very long time and does not show me an output if i assign to a 'check dataframe' or write the tests to disk. Happy to have a call to discuss if it is easier

I've found and resolved a bug, although it doesn't sound the same as you have been having.

Do you want to check the latest code then give me a call tomorrow if it's still got issues?

Jennit07 (approved):
Tried this again after testing the new code and it looks like it's working now. No idea what was wrong with it in the first place as used ae data and the same tests function again which worked fine. Happy to approve

PR #463, opened 2022-11-23
Fix homelessness filtering when we don't have completeness data

Moohan:
All 22/23 was being filtered (except East Ayrshire) as we don't have completeness data for 22/23.

This changes checks to see what the latest year we have completeness data for, and will only filter for those years.


Jennit07 (approved):
Looks good

PR #414, opened 2022-10-12
Service use cohort 

shintoLamp:
Set of functions to create the service use cohort lookup. 

Tested against actual data and I'm happy with the results, but the logic of the helper functions probably needs to be double-checked against the SPSS syntax.

PR #464, opened 2022-11-28
Rename demographic cohort functions

Jennit07:
Opening this PR to keep consistent with #414 - Renaming of demographic cohort functions to prevent any conflicts in the documentation. Follows the format `assign_d_cohort_xxx`


Moohan (approved):
The only issue was that the rename had missed the tests. I also re-jigged some of the documentation so it's not producing a file for every single (sub)function.

Could you have a check-through again before you merge?

PR #467, opened 2022-12-06
C01 create episode file

Moohan:
Just opening a draft so we can keep track of what's going on with this branch.


Jennit07 (approved):
Merging this into mainR and then going to take this onto the `test-production-ep-file` branch to test the episode file using targets

Issue #479, opened 2022-12-09
Convert (year specific) care homes into a function for processing


Jennit07:
The care home scripts found here should follow the format for other functions:

Source-Linkage-Files/Production_scripts/Process_extracts

- [x] extract_care_homes
- [x] extract_care_homes_tests

Issue #477, opened 2022-12-09
Investigate non-matching SPSS vs R test outputs


Jennit07:
Please check and update the spreadsheet on teams with the findings. Check instances where the SPSS and R versions do not match cost tests: 

- [x] Outpatients
- [x] CMH 
- [x] Care Home

lizihao-anu:
- Outpatients
The issue was costs from SPSS and R do not match. It is because SPSS outputs do not add each month costs to cost_net_total properly while R outputs do.

- CMH
In year 1718, more than 700 records in R outputs have missing postcodes, leading to less episode numbers in health boards.

lizihao-anu:
See excel spreedesheet for more details

PR #489, opened 2022-12-13
Convert (year specific) Care home into a function for processing

Jennit07:
Addresses #479 and converts two scripts into a function:

` extract_care_homes.R`  and  `extract_care_homes_tests`

I have compared the R version to SPSS and updated the spreadsheet on teams. The output looks good with all variables and the number of rows matching. However there are some differences in beddays and costs compared to the SPSS version. I have added CH to issue #477 for checking other non-matching SPSS vs R outputs.

I have also tidied up some scripts on this branch e.g. HC and AT scripts in the production folder which are not needed as these are already converted into a function.


R/process_extract_care_home.R
```diff
@@ -0,0 +1,119 @@
...
+    dplyr::left_join(client_lookup, by = c("sending_location", "social_care_id"))
+
+
+  # Data Cleaning ---------------------------------------
+  source_ch_clean <- ch_data %>%
+    # create variables
+    dplyr::mutate(
+      year = year,
+      recid = "CH",
+      smrtype = "Care-Home"
```
Moohan, line 53: Implement this in the `smrtype` function, please.

R/process_extract_care_home.R
```diff
@@ -0,0 +1,119 @@
...
+    ) %>%
+    # compute lca variable from sending_location
+    dplyr::mutate(sc_send_lca = convert_sending_location_to_lca(.data$sending_location)) %>%
+    # bed days
+    create_monthly_beddays(year, .data$record_keydate1, .data$record_keydate2) %>%
+    # year stay
+    dplyr::mutate(
+      yearstay = rowSums(dplyr::across(tidyselect::ends_with("_beddays"))),
+      # total length of stay
+      stay = calculate_stay(year, .data$record_keydate1, .data$record_keydate2, .data$sc_latest_submission)
```
Moohan, line 63: I think using named parameters in this case (and possibly others) makes it easier to see what's going on.
```suggestion
      stay = calculate_stay(year,
       start_date = .data$record_keydate1,
       end_date =  .data$record_keydate2,
        sc_qtr =  .data$sc_latest_submission)
```

Moohan (changes requested):
I just did a quick first pass. Looking good!

R/process_extract_care_home.R
Jennit07: Done

R/process_extract_care_home.R
Jennit07: Done, also implemented for calculate beddays function

Jennit07:

Issue #475, opened 2022-12-09
Compare SPSS vs R outputs 


Jennit07:
Still to do and update the spreadsheet on teams: 

- [x] Care Homes (year specific)
- [ ] ALL Care Homes (on hold, still WIP)
- [ ] GP OOH (on hold - waiting on approved PR)
- [ ] Delayed Discharges 
- [x] DN
- [x] Home Care 
- [x] A & T 
- [x] SDS

PR #505, opened 2022-12-14
Convert DN tests into a function for processing

### LINKED ISSUE #494: Convert District Nursing tests into a function for processing
/production_scripts/process_extracts/produce_extract_tests

Jennit07:
I've tested this through and happy it is working. I have compared the R output vs SPSS (#475) which looks like it is matching so i have updated the spreadsheet accordingly. 

Assigned as low priority for checking, however i am prioritising the remaining scripts>functions before i stop, so there will be multiple similar PRs open.

Closes #494


R/process_tests_district_nursing.R
```diff
@@ -0,0 +1,22 @@
...
+#' @export
+#'
+process_tests_district_nursing <- function(data, year) {
+  old_data <- get_existing_data_for_tests(data)
+
+  comparison <- produce_test_comparison(
+    old_data = produce_source_dn_tests(old_data),
+    new_data = produce_source_dn_tests(data)
+  ) %>%
+    write_tests_xlsx(sheet_name = "DN", year)
```
Moohan, line 19:
```suggestion
    write_tests_xlsx(sheet_name = "dn", year)
```

Moohan (approved):

Issue #506, opened 2022-12-14
ToDo - Run the ALL CH episodes script to create Dec22 file


Jennit07:
As the ALL CH script is still WIP, the R version has not been ran. To test the ALL CH tests function i have read in the SPSS Dec version and saved this as an rds version just so that i can test that the output for the tests is working correctly.

Issue #472, opened 2022-12-09
Improve extract tests


Jennit07:
Following comparison of SPSS vs R outputs there are a few small bugs to be addressed. 

In extract test scripts it was identified that we were missing some tests which would be useful:
- [ ] Add HB breakdown for beddays
- [ ] Add n_episodes

These should be added to all applicable test scripts, including for beddays only when available

Jennit07:
Add n_episodes to demographics function

Jennit07:
We can address this one when we review the tests workbooks

PR #507, opened 2022-12-14
Convert all SC CH tests into a function

### LINKED ISSUE #495: Convert ALL sc CH tests into a function for processing
/production_scripts/process_extracts/produce_extract_tests

Jennit07:
This test compares the 'old' ALL sc CH episodes to the 'new' extract. I think this is missing some tests such as n_episodes which would be useful (issue open #472). I have tested this using the SPSS outputs which i have read into R and then saved as an RDS version which seems to be working. However we will need to run the ALL sc CH script when this is available (still WIP, Issue open #506) 

Assigned as low priority for checking, however i am prioritising the remaining scripts>functions before i stop, so there will be multiple similar PRs open.

Closes #495


R/process_tests_sc_ch_episodes.R
```diff
@@ -0,0 +1,17 @@
...
+#' @export
+#'
+process_tests_sc_ch_episodes <- function() {
+  comparison <- produce_test_comparison(
+    old_data = produce_sc_ch_episodes_tests(readr::read_rds(get_sc_ch_episodes_path(update = previous_update()))),
+    new_data = produce_sc_ch_episodes_tests(readr::read_rds(get_sc_ch_episodes_path()))
+  ) %>%
+    write_tests_xlsx(sheet_name = "all_ch_episodes")
+
+  return(comparison)
```
Moohan, line 16: To check: I think `comparison` will be the path to the Excel workbook (because that's what `write_tests_xlsx()` returns).

Options are:

1. Leave it as it is (but change the documentation slightly so that the path to the workbook will be returned.
2. Return the comparison from this function.
```suggestion
  comparison <- produce_test_comparison(
    old_data = produce_sc_ch_episodes_tests(readr::read_rds(get_sc_ch_episodes_path(update = previous_update()))),
    new_data = produce_sc_ch_episodes_tests(readr::read_rds(get_sc_ch_episodes_path()))
  )
  
  write_tests_xlsx(comparison, sheet_name = "all_ch_episodes")

  return(comparison)
```
3. Or modify `write_tests_xlsx()` to return the comparison (i.e. the actual data / tibble), which would mean this function doesn't need changing.

I think I prefer option 3 but it's obviously the most far-reaching change that would affect other bits of code, possibly that's fine though.

Moohan (approved):
Looks good. I'm approving but please have a look at my comment on documentation / return values.

R/process_tests_sc_ch_episodes.R
Jennit07: Had a call with @Moohan and made these changes to other sc functions

Jennit07:
Changes to documentation made

R/process_tests_sc_ch_episodes.R
```diff
@@ -0,0 +1,19 @@
...
+#'
+#' @description This script takes the processed demographic extract and produces
+#' a test comparison with the previous data. This is written to disk as a csv.
+#'
+#' @return a [tibble][tibble::tibble-package] containing a test comparison.
+#' @export
+#'
+process_tests_sc_ch_episodes <- function() {
+  comparison <- produce_test_comparison(
+    old_data = produce_sc_ch_episodes_tests(readr::read_rds(get_sc_ch_episodes_path(update = previous_update()))),
```
Moohan, line 11:
```suggestion
    old_data = produce_sc_ch_episodes_tests(
    readr::read_rds(get_sc_ch_episodes_path(update = previous_update()))
    ),
```

R/process_tests_sc_ch_episodes.R
```diff
@@ -0,0 +1,19 @@
...
+    new_data = produce_sc_ch_episodes_tests(readr::read_rds(get_sc_ch_episodes_path()))
```
Moohan, line 12:
```suggestion
    new_data = produce_sc_ch_episodes_tests(
    readr::read_rds(get_sc_ch_episodes_path())
    )
```

Moohan (approved):

PR #510, opened 2022-12-15
Add new functions to get the path of the temp files

Moohan:
These can / will be used for testing the 'make episode' and 'make individual' files code.

At the moment we can either use `slfhelper` to read in the final file (with select variables), or use `haven` to read the zsav temporary versions that SPSS produced as part of the process.

I have re-created the temp `.zsav` files as `.rds` files and these functions get the file paths to them.

There are some error messages which should hopefully make it easier to find the file you need.


lizihao-anu (approved):
Looks great. I have tested and it works.

PR #525, opened 2023-01-10
Convert GP OOH script into a function for processing

Jennit07:
I've converted GP OOH into a function as we have discussed which consists of:  
- a function for reading each OOH extract (x3)
- a function for processing each OOH extract (x3)
- a higher level function for reading the three extracts as a data list (x1)
- a top-level function which calls upon all processed data which is then joined together/cleaned up to produce the final file (x1)

I have tested this and am happy this is working - I compared this with the first run I did of the extract the other day and I get a matching output as expected. I have fixed all documentation notes and I think I'm happy with the documentation. There is one note I cannot get rid of and I am not too sure what this means: 

```
  get_slf_temp_path: no visible binding for global variable ‘.’
  Undefined global functions or variables:
    .
``` 
Any ideas to fix this? There are tests which are failing due to capitalisation of the demographic and service use cohorts - should I rename these in the folder? I'm happy to look at both of these again to correct them! Thanks


Moohan (commented):
Looking good, just some notes on the documentation which I'll commit then re-review.

Moohan (approved):

PR #533, opened 2023-01-20
Add `file_path` parameter to `read_XXX` functions

Jennit07:
This PR deals with adding a file path parameter to functions for use with targets which I've split out from the test-production branch. I've tested this as i have went along and i'm happy this is working - it is also working quite nicely with targets. Mainly this deals with `read_extract` and `read_lookups` functions. I've left out social care just now as i think these may require more tests since they already have a parameter for reading from the sc platform.


R/read_extract_acute.R
```diff
@@ -1,15 +1,13 @@
...
+#' @param year Year of BOXI extract
+#' @param file_path BOXI extract location
 #'
 #' @return a [tibble][tibble::tibble-package].
 #'
-#' @export
-read_extract_acute <- function(year) {
-  extract_acute_path <- get_boxi_extract_path(year = year, type = "Acute")
-
+read_extract_acute <- function(year, file_path = get_boxi_extract_path(year = year, type = "Acute")) {
```
Moohan, line 8: I think we want this one to still be exported.
```suggestion
#' @export
read_extract_acute <- function(year, file_path = get_boxi_extract_path(year = year, type = "Acute")) {
```

Moohan (changes requested):
Looks good, I'll see about testing it on Monday!

Only notes are on documentation - obviously you're adding a new param to all the functions so that needs to be documented but a lot of the documentation looks like a regression from what we had before... Check the file changed tab but things like using `@inherit ...` when lots of functions share the same params make things easier to maintain.

Another note is that the return type of these is a always a `tibble`, not a csv (which is a type of file, not an R object type 😉)

Jennit07:
> Looks good, I'll see about testing it on Monday!
> 
> Only notes are on documentation - obviously you're adding a new param to all the functions so that needs to be documented but a lot of the documentation looks like a regression from what we had before... Check the file changed tab but things like using `@inherit ...` when lots of functions share the same params make things easier to maintain.
> 
> Another note is that the return type of these is a always a `tibble`, not a csv (which is a type of file, not an R object type 😉)

Good spot! I think what may have happened was that my test-production branch wasnt up to date and had the old format for parameters. I can see in the changes where this has happened so i have pushed a fix for this just now

Moohan (approved):

PR #536, opened 2023-01-23
Restyle code

Moohan:
I came across a tool/package called [codegrip](https://github.com/lionel-/codegrip) which made it super easy to change the layout of the code. I've used it to restyle many lines to make them shorter (the guideline is keeping code lines under 80 chars).

I've avoided any files that I know are currently undergoing PRs to minimise confusion. I've also avoided Social Care files as I know we're still checking them against the previous update.

**To check** Before merging please make sure:

- [x] I've not introduced any regressions etc. This PR should only be changing code style (and a couple of typos).
- [x] I'm not editing any files which currently have open PRs on them.

### DIFF
--- a/R/create_hb_test_flags.R
+++ b/R/create_hb_test_flags.R
@@ -11,19 +11,67 @@
 create_hb_test_flags <- function(data, hb_var) {
   data <- data %>%
     dplyr::mutate(
-      NHS_Ayrshire_and_Arran = dplyr::if_else({{ hb_var }} == "S08000015", 1L, 0L),
+      NHS_Ayrshire_and_Arran = dplyr::if_else(
+        {{ hb_var }} == "S08000015",
+        1L,
+        0L
+      ),
       NHS_Borders = dplyr::if_else({{ hb_var }} == "S08000016", 1L, 0L),
-      NHS_Dumfries_and_Galloway = dplyr::if_else({{ hb_var }} == "S08000017", 1L, 0L),
+      NHS_Dumfries_and_Galloway = dplyr::if_else(
+        {{ hb_var }} == "S08000017",
+        1L,
+        0L
+      ),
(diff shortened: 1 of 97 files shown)


R/create_hb_test_flags.R
```diff
@@ -11,19 +11,67 @@
 create_hb_test_flags <- function(data, hb_var) {
   data <- data %>%
     dplyr::mutate(
-      NHS_Ayrshire_and_Arran = dplyr::if_else({{ hb_var }} == "S08000015", 1L, 0L),
+      NHS_Ayrshire_and_Arran = dplyr::if_else(
+        {{ hb_var }} == "S08000015",
+        1L,
+        0L
+      ),
       NHS_Borders = dplyr::if_else({{ hb_var }} == "S08000016", 1L, 0L),
```
Jennit07, line 19: Borders and FV have not changed on here, should these also reshape? I tried to install code grip and run the reshape but it also did not change so maybe it is not an issue?

R/process_extract_outpatients.R
```diff
@@ -30,8 +30,6 @@ process_extract_outpatients <- function(data, year, write_to_disk = TRUE) {
       # Set smrtype variable
       smrtype = add_smr_type(.data$recid)
     ) %>%
-    # Recode GP Practice into a 5 digit number
-    # assume that if it starts with a letter it's an English practice and so recode to 99995
     dplyr::mutate(gpprac = convert_eng_gpprac_to_dummy(.data$gpprac)) %>%
```
Jennit07, line 33: Just checking if we wanted to remove these notes here?

Jennit07 (approved):
This looks good to me! I cant see any regressions and checked the open PR's. Happy to approve, but there are two comments i have made just to clarify

R/process_extract_outpatients.R
Moohan: Yep, I don't think that's needed as it's just describing what the single function does.

PR #542, opened 2023-01-25
SDS year fix

lizihao-anu:
Only the first push fixes the issue. The last two push need discretion to decide whether to inlcude them.


All_years/04-Social_Care/04-SDS_data.R
```diff
@@ -76,7 +76,7 @@ sds_full_clean <- sds_full_data %>%
   )) %>%
   # Fix sds_end_date is earlier than sds_start_date by setting end_date to be the end of fyear
   mutate(sds_end_date = if_else(
-    sds_start_date >= sds_end_date,
+    sds_start_date > sds_end_date,
```
lizihao-anu, line 79: I remember I had a conversation with James, and we decide use > rather than >= in fix_sc_end_date.R. So, just to align the logic here, for consistence. This fix may affect around 4888 out of 58000 records.

R/get_source_extract_path.R
```diff
@@ -46,7 +46,7 @@ get_source_extract_path <- function(year,
     type == "AT" ~ "Alarms-Telecare-for-source",
     type == "CH" ~ "care_home_for_source",
     type == "CMH" ~ "CMH_for_source",
-    type == "Client" ~ "Client_for_source",
+    type == "Client" ~ "client_for_source",
```
lizihao-anu, line 49: I guess the file name is now in lower case as I cannot fetch the file with the capital letter

R/process_extract_sds.R
```diff
@@ -36,8 +36,9 @@ process_extract_sds <- function(data = NULL, year, client_lookup = NULL, write_t
 
   outfile <- data %>%
     dplyr::filter(is_date_in_fyyear(
-      convert_year_to_fyyear(substr(.data$sc_latest_submission, 1L, 4L)),
-      .data$record_keydate1, .data$record_keydate2
```
lizihao-anu, line 40: This filter returns true for every single entry as it checks whether the record submission date within the start date and end date, which is always true. What we want is to check whether the financial year lays between dates.

R/process_extract_alarms_telecare.R
```diff
@@ -34,11 +34,9 @@ process_extract_alarms_telecare <- function(data = NULL, year, client_lookup = N
 
   # Now select epsiodes for given FY
   outfile <- data %>%
-    dplyr::filter(is_date_in_fyyear(
-      convert_year_to_fyyear(substr(.data$sc_latest_submission, 1L, 4L)),
-      .data$record_keydate1,
-      .data$record_keydate2
-    )) %>%
```
lizihao-anu, line 41: the same issue of a&t with sds

All_years/04-Social_Care/04-SDS_data.R
Moohan: Good to know, that's a definite difference and should be `sds_start_date > sds_end_date` as you say.

R/process_extract_sds.R
Moohan: Great find, your change is correct and it looks like this 'error' also applies to Alarms and Telecare.

Moohan (commented):
@lizihao-anu With the two errors you've identified do those changes combined make up for **all* of the difference we're seeing?

If so, amazing work, and happy to approve the changes. It also looks like these changes should be applied to Alarms & Telecare (in a new PR) and will hopefully account for all the differences we see there too!

lizihao-anu:
> @lizihao-anu With the two errors you've identified do those changes combined make up for *_all_ of the difference we're seeing?
> 
> If so, amazing work, and happy to approve the changes. It also looks like these changes should be applied to Alarms & Telecare (in a new PR) and will hopefully account for all the differences we see there too!

Yes, both A&T and SDS can be fixed by this patch.

Moohan (approved):

PR #543, opened 2023-01-27
Amend extracts to return `null` when there is no year available

Jennit07:
Quick PR - Amending the code on extracts to return `null` if this is not available for the given year. Affects CMH, DN, DD and Homelessness extracts which do not have extracts for older years. Tested this with running targets and i'm happy this is working so bringing this into `main-R`. 

Note that this same concept should be applied to social care functions as this applies to 2017 onwards. However there are some issues with trying to use targets and reading the data from the sc platform so i have left these for now.


R/process_extract_cmh.R
```diff
@@ -21,56 +21,62 @@ process_extract_cmh <- function(data,
   # Check that the supplied year is in the correct format
   year <- check_year_format(year)
 
-  # Data Cleaning  ---------------------------------------
+  # If data is available in the FY then run processing.
+  # If no data has passed through, return NULL.
+  if (is.null(data)) {
+    outfile <- NULL
+  } else {
```
Moohan, line 28: I think this would make the code read a bit easier?

```suggestion
  # If no data has passed through, return NULL.
  if (is.null(data)) {
    return(NULL)
  } 
```
(You would also need to delete the last `}`)

R/process_extract_delayed_discharges.R
```diff
@@ -22,91 +22,96 @@ process_extract_delayed_discharges <- function(data,
   # Check that the supplied year is in the correct format
   year <- check_year_format(year)
 
-  # Data Cleaning---------------------------------------
+  # Specify years available for running
+  if (year >= 1617) {
```
Moohan, line 26: Again a simplification is probably something like 

```suggestion
  # Specify years available for running
  if (year < 1617) {
return(NULL)
}

... rest of code.
```

Moohan (changes requested):
Just two small changes which I think will apply to all the functions but should make the code more readable and will definitely make the changes easier to parse.

Jennit07:
Just realised i need to check the higher level extracts on this branch as well. e.g. `get_file_path` and `get_boxi_extract_path`

R/process_extract_delayed_discharges.R
Jennit07: Thanks James, I agree, those look much easier to read!

R/read_extract_cmh.R
```diff
@@ -5,6 +5,11 @@
 #' @export
 #'
 read_extract_cmh <- function(year, file_path = get_boxi_extract_path(year = year, type = "CMH")) {
+  # Specify years available for running
+  if (year >= 1617 & year <= 2021) {
+    return(NULL)
+  }
```
Moohan, line 11: This should be a *NOT* `!`. Also I think it's better to specify the years explicitly for CMH (and DN) as they are now fixed and won't change.

```suggestion
  if (!(year %in% c("1617", "1718", "1819", "1920", "2021"))) {
    return(NULL)
  }
```

I think `year` should be string? Please change to integers if not (`1617L` etc.)

R/process_extract_delayed_discharges.R
```diff
@@ -22,6 +22,10 @@ process_extract_delayed_discharges <- function(data,
   # Check that the supplied year is in the correct format
   year <- check_year_format(year)
 
+  # Specify years available for running
+  if (year < 1617) {
+    return(NULL)
+  }
```
Moohan, line 28: Shouldn't this be checking `data` rather than `year`.

```suggestion
  if (is.null(data)) {
    return(NULL)
  }
```

R/read_extract_district_nursing.R
```diff
@@ -3,9 +3,13 @@
...
 #' @export
-read_extract_district_nursing <- function(
-    year,
-    file_path = get_boxi_extract_path(year = year, type = "DN")) {
+read_extract_district_nursing <- function(year,
+                                          file_path = get_boxi_extract_path(year = year, type = "DN")) {
+  # Specify years available for running
+  if (year >= 1516 & year <= 2021) {
+    return(NULL)
+  }
```
Moohan, line 11: As with CMH
```suggestion
  if (!(year %in% c("1516", "1617", "1718", "1819", "1920", "2021"))) {
    return(NULL)
  }
```

R/read_extract_homelessness.R
```diff
@@ -3,9 +3,13 @@
 #' @inherit read_extract_acute
 #'
 #' @export
-read_extract_homelessness <- function(
-    year,
-    file_path = get_boxi_extract_path(year = year, type = "Homelessness")) {
+read_extract_homelessness <- function(year,
+                                      file_path = get_boxi_extract_path(year = year, type = "Homelessness")) {
+  # Specify years available for running
+  if (year < 1617) {
```
Moohan, line 9:
```suggestion
  if (year < "1617") {
```

Moohan (changes requested):
A couple more changes 🙂

R/read_extract_cmh.R
```diff
@@ -5,6 +5,11 @@
 #' @export
 #'
 read_extract_cmh <- function(year, file_path = get_boxi_extract_path(year = year, type = "CMH")) {
+  # Specify years available for running
+  if (year >= 1617 & year <= 2021) {
```
Jennit07, line 9: if file_path is null then return null. This is because we are going to do this in get_boxi_extract_path and do this in one place

R/process_extract_cmh.R
```diff
@@ -21,6 +21,11 @@ process_extract_cmh <- function(data,
   # Check that the supplied year is in the correct format
   year <- check_year_format(year)
 
+  # If data is available in the FY then run processing.
+  # If no data has passed through, return NULL.
+  if (is.null(data)) {
+    return(NULL)
```
Jennit07, line 27: return NA and check for `is.NA`

R/read_extract_cmh.R
Jennit07: I've now committed `check_year_valid` to this branch which checks for the years. This means that if `is.na(file_path` then `return(NULL)` for passing through functions

R/read_extract_homelessness.R
Jennit07: The string will apply to Delayed discharges only where this is supplied by the team and not via BOXI extract

R/read_extract_district_nursing.R
Jennit07: We discussed changing the `read_XXX` extracts to `return(NULL)` when `is.na(file_path)` is supplied as this is now passed through `check_year_valid`. However, do you think we should update this function with this suggestion?

R/process_extract_delayed_discharges.R
Jennit07: I think this should be year here as the DD extract using `read_extract_delayed_discharges` will pass through data for all years and so we want to filter on this?

R/read_extract_district_nursing.R
Moohan: Yes, I think only the file path functions should be checking the year, the other ones should check a) is the file path `NA` or b) is the `data` `NULL`.

R/process_extract_delayed_discharges.R
Moohan: Yep, you're right!

R/get_boxi_extract_path.R
```diff
@@ -13,38 +13,42 @@
...
+             "Outpatients"
            )) {
     type <- match.arg(type)
 
     year_dir <- get_year_dir(year, extracts_dir = TRUE)
 
+    if (!check_year_valid(year, type)) {
+      return(NA)
+    }
+
```
Moohan, line 37: I think we should also just put this function / code block in `get_source_extract_path()` now.

Moohan (commented):
Looks good, just check the build errors / warnings please.

Jennit07:
@Moohan This one is ready to merge if you're happy with it. I've made all changes, fixed build notes and added some more tests to `check_year_valid` function :)

R/read_extract_district_nursing.R
```diff
@@ -3,9 +3,13 @@
 #' @inherit read_extract_acute
 #'
 #' @export
-read_extract_district_nursing <- function(
-    year,
-    file_path = get_boxi_extract_path(year = year, type = "DN")) {
+read_extract_district_nursing <- function(year,
+                                          file_path = get_boxi_extract_path(year = year, type = "DN")) {
+  # Specify years available for running
```
Moohan, line 8:
```suggestion
```

R/read_extract_homelessness.R
```diff
@@ -3,9 +3,13 @@
 #' @inherit read_extract_acute
 #'
 #' @export
-read_extract_homelessness <- function(
-    year,
-    file_path = get_boxi_extract_path(year = year, type = "Homelessness")) {
+read_extract_homelessness <- function(year,
+                                      file_path = get_boxi_extract_path(year = year, type = "Homelessness")) {
+  # Specify years available for running
```
Moohan, line 8:
```suggestion
```

Moohan (approved):

PR #544, opened 2023-01-27
Add `file_path` parameter to `read_XXX` functions

Jennit07:
Additional PR carrying on from #533. This includes GP OOH which was more complex/needed testing. I have now set this up in targets which is working well with the new `file_path` parameter. This also includes the source pc and gpprac lookups which have also been amended to include `file_path` as a parameter.


R/process_lookup_gpprac.R
```diff
@@ -4,21 +4,22 @@
 #' gpprac lookup, it will return the final data
 #' but also write this out as a zsav and rds.
 #'
+#' @param file_path Link to phs open dataset
```
Moohan, line 7: I don't think like this - I see why you've done it but, this isn't a link to the open data (arguably that would be `"gp-practice-contact-details-and-list-sizes"`)... Do we need to have this as a target? What do we gain from having it as a target?

I think we can just leave it as it was before, so the only parameter (and therefore target) is `write_to_disk`. Happy to have a chat though!

R/process_lookup_postcode.R
```diff
@@ -4,18 +4,19 @@
...
+#' @param file_path Path to spd lookup.
 #' @param write_to_disk (optional) Should the data be written to disk default is
 #' `TRUE` i.e. write the data to disk.
 #'
 #' @return the final data as a [tibble][tibble::tibble-package].
 #' @export
 #' @family process extracts
 
-process_lookup_postcode <- function(write_to_disk = TRUE) {
+process_lookup_postcode <- function(file_path = get_spd_path(), write_to_disk = TRUE) {
```
Moohan, line 15: The simd and locality files should also be parameters (and therefore targets). This function should be re-run if any files change on disk, so we need to have them all as targets.

R/read_extract_ooh_consultations.R
```diff
@@ -1,17 +1,11 @@
...
+#' @inherit read_extract_acute
 #'
 #' @return a [tibble][tibble::tibble-package] with OOH Consultations extract data
-read_extract_ooh_consultations <- function(year) {
-  extract_consultations_path <- get_boxi_extract_path(
-    year = year,
-    type = "GP_OoH-c"
-  )
-
+read_extract_ooh_consultations <- function(year, file_path = get_boxi_extract_path(year = year, type = "GP_OoH-c")) {
```
Moohan, line 6:
```suggestion
read_extract_ooh_consultations <- function(
year, 
file_path = get_boxi_extract_path(year = year, type = "GP_OoH-c"))
 {
```

R/read_extract_gp_ooh.R
```diff
@@ -9,11 +9,14 @@
 #' @return the final data as a [tibble][tibble::tibble-package].
 #' @export
 #' @family process extracts
-read_extract_gp_ooh <- function(year) {
+read_extract_gp_ooh <- function(year,
+                                diagnosis_path = get_boxi_extract_path(year = year, type = "GP_OoH-d"),
+                                outcomes_path = get_boxi_extract_path(year = year, type = "GP_OoH-o"),
+                                consultations_path = get_boxi_extract_path(year = year, type = "GP_OoH-c")) {
```
Moohan, line 15:
```suggestion
read_extract_gp_ooh <- function(
year,
                                diagnosis_path = get_boxi_extract_path(year = year, type = "GP_OoH-d"),
                                outcomes_path = get_boxi_extract_path(year = year, type = "GP_OoH-o"),
                                consultations_path = get_boxi_extract_path(year = year, type = "GP_OoH-c")) 
                                {
```

R/read_extract_ooh_diagnosis.R
```diff
@@ -1,16 +1,12 @@
...
 #'
 #' @return a [tibble][tibble::tibble-package] with OOH Diagnosis extract data
-read_extract_ooh_diagnosis <- function(year) {
-  extract_diagnosis_path <- get_boxi_extract_path(
-    year = year,
-    type = "GP_OoH-d"
-  )
-
+#'
+read_extract_ooh_diagnosis <- function(year, file_path = get_boxi_extract_path(year = year, type = "GP_OoH-d")) {
```
Moohan, line 7:
```suggestion
read_extract_ooh_diagnosis <- function(
year, 
file_path = get_boxi_extract_path(year = year, type = "GP_OoH-d")) 
{
```

R/read_extract_ooh_outcomes.R
```diff
@@ -1,13 +1,11 @@
 #' Read GP OOH Outcomes extract
 #'
-#' @param year Year of BOXI extract
+#' @inherit read_extract_acute
 #'
 #' @return a [tibble][tibble::tibble-package] with OOH Outcomes extract data
-read_extract_ooh_outcomes <- function(year) {
-  extract_outcomes_path <- get_boxi_extract_path(year = year, type = "GP_OoH-o")
-
+read_extract_ooh_outcomes <- function(year, file_path = get_boxi_extract_path(year = year, type = "GP_OoH-o")) {
```
Moohan, line 6:
```suggestion
read_extract_ooh_outcomes <- function(
year, 
file_path = get_boxi_extract_path(year = year, type = "GP_OoH-o"))
 {
```

Moohan (changes requested):
Some comments and suggestions. Looks good though.

I'll commit the styling ones just now.

Moohan:
Also check the documentation please (getting build warnings)

R/process_lookup_gpprac.R
Jennit07: TODO - split phs open data out into its own function

.Rbuildignore
```diff
@@ -23,3 +23,5 @@
 ^codecov\.yml$
 ^Rmarkdown$
 ^\.lintr$
+^_targets\.R$
+^_targets$
```
Moohan, line 27: I think only one of these is needed?

R/process_lookup_postcode.R
```diff
@@ -4,18 +4,25 @@
 #' postcode lookup, it will return the final data
 #' but also write this out as a zsav and rds.
 #'
+#' @param spd_path Path to spd lookup.
+#' @param simd_path Path to simd lookup.
```
Moohan, line 8:
```suggestion
#' @param spd_path Path to Scottish Postcode Directory.
#' @param simd_path Path to SIMD lookup.
```

.Rbuildignore
Jennit07: The _targets.R file sits outside of the folder. However i am not convinced i have done this right as the `_targets.R` file still shows in my console when i switch branch - i committed this to mainR so i would expect it to disappear?

R/get_gpprac_opendata.R
```diff
@@ -0,0 +1,47 @@
+#' GP Practice details from PHS open data
+#'
+#' @return The final data as a [tibble][tibble::tibble-package].
+#' @export
+#'
+get_gpprac_opendata <- function() {
```
Jennit07, line 6: I've pulled this code from `process_lookup_gpprac` so that i could set it up as a parameter along with spd etc. Hopefully this makes sense to have this set up in this way for targets

R/process_lookup_gpprac.R
Jennit07: I've split this out into its own function. Ready for re-review

.Rbuildignore
Moohan: [`.Rbuildignore`](https://github.com/Public-Health-Scotland/source-linkage-files/pull/544/files/84c4e5129c6989d01864ec9b7cda03e78b2fabc3..bfe04c38888aa041be0ebf9b3941a98e9cf621b2#diff-14fda507bd771db16c0fcb3a7d99002e54911cbe9f86bb2a6002562ee0fc0927) is for getting R CMD CHECK to ignore files but `.gitignore` for getting git to ignore files.

We want `_targets.R` to be tracked by git but not for R CMD to complain about it.

If you commit all the changes to `_targets.R` before you change branches, it should then just disappear. It's just that the uncommitted changes are transferring with you when you change branches.

For the `.Rbuildignore` above `^_targets\.R$` will find a file called (exactly) `_targets.R` whereas `^_targets$` will find a file called (exactly) `_targets` so I think you want the first line only?

R/process_lookup_gpprac.R
```diff
@@ -4,69 +4,38 @@
...
-    dplyr::mutate(
-      postcode = phsmethods::format_postcode(.data$postcode)
-    ) %>%
-    # keep distinct gpprac
-    dplyr::distinct(.data$gpprac, .keep_all = TRUE) %>%
-    # Sort for SPSS matching
-    dplyr::arrange(.data$gpprac) %>%
-    # Write rds file
-    write_rds(get_practice_details_path(check_mode = "write"))
+  opendata <- open_data
```
Moohan, line 24: I think just leave the parameter as `open_data` and use that wherever `opendata` was used previously. There's no need to have two very similarly named objects containing the same thing.

R/process_lookup_postcode.R
```diff
@@ -4,18 +4,25 @@
 #' postcode lookup, it will return the final data
 #' but also write this out as a zsav and rds.
 #'
+#' @param spd_path Path to Scottish Postcode Directory.
+#' @param simd_path Path to SIMD lookup.
+#' @param locality_path Path to locality lookup.
+#'
 #' @param write_to_disk (optional) Should the data be written to disk default is
 #' `TRUE` i.e. write the data to disk.
```
Moohan, line 12: Just a general comment but we can use `@inheritParams` to inherit very common param descriptions like `write_to_disk` 
```suggestion
#' @inheritParams <top_level_fun>
```

.Rbuildignore
Jennit07: Thanks @Moohan this is really useful to know the difference and the reason why the file keeps appearing! I will change this to find `_targets\.R` only as i had thought this would also need to pick up the targets folder which is why i also included `^_targets$`. But good to know that this is not needed in this instance and the `.Rbuildignore` is looking for specific files.

R/process_lookup_gpprac.R
```diff
@@ -4,69 +4,32 @@
...
-    # keep distinct gpprac
-    dplyr::distinct(.data$gpprac, .keep_all = TRUE) %>%
-    # Sort for SPSS matching
-    dplyr::arrange(.data$gpprac) %>%
-    # Write rds file
-    write_rds(get_practice_details_path(check_mode = "write"))
-
-
+process_lookup_gpprac <- function(open_data = get_gpprac_opendata(),
+                                  gpprac_ref_file = get_gpprac_ref_path(),
```
Moohan, line 18: Can we rename this `gpprac_ref_path` for accuracy and consistency please.

R/process_lookup_postcode.R
```diff
@@ -4,18 +4,24 @@
 #' postcode lookup, it will return the final data
 #' but also write this out as a zsav and rds.
 #'
-#' @param write_to_disk (optional) Should the data be written to disk default is
-#' `TRUE` i.e. write the data to disk.
+#' @param spd_path Path to Scottish Postcode Directory.
```
Moohan, line 7: This should already be inherited from `process_lookup_gpprac`
```suggestion
```

Moohan (approved):
Looks good to me, just one outstanding comment around renaming a parameter.

Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4134157057/attempts/1).

Moohan (approved):

PR #546, opened 2023-01-30
Update phs connection to use keyring

Jennit07:
I've split out the changes we have made to the phs connection to merge this back into mainR. I have updated some of the messages to be a bit more meaningful, however i am unsure how to include R code in the message? For example on lines 31-35 i wanted to include that chunk of code in the message `The keyring for createslf does not exist. Please set this up by updating your R profile with the following:` We can obviously come back to this further down the line as it is more important that the code is working and doing as we need it.


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4045397710/attempts/1).

Moohan:
Needs keyring adding with `use_package(package = "keyring", min_version = "1.3.0")`

R/get_connection_PHS_database.R
```diff
@@ -17,21 +17,53 @@ phs_db_connection <- function(dsn, username = Sys.getenv("USER")) {
...
+    cli::cli_abort("The keyring for createslf does not exist. Please set this up by updating your R profile")
+  }
+
+  if (!("db_password" %in% keyring::key_list(keyring = "createslf")[["service"]])) {
+    cli::cli_abort("The keyring password has not been supplied. Please set this up in your R profile")
+  }
+
+  if (Sys.getenv("CREATESLF_KEYRING_PASS") == "") {
+    cli::cli_abort("createslf keyring pw needs to be set in the environment variable
+                   {.env CREATESLF_KEYRING_PASS}")
```
Moohan, line 47:
```suggestion
                   {.envvar CREATESLF_KEYRING_PASS}")
```

R/get_connection_PHS_database.R
```diff
@@ -17,21 +17,53 @@ phs_db_connection <- function(dsn, username = Sys.getenv("USER")) {
...
+  }
+
+  keyring::keyring_unlock(keyring = "createslf", password = Sys.getenv("CREATESLF_KEYRING_PASS"))
+
+  if (keyring::keyring_is_locked(keyring = "createslf")) {
+    cli::cli_abort("Keyring is locked. To unlock createslf keyring, please use {.fun keyring_unlock}")
```
Moohan, line 53:
```suggestion
    cli::cli_abort("Keyring is locked. To unlock createslf keyring, please use {.fun keyring::keyring_unlock}")
```

Moohan:
> I've split out the changes we have made to the phs connection to merge this back into mainR. I have updated some of the messages to be a bit more meaningful, however I am unsure how to include the R code in the message. For example, on lines 31-35 I wanted to include that chunk of code in the message `The keyring for createslf does not exist. Please set this up by updating your R profile with the following:` We can obviously come back to this further down the line as it is more important that the code is working and doing as we need it.

I find this page super helpful - [https://cli.r-lib.org/reference/inline-markup.html](https://cli.r-lib.org/reference/inline-markup.html) but yep you're right that writing great error messages can take a while to do well, so we can leave it as a TODO for now.

R/get_connection_PHS_database.R
```diff
@@ -17,21 +17,53 @@ phs_db_connection <- function(dsn, username = Sys.getenv("USER")) {
 
   # Check the username is not empty and take input if not
   if (is.na(username) | username == "") {
-    username <- rstudioapi::showPrompt(
-      title = "Username",
-      message = "Username",
-      default = ""
-    )
+    if (interactive()) {
```
Moohan, line 20: Using `rlang::is_interactive()` works the same in normal use, but will let us test things easier e.g. with 

```R
rlang::with_interactive({phs_db_connection("DVPROD")}, TRUE)
rlang::with_interactive({phs_db_connection("DVPROD")}, FALSE)
```

```suggestion
    if (rlang::is_interactive()) {
```

R/get_connection_PHS_database.R
```diff
@@ -17,21 +17,53 @@ phs_db_connection <- function(dsn, username = Sys.getenv("USER")) {
...
   }
 
+  # TODO improve error messages and provide instructions for setting up keyring
+  # Add the following code to R profile.
+  # Sys.setenv("CREATESLF_KEYRING_PASS" = "createslf"),
+  # keyring_create("createslf", password = Sys.getenv("CREATESLF_KEYRING_PASS")),
+  # key_set(keyring = "createslf", service = "db_password")
+
+  if (!("createslf" %in% keyring::keyring_list()[["keyring"]])) {
+    cli::cli_abort("The keyring for createslf does not exist. Please set this up by updating your R profile")
```
Moohan, line 38:
```suggestion
    cli::cli_abort("The {.val createslf} keyring does not exist.")
```

R/get_connection_PHS_database.R
```diff
@@ -17,21 +17,53 @@ phs_db_connection <- function(dsn, username = Sys.getenv("USER")) {
...
+  }
+
+  if (!("db_password" %in% keyring::key_list(keyring = "createslf")[["service"]])) {
+    cli::cli_abort("The keyring password has not been supplied. Please set this up in your R profile")
+  }
+
+  if (Sys.getenv("CREATESLF_KEYRING_PASS") == "") {
+    cli::cli_abort("createslf keyring pw needs to be set in the environment variable
+                   {.env CREATESLF_KEYRING_PASS}")
+  }
```
Moohan, line 48:
```suggestion
  if (Sys.getenv("CREATESLF_KEYRING_PASS") == "") {
    cli::cli_abort("You must have the password to unlock the {.val createslf} keyring in your environment as
                   {.env CREATESLF_KEYRING_PASS}. Please set this up in your {.file .Renviron} or {.file .Rprofile}")
  }
```

Moohan (changes requested):
I think those changes are fine for now. I'll just commit them and then test and approve.

We can come back to this in the future, I think it will look something like:

```R
if (rlang::is_interactive()) {

<Check for keyring>
# if anything is wrong, guide the user to interactively create the keyring, and add their password to it. The first thing they will need to do this will be the 'createslf' shared password in their env so this would be the first check.

} else {

<Do checks and errors as we're currently doing>

# In this scenario we can't fix anything, so we just have to give good error messages.

# Create the connection
  password_text <- glue::glue("{dsn} password for user: {username}")
  db_connection <- odbc::dbConnect(
    odbc::odbc(),
    dsn = dsn,
    uid = username,
    pwd = rstudioapi::askForPassword(password_text)
    pwd = keyring::key_get(keyring = "createslf", service = "db_password")
  )

  keyring::keyring_lock(keyring = "createslf")

  return(db_connection)
}
```

Moohan (approved):

PR #547, opened 2023-01-30
Update `find_latest_file()`

Moohan:
Some changes to make the messages from `find_latest_file` a bit cleaner and shorter - if it only finds a single file matching the regex it will now give a much shorter message.


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4047161771/attempts/1).

lizihao-anu (approved):

PR #556, opened 2023-02-09
Fix fill geogs

Moohan:
Not sure if this fixes the bug you were noticing but hopefully it fixes something!


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4135311982/attempts/1).

shintoLamp:
Made a change, the way I've done it works but it's not as nice looking unfortunately

Moohan:
> Made a change, the way I've done it works but it's not as nice looking unfortunately

Haha, did my way not work? It was relying on using `{dplyr} >= 1.1.0` as `case_when` and `if_else` no longer need to use type-specific `NA`, so I was thinking just using `NA` would mean we got a `NA_real_` for `gpprac` and an `NA_character_` for `postcode`.

Happy with your way though, if mine wasn't working!

PR #557, opened 2023-02-13
Add @export to missed documentation

Jennit07:
Quick PR - Add @export to functions which are needed for targets. I have looked through all the ones used by targets and exported those missed. On this branch i have also updated the costs parameters on `process_costs_rmd.R` as there was some redundant code on the test-production branch which made the change on old code. I have updated this here which should be merged onto mainR


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4165081374/attempts/1).

Moohan (approved):
Looks good and works fine for me.

At some point, we should do the opposite of this and 'unexport' functions which are just internal and not expected to be used other than by other package functions.

PR #560, opened 2023-02-15
Update social care parameters

Jennit07:
This PR deals with updating the social care function parameters for use with targets. This has been split out from the test-production branch and works within targets. Happy to make any further changes if i have missed anything.


R/process_extract_alarms_telecare.R
```diff
@@ -4,33 +4,32 @@
...
 #' @family process extracts
-process_extract_alarms_telecare <- function(data = NULL, year, client_lookup = NULL, write_to_disk = TRUE) {
-  # Include is.null for passing the processed ALL alarms Telecare data through a list
-  if (is.null(data)) {
-    data <- readr::read_rds(get_sc_at_episodes_path())
-  }
-
+process_extract_alarms_telecare <- function(data = get_sc_at_episodes_path(),
+                                            year,
+                                            client_lookup = get_source_extract_path(year, type = "Client"),
```
Moohan, line 18: I don't like having parameters called data which are expecting a file path. Can we rename them please?
```suggestion
process_extract_alarms_telecare <- function(file_path = get_sc_at_episodes_path(),
                                            year,
                                            client_lookup_path = get_source_extract_path(year, type = "Client"),
```

Moohan (changes requested):
Just the comments on parameter names.

I remember we discussed having a `data` and a `file_path` parameter and using `data` unless it was `NULL` in which case reading the data from disk with `file_path`... I know we also talked about wanting to track files in targets so maybe this was the reason?

Jennit07:
> Just the comments on parameter names.
> 
> I remember we discussed having a `data` and a `file_path` parameter and using `data` unless it was `NULL` in which case reading the data from disk with `file_path`... I know we also talked about wanting to track files in targets so maybe this was the reason?

I think i left this as one parameter rather than splitting data and file path due to this not working in targets for now. I could only get it to work when this was read from disk using the file_path. I've updated this to rename the parameters. Maybe we could come back to this to check if that was working again when we revisit targets?

PR #562, opened 2023-02-23
Tweaks to SC Client lookup

Moohan:
* Data is now extracted as integers (instead of characters) this makes the extract much smaller, and a bit faster too.
 * There's some code which handles the 'unknown' values better - this matches the SPSS but only affects a small number of records (if any). It is required due to bad coding in the data where for example "6" should be unknown but sometimes the value is completely missing.


Moohan:
> @check-spelling-bot apply updates.

lizihao-anu (approved):

PR #564, opened 2023-03-01
Update test functions

Jennit07:
Updates to `produce_test_comparison` and `calculate_measures` to include flexibility for episode and individual file tests. This should be merged before the episode test functions in a separate PR #565

### DIFF
--- a/R/calculate_measures.R
+++ b/R/calculate_measures.R
@@ -12,6 +12,9 @@
 #'
 #' @param measure The measure you want to apply to variables
 #'
+#' @param group_by Default as NULL for grouping variables. Specify
+#' variables for grouping e.g recid for episode file testing.
+#'
 #' @return a tibble with a summary
 #' @export
 #'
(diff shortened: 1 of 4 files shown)


R/calculate_measures.R
```diff
@@ -64,12 +70,22 @@ calculate_measures <- function(data,
...
   }
 
-  pivot_data <- data %>%
-    tidyr::pivot_longer(
-      cols = tidyselect::everything(),
-      names_to = "measure",
-      values_to = "value"
-    )
+
+  if (group_by) {
```
Moohan, line 74: What if we make `group_by` `NULL` by default, and then give it `"recid"` for this case?
```suggestion
  if (group_by == "recid") {
```

This will make it slightly more flexible if we wanted to change things in the future.

Happy either way though.

R/calculate_measures.R
```diff
@@ -19,9 +19,15 @@
 #' @seealso produce_source_extract_tests
 calculate_measures <- function(data,
                                vars = NULL,
-                               measure = c("sum", "all", "min-max")) {
+                               measure = c("sum", "all", "min-max"),
+                               group_by = FALSE) {
```
Moohan, line 23: This param needs adding to the function documentation.

Moohan (approved):
Just a minor bit of documentation, and a comment on how we use the param in the future!

PR #565, opened 2023-03-02
Episode file tests

Jennit07:
PR #564 must be merged into MainR before this one otherwise there will be conflicts. 

This PR deals with creating episode tests for testing the newly created episode file in R by @shintoLamp. Help is needed to look into why the percentage change is not quite right, even though this uses existing code and follows the same formats as other 'extract tests'. Hopefully this will give us an idea of how close the episode creation is to the existing SPSS files. This can also be applied to the individual file tests which i will also develop in a separate PR


R/produce_episode_file_tests.R
```diff
@@ -0,0 +1,68 @@
...
+#'
+#' @param data new or old data for testing summary flags
+#' (data is from [get_source_extract_path()])
+#' @param sum_mean_vars variables used when selecting 'all' measures from [calculate_measures()]
+#' @param max_min_vars variables used when selecting 'min-max' from [calculate_measures()]
+#' @inheritParams calculate_measures
+#'
+#' @return a dataframe with a count of each flag
+#' from [calculate_measures()]
+#' @export
```
Moohan, line 20:
```suggestion
```

Moohan (approved):

PR #567, opened 2023-03-13
Refactor test scripts

Moohan:
I've refactored the tests so that they use far fewer individual scripts. The lower-level functions are no longer exported.


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4407392276/attempts/1).

PR #570, opened 2023-03-15
fix missing chi after adding nsu

lizihao-anu:


Moohan (approved):
I found another couple of bugs when I was checking this but it looks good to me!

PR #571, opened 2023-03-15
fix the bug of dropping missing chi episodes

lizihao-anu:
fix the bug of dropping missing chi epsidoes in correct_demographics.R


R/correct_demographics.R
```diff
@@ -8,23 +8,35 @@
...
       chi_dob_max = phsmethods::dob_from_chi(
         chi_number = .data$chi,
         chi_check = FALSE,
         min_date = lubridate::ymd("2000-01-01"),
-        max_date = pmax(.data$keydate1_dateformat, lubridate::ymd("2000-01-01"), na.rm = TRUE)
+        max_date = pmax(
+          .data$keydate1_dateformat,
+          lubridate::ymd("2099-12-31"),
+          na.rm = TRUE
+        )
```
Moohan, line 39: I don't think this is correct. Here we're looking for a date of birth in the 20thC. The latest that date could be is the end of the year, or the date of the episode (since a person can't be born after an episode has started). So should it be:
```suggestion
        max_date = pmin(
          .data$keydate1_dateformat,
          lubridate::ymd("2099-12-31"),
          na.rm = TRUE
        )
```

Not sure ... 🤯 I'm off home, but I/We maybe need to draw this one out to nail the logic!

PR #583, opened 2023-03-21
Add functions to automatically prepare the environment

Moohan:
This is ready for review, both functions work and I've added them to targets.


lizihao-anu (approved):
Looks good

PR #585, opened 2023-03-22
Fix bug in episode file tests

Jennit07:
Applying a fix to the episode file tests which was preventing this from running. The main issue was that dplyr was not supplied in a `group_by` and this was also `TRUE` which should have been `"recid"`. I have also updated the `run_episode_file` documentation here as this was failing.


shintoLamp (approved):
Checked over the code, also spoke to Jen and she produced tests using this new code so we know it works. Happy for it to be merged.

PR #588, opened 2023-03-23
Delete redundant scripts

Moohan:
These scripts are now redundant as they have been included elsewhere.


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4503027836/attempts/1).

Jennit07 (approved):
Agree, these should all be removed

PR #597, opened 2023-03-27
Remove the dependency on `glue`

Moohan:
We were only using `glue::glue` which is re-implemented by `stringr::str_glue` with a few of the more niche bits removed.

This removes a dependency for us which means one fewer package to keep track of updates for!


Jennit07 (approved):
Added this to the new homelessness completeness function. Ready to approve/merge

PR #598, opened 2023-03-28
fix calculate_measures.R

lizihao-anu:


Jennit07:
Small change to `calculate_measures` function. There was a bug which was preventing the extract tests to run, but episode tests were fine. This is now fixed by Zihao and we talked through this. Happy to approve/merge.

Jennit07 (approved):
ready to merge

PR #599, opened 2023-03-28
Remove `{readxl}` dependency

Moohan:
In my testing, `{openxl}` is faster (they're both quite fast), which is nice. The features are not quite equivalent though so some things have to be done slightly differently.


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4730213045/attempts/1).

Jennit07 (approved):
Looks good. I did a find in files search to see if there were more instances but i think we have caught them all here. Will approve/merge

PR #600, opened 2023-03-29
Fix costs 20230328

lizihao-anu:
The current create_monthly_costs function for maternity costs overlooks the daycases where the keydate1 equals keydate2. The daycases come with the yearstays being 0.33. The patch reduces the gap of monthly costs to less than 1% to each month, not a perfect match though.

As for the patch, I could not think of a more elegantly vectorized way. This patch runs much faster than for loop though the logic is not straightforward. So feel free to make any comments and happy to discuss.


R/create_monthly_costs.R
```diff
@@ -19,16 +20,57 @@ create_monthly_costs <- function(data,
...
+      .data$daycase_added,
+      lubridate::month(data$record_keydate1),
+      NA
+    )) %>%
+    dplyr::mutate(
+      whichmonth = month.abb[.data$whichmonth] %>%
+        tolower() %>%
+        paste0("_cost"),
+      daycase_added = dplyr::if_else(.data$daycase_added, 1, 0)
+    ) %>%
```
Moohan, line 42: Can trim this code a bit by getting lubridate to give the abbreviated date label
```suggestion
    dplyr::mutate(whichmonth = dplyr::if_else(
      .data$daycase_added,
      lubridate::month(data$record_keydate1, label = TRUE) %>%
        tolower() %>%
        paste0("_cost"),
      NA
    )) %>%
    dplyr::mutate(
      daycase_added = dplyr::if_else(.data$daycase_added, 1, 0)
    ) %>%
```

R/create_monthly_costs.R
```diff
@@ -19,16 +20,57 @@ create_monthly_costs <- function(data,
...
+    tidyr::pivot_wider(
+      names_from = .data$whichmonth,
+      values_from = .data$daycase_added,
+      values_fill = 0
+    ) %>%
```
Moohan, line 47: We can do the rename here which tidies up the above code.
```suggestion
    tidyr::pivot_wider(
      names_from = "whichmonth",
      names_glue = "{whichmonth}_cost",
      values_from = "daycase_added",
      values_fill = 0
    ) %>%
```

R/create_monthly_costs.R
Moohan, line 42:
```suggestion
    dplyr::mutate(
      daycase_check = (.data$record_keydate1 == .data$record_keydate2)
    ) %>%
    dplyr::mutate(daycase_added = dplyr::if_else(.data$daycase_check, 1, 0)) %>%
    dplyr::mutate(whichmonth = dplyr::if_else(
      .data$daycase_check,
      lubridate::month(data$record_keydate1),
      NA
    )) %>%
    dplyr::mutate(
      whichmonth = month.abb[.data$whichmonth] %>%
        tolower() %>%
        paste0("_cost")
    ) %>%
```

Moohan (changes requested):
Great work finding the bug. I've made a few overlapping suggestions, so hopefully, you can unravel them!

R/create_monthly_costs.R
```diff
@@ -19,16 +20,57 @@ create_monthly_costs <- function(data,
...
+      daycase_check
+    )
+
+  costs <- (costs_daycase[1:12] + costs) %>%
+    dplyr::bind_cols(daycase_check = costs_daycase$daycase_check)
 
   data <- dplyr::bind_cols(data, costs) %>%
+    dplyr::mutate(yearstay2 = dplyr::if_else(
+      {{ yearstay }} < 0.99, 0, {{ yearstay }}
+    )) %>%
```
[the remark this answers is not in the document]
lizihao-anu: The cost_daycases bit of code (Line 23-56 currently) decides which episodes are day cases and gives the indication later of which month the costs happen in. So the cost_daycases bit has to be kept.

But the yearstay2 bit (line 59-61) is not really necessary and feel free to delete this bit. I keep this bit of code only to allow for any possible changes in the yearstay as I am not sure how the raw data would change in the future. To be specific, the yearstays for the day cases are only 0.33, but I am (unnecessarily) worried that those could be changed in the future to 0 or some number between 0 and 1. Hence, I think it is completely ok to delete this yearstay2 bit.

R/create_monthly_costs.R
```diff
@@ -19,16 +20,57 @@ create_monthly_costs <- function(data,
...
-        .x != 0.0,
-        .x / {{ yearstay }} * {{ cost_total_net }},
-        0.0
+        .x != 0.0 & !daycase_check,
+        .x / yearstay2 * {{ cost_total_net }},
+        dplyr::if_else(.x != 0.0 & daycase_check,
+          cost_total_net,
+          0
+        )
       )
```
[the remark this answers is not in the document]
lizihao-anu: Yes! Why didn't I think of `case_when`

R/create_monthly_costs.R
lizihao-anu: Hi James. Sorry I have to revert this change. Actually, the variable `whichmonth` does not contain all 12 months. Hence, it will be short of some columns, such as `apr_cost`, if we use `name_glue`. However, the old bit of the pivot_wider code strangely creates all columns for 12 months. I guess it is because `month.abb[]` in fact creates a vector with levels being 12 months, and all levels will be created as columns.

Feel free to discuss the unelegant code @_@.

lizihao-anu:
Should be ok for a review again.

lizihao-anu:
In `create_monthly_beddays`, if we set the variable count_last to FALSE, then the R result is a perfect match with the SPSS result. @Moohan

Moohan (approved):

PR #601, opened 2023-03-29
Fix dn tests

lizihao-anu:
The gap of mean monthly costs between SPSS and R results of DN is because of the SPSS results contains NA for 0 in monthly costs. Hence, replacing NA with 0 in the testing script can fix the gap.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -67,10 +67,9 @@ pattype
 phs
 phsmethods
 phsopendata
-postcodes
 placeinc
+postcodes
 PPAs
-praccode
 prac
 praccode
 purrr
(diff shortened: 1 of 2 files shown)


Moohan (commented):
Seems sensible to me, though we should add a `#TODO` to come back to it when we've stopped using SPSS.

I'm wondering why there were any `NA` for costs in SPSS?

lizihao-anu:
> 

In B07, Line 131-134, the code only gives values (>) to monthly costs while giving no values if cost = 0. It leaves NA to those monthly costs without a positive value.

Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4554636876/attempts/1).

PR #602, opened 2023-03-29
replace NA with 0 when reading the extract of acute and mental health

lizihao-anu:
The issue with acute test results is the mean total costs and monthly mean costs go slightly up. It is because Function `read_extract_acute` will give data with NA in `total_net_cost`. Hence, replacing NA with 0 can fix the issue.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -1,4 +1,6 @@
+admloc
 admtype
+adtf
 aut
 bedday
 beddays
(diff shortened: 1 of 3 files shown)


shintoLamp:
No problem that I can see, I would consider using this branch to also ensure NA values are converted to zeroes in our other extracts too

Moohan (approved):

Moohan:
> No problem that I can see, I would consider using this branch to also ensure NA values are converted to zeroes in our other extracts too

Hopefully, it's just these two, probably due to the way we take the extracts from BOXI meaning we're processing the costs slightly differently!

Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4554557367/attempts/1).

PR #607, opened 2023-03-30
Delete old Social Care scripts

Moohan:
These scripts are no longer required as they've been full incorporated into the R package.

The `00-Social_Care_functions.R` script would occasionally conflict with package functions too, which is annoying.


Jennit07 (approved):
Agree, these should now be deleted. Will approve and merge

PR #615, opened 2023-04-06
Fix a Home Care bug

Moohan:
Fix bug in Home Care where reablement was being left with NA (should be 9).

In the data `reablement` has both missing (NA) and 9. Unknown should be coded as 9. We first set all the 9s to NA, then use `tidyr::fill` to fill in from other records. The bit that was missed was replacing the remaining NAs with 9.


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4630780774/attempts/1).

Jennit07 (approved):
Looks good

PR #616, opened 2023-04-06
Fix a bug in the SC Demographics

Moohan:
This has probably been outstanding for some time.

`phsmethods::format_postcode` forces them to uppercase, so all the postcodes will definitely be uppercase. Previously the regex didn't match anything as it was only looking for lowercase (it probably should have been case insensitive.)

The regex now looks exclusively for uppercase so we should continue keeping valid postcodes from the SC data!


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4630804703/attempts/1).

Jennit07 (approved):
Looks good!

PR #618, opened 2023-04-06
Use integers for appropriate variables

Moohan:
This fixes a bug where the client file was storing `sending_location` as an integer and couldn't be matched against the character type in the other files. Other variables will take less storage as an integer, likewise, a couple of variables which aren't actually needed are now dropped earlier.

### DIFF
--- a/R/read_sc_all_care_home.R
+++ b/R/read_sc_all_care_home.R
@@ -7,9 +7,6 @@
 #' @export
 #'
 read_sc_all_care_home <- function(sc_dvprod_connection = phs_db_connection(dsn = "DVPROD")) {
-  # Read in data---------------------------------------
-
-  ## read in data - social care 2 demographic
   ch_data <- dplyr::tbl(
     sc_dvprod_connection,
     dbplyr::in_schema("social_care_2", "carehome_snapshot")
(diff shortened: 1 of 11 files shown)


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4630957316/attempts/1).

Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4630973290/attempts/1).

R/read_sc_all_home_care.R
```diff
@@ -31,15 +31,28 @@ read_sc_all_home_care <- function(sc_dvprod_connection = phs_db_connection(dsn =
     # fix 2017
     dplyr::mutate(
       financial_quarter =
-        dplyr::if_else(.data$financial_year == 2017L & is.na(.data$financial_quarter),
+        dplyr::if_else(
+          .data$financial_year == 2017L & is.na(.data$financial_quarter),
           4,
```
Jennit07, line 36:
```suggestion
          4L,
```
should financial quarter be an integer here?

Jennit07 (approved):
One query about financial quarter. Other than that, happy to approve this/merge when ready.

R/read_sc_all_home_care.R
Moohan: Just looking at the code I think we can avoid reading in `financial_quarter` completely, as we correct it here but it's never used for anything, so we may as well just not read it in the first place!

R/read_sc_all_care_home.R
```diff
@@ -43,7 +40,17 @@ read_sc_all_care_home <- function(sc_dvprod_connection = phs_db_connection(dsn =
         .data$period
       )
     ) %>%
-    dplyr::collect()
+    dplyr::select(!c("financial_year", "financial_quarter")) %>%
```
Moohan, line 43: Similar here as in home_care - We don't actually use `financial_year` or `financial_quarter` for anything so it's pointless reading them in and doing these corrections!

Jennit07: We spoke about this and these variables were needed for correcting period where this was in the format `2017` instead of `2017Q4`. This is now simplified to do this without using `financial_year` or `financial_quarter` which can be removed from the read in. I've pushed the commit and will now merge.

PR #627, opened 2023-04-18
Fix some variable names in acute

Moohan:
These variables should be lower-case but the casing was apparently copied over from SPSS. The code later for the episode file expects `hscp` and `datazone` so the data from these was just being ignored.


Jennit07 (approved):
Good spot! Agree these should now be lowercase

PR #629, opened 2023-04-18
Fix the case of the keytimex variables in CMH

Moohan:
This casing is different to some other datasets so wasn't joining properly at the episode file stage.


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4734110588/attempts/1).

PR #631, opened 2023-04-18
Fix a misnamed variable in `correct_demographics`

Moohan:
We were testing with the old files which had `keydate*` as the date variable but in the new file we only have `record_keydate*` which is of a date type.


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4734392107/attempts/1).

Jennit07 (approved):
Looks good, ready to merge

PR #633, opened 2023-04-19
Reduce memory usage for the fill geographies functions

shintoLamp:
So for this I have created two new functions that create lookups, one for postcode and one for GP practice. I've left the cascade geographies functions as they are because there's no aggregates in those.

Running the fill_geographies function takes much less time now and significantly less memory for a full episode file, and I'm pretty sure the logic is solid on how `dplyr::distinct` will choose the first instance in a properly ordered dataset, but I would appreciate if someone could run it/check it too.


R/fill_geographies.R
```diff
@@ -27,21 +27,70 @@ fill_geographies <- function(data) {
...
+#' @param data A data frame containing chi, postcode, and a date of last contact
+#'
+#' @return A lookup containing CHI numbers and their most recent postcodes
+make_postcode_lookup <- function(data) {
+  postcode_lookup <-
+    data %>%
+    # Get just the chi, postcode and discharge date
+    dplyr::select("chi", "postcode", "record_keydate2") %>%
+    # Get rid of missing chis
+    dplyr::filter(!is_missing(.data$chi)) %>%
```
Moohan, line 41: I think if we do a `distinct` right at the top, it will make it even quicker as there will be fewer rows for the `mutate` to work on.
```suggestion
    # Get rid of records with missing CHIs
    dplyr::filter(!is_missing(.data$chi)) %>%
    # Keep one row for each CHI, postcode, and date combination.
    dplyr::distinct("chi", "postcode", "record_keydate2") %>%
```

R/fill_geographies.R
```diff
@@ -27,21 +27,70 @@ fill_geographies <- function(data) {
...
+    # Format postcodes to 7-character format and replace dummy with NA
+    dplyr::mutate(
+      postcode = phsmethods::format_postcode(postcode, format = "pc7"),
+      postcode = dplyr::na_if(.data$postcode, "NK010AA")
+    ) %>%
+    # Arrange with most recent date at the top for each CHI
+    dplyr::arrange(.data$chi, desc(.data$record_keydate2)) %>%
```
Moohan, line 48:
```suggestion
    # Arrange with the most recent date at the top for each CHI
    dplyr::arrange(.data$chi, dplyr::desc(.data$record_keydate2)) %>%
```

R/fill_geographies.R
```diff
@@ -27,21 +27,70 @@ fill_geographies <- function(data) {
...
+fill_postcode_geogs <- function(data) {
+  spd <- read_file(get_slf_postcode_path())
 
-  filled_geographies <- data_geographies %>%
+  filled_postcodes <-
+    dplyr::left_join(data, make_postcode_lookup(data), by = "chi") %>%
+    dplyr::select(-postcode) %>%
+    dplyr::rename(postcode = most_recent_postcode) %>%
+    # Fill geographies
+    dplyr::left_join(., spd, by = "postcode", suffix = c("_old", "")) %>%
```
Moohan, line 93:
```suggestion
    dplyr::left_join(spd, by = "postcode", suffix = c("_old", "")) %>%
```

R/fill_geographies.R
```diff
@@ -58,47 +107,24 @@ fill_postcode_geogs <- function(data) {
...
-    by = "gpprac",
-    suffix = c("_old", "")
-  )
-
-  filled_geographies <- data_geographies %>%
+  filled_gpprac <-
+    dplyr::left_join(data, make_gpprac_lookup(data), by = "chi") %>%
+    dplyr::select(-gpprac) %>%
+    dplyr::rename(gpprac = most_recent_gpprac) %>%
+    dplyr::left_join(., gpprac_ref, by = "gpprac", suffix = c("_old", "")) %>%
```
Moohan, line 121:
```suggestion
    dplyr::left_join(gpprac_ref, by = "gpprac", suffix = c("_old", "")) %>%
```

R/fill_geographies.R
```diff
@@ -27,21 +27,70 @@ fill_geographies <- function(data) {
...
+      postcode = dplyr::na_if(.data$postcode, "NK010AA")
+    ) %>%
+    # Arrange with most recent date at the top for each CHI
+    dplyr::arrange(.data$chi, desc(.data$record_keydate2)) %>%
+    # Take only the first of each chi
+    dplyr::distinct(.data$chi, .keep_all = TRUE) %>%
+    # Rename
+    dplyr::rename(most_recent_postcode = postcode) %>%
+    # We don't need to keep discharge date after this point
+    dplyr::select(-"record_keydate2")
```
Moohan, line 54:
```suggestion
    dplyr::select(-"record_keydate2") %>%
    # Drop any CHIs with no postcode
    tidyr::drop_na()
```

R/fill_geographies.R
```diff
@@ -27,21 +27,70 @@ fill_geographies <- function(data) {
...
+    # Replace dummy with NA
+    dplyr::mutate(gpprac = dplyr::na_if(.data$gpprac, 99999L)) %>%
+    # Arrange with most recent date at the top for each CHI
+    dplyr::arrange(.data$chi, desc(.data$record_keydate2)) %>%
+    # Take only the first of each chi
+    dplyr::distinct(.data$chi, .keep_all = TRUE) %>%
+    # Rename
+    dplyr::rename(most_recent_gpprac = gpprac) %>%
+    # We don't need to keep discharge date after this point
+    dplyr::select(-"record_keydate2")
```
Moohan, line 80:
```suggestion
    dplyr::select(-"record_keydate2") %>%
    # Drop any CHIs with no gpprac code
    tidyr::drop_na()
```

Moohan (changes requested):
Looks good and great that it's faster and uses less memory!

I think it's missing a bit where we'd check against the lookup? In the postcode one, for example, any postcodes which don't match the regex would be made NA by `format_postcode` but we can also check against the SPD - If you take just the postcode column from the SPD and do an inner join you'll be left with only valid Scottish postcodes in the lookup.

I also think we need to be slightly more careful about re-applying them. We only want to overwrite an existing postcode with a new valid one but I think your code would also overwrite with an NA.

The above also applies to gpprac!

Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4745164479/attempts/1).

Jennit07:
I have tested this using my temp 1920 file and ive ran into an error with the function `recode_hscp` did anyone else find this? maybe it is caused by the data i was testing but the temp file should be down to before `fill_geographies` has been used

Error in `dplyr::mutate()`:
ℹ In argument: `hscp2018 = recode_hscp(.data$hscp2018)`.
Caused by error in `dplyr::case_when()` at Source_Linkage_Files/R/fill_geographies.R:332:2:
! Failed to evaluate the left-hand side of formula 1.
Caused by error in `vec_equal()`:
! Can't combine `..1` <haven_labelled> and `..2` <character>.

R/fill_geographies.R
```diff
@@ -27,21 +27,70 @@ fill_geographies <- function(data) {
...
-    suffix = c("_old", "")
-  )
+fill_postcode_geogs <- function(data) {
+  spd <- read_file(get_slf_postcode_path())
 
-  filled_geographies <- data_geographies %>%
+  filled_postcodes <-
+    dplyr::left_join(data, make_postcode_lookup(data), by = "chi") %>%
+    dplyr::select(-postcode) %>%
+    dplyr::rename(postcode = most_recent_postcode) %>%
```
Moohan, line 91:
```suggestion
    # If the existing postcode is missing or not known (NK010AA) replace it.
    dplyr::mutate(postcode = dplyr::if_else(is.na(.data$postcode) || .data$postcode == "NK010AA", .data$most_recent_postcode, .data$postcode)) %>%
```

R/fill_geographies.R
```diff
@@ -58,47 +107,24 @@ fill_postcode_geogs <- function(data) {
...
-    gpprac_ref,
-    by = "gpprac",
-    suffix = c("_old", "")
-  )
-
-  filled_geographies <- data_geographies %>%
+  filled_gpprac <-
+    dplyr::left_join(data, make_gpprac_lookup(data), by = "chi") %>%
+    dplyr::select(-gpprac) %>%
+    dplyr::rename(gpprac = most_recent_gpprac) %>%
```
Moohan, line 120:
```suggestion
    # If the existing gpprac is missing or not known (99999) replace it.
    dplyr::mutate(gpprac = dplyr::if_else(is.na(.data$gpprac) || .data$gpprac == 99999L, .data$most_recent_gpprac, .data$gpprac)) %>%
```

Moohan (changes requested):
Just adding suggestions for corrections I mentioned in my previous review.

Moohan (approved):

PR #635, opened 2023-04-20
NI episode file

Moohan:
This works so far, i.e. it produces a file and uses max ~70GB memory. It would be good to try running NI 12, 13 and 20 using it. It will also likely require some changes once we've merged the other episode file related PRs but as an idea it's solid!

### DIFF
--- a/R/run_episode_file.R
+++ b/R/run_episode_file.R
@@ -8,9 +8,18 @@
 #' @return a [tibble][tibble::tibble-package] containing the episode file
 #' @export
 #'
-run_episode_file <- function(processed_data_list, year, write_to_disk = TRUE) {
+run_episode_file <- function(processed_data_list, year, write_to_disk = TRUE, ni_only = FALSE) {
   # Bring all the datasets together from Jen's process functions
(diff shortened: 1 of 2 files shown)


R/run_episode_file.R
```diff
@@ -8,8 +8,18 @@
...
   episode_file <- dplyr::bind_rows(processed_data_list) %>%
+    {
+      if (ni_only) {
+        # Drop datasets not needed for NIs ()
+        dplyr::filter(., .data$recid %in% c("01B", "02B", "04B", "GLS", "00B", "PIS")) %>%
+          dplyr::select(dplyr::where(~ any(!is.na(.x))))
+      } else {
+        .
+      }
+    } %>%
```
Moohan, line 22: Needs to be re-tested to see how this code interacts with the `store_ep_file_vars` code.

Moohan:
Two build issues

- [ ] Document `ni_only` param
>   Undefined global functions or variables:
    .

- [ ] My usage of `.` and `{}` to pass through the data based on the `ni_only` param is producing a note because it doesn't recognise `.` - Anyone have any ideas?

PR #641, opened 2023-05-01
Resolve issue with years having no data

Moohan:
We were previously passing NA for the file path and NULL for the data when the year/extract had no data (e.g. District Nursing 2223). Now we point to a dummy (empty) file for the file path, then in the processing recognise this and just return an empty tibble. This should work well with targets and with the episode file `bind_rows`.


R/process_extract_alarms_telecare.R
```diff
@@ -20,6 +20,12 @@ process_extract_alarms_telecare <- function(
   # Check that the supplied year is in the correct format
   year <- check_year_format(year)
 
+  # Check that we have data for this year
+  if (!check_year_valid(year, "CH")) {
```
Jennit07, line 24:
```suggestion
  if (!check_year_valid(year, "AT")) {
```
Do you mean AT in this script?

R/process_extract_home_care.R
```diff
@@ -20,6 +20,12 @@ process_extract_home_care <- function(
   # Check that the supplied year is in the correct format
   year <- check_year_format(year)
 
+  # Check that we have data for this year
+  if (!check_year_valid(year, "CH")) {
```
Jennit07, line 24:
```suggestion
  if (!check_year_valid(year, "HC")) {
```
HC instead?

R/process_extract_sds.R
```diff
@@ -20,6 +20,12 @@ process_extract_sds <- function(
   # Check that the supplied year is in the correct format
   year <- check_year_format(year)
 
+  # Check that we have data for this year
+  if (!check_year_valid(year, "CH")) {
```
Jennit07, line 24:
```suggestion
  if (!check_year_valid(year, "SDS")) {
```

Jennit07 (approved):
Looks good and working as expected. Nicely set up to create a dummy file for extracts which have no data for certain years. Ready to merge

Issue #581, opened 2023-03-17
Reduce the number of exports


Moohan:
We are currently exporting a lot of functions, a lot of these are quite 'low-level', where we would only ever expect them to be called by another function in the package, these should not be exported.

shintoLamp:
Looks like James has worked on this and merged the relevant code, will close the issue

PR #643, opened 2023-05-01
Unexport functions

Moohan:
These functions are considered lower-level and are not intended to be called separately from other functions. Therefore, there is no need to export them.


Moohan:
Closes #581

lizihao-anu (approved):

PR #644, opened 2023-05-02
Resolve #642 error in creating Acute smrtype

### LINKED ISSUE #642: Investigate targets acute 22/23 error
Investigate error from targets where NA values are present when trying to add smrtype to acute. This applies to 2022/23 only as other extracts ran successfully
```
✖ error target acute_source_extract_2223
Error: ℹ In argument: `smrtype = add_smr_type(recid = .data$recid, ipdc =
  .data$ipdc)`.
Caused by error in `add_smr_type()`:
! In Acute records, `ipdc` is required to assign an smrtype, and there
  are some "NA" values. Please check the data.
```

Jennit07:
In 2022/23 file there was one record where idpc was `NA` which was causing an error when trying to add the smrtype for acute. I have now added this as `Acute-Unknown` and also changed the error messages to warnings instead. Tested this on `process_extract_acute` and happy this is working.

### DIFF
--- a/R/add_smr_type.R
+++ b/R/add_smr_type.R
@@ -138,7 +138,8 @@ add_smr_type <- function(recid,
       recid == "01B" & ipdc == "I" ~ "Acute-IP",
       recid == "01B" & ipdc == "D" ~ "Acute-DC",
       recid == "GLS" & ipdc == "I" ~ "GLS-IP",
-      TRUE ~ "Acute-Unknown"
+      recid == "GLS" ~ "GLS-Unknown",
+      .default = "Acute-Unknown"
     )
   } else if (all(recid == "HC")) {
     # Home care
--- a/R/add_smr_type.R
+++ b/R/add_smr_type.R
@@ -109,7 +109,7 @@ add_smr_type <- function(recid,
 
   # Situation where an Acute/GLS recid is given but no ipdc marker
   if (any(recid %in% c("01B", "GLS")) & missing(ipdc)) {
-    cli::cli_warn(
+    cli::cli_abort(
       "An {.var ipdc} vector has not been supplied, and therefore Acute/GLS
                    records cannot be given an {.var smrtype}"
     )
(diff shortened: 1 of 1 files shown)


R/add_smr_type.R
```diff
@@ -103,7 +103,7 @@ add_smr_type <- function(recid,
 
   # Situation where an Acute/GLS recid is given but no ipdc marker
   if (any(recid %in% c("01B", "GLS")) & missing(ipdc)) {
-    cli::cli_abort(
+    cli::cli_warn(
       "An {.var ipdc} vector has not been supplied, and therefore Acute/GLS
                    records cannot be given an {.var smrtype}"
     )
```
Moohan, line 109: I think this should still be an error. If `ipdc` isn't supplied that's incorrect and should error which is different from the variable being supplied and some being `NA`.

R/add_smr_type.R
```diff
@@ -63,7 +63,7 @@ add_smr_type <- function(recid,
 
   # Situation where acute records are present without a corresponding ipdc
   if (all(recid %in% c("01B", "GLS")) & anyNA(ipdc)) {
-    cli::cli_abort(
+    cli::cli_warn(
       "In Acute records, {.var ipdc} is required to assign an smrtype,
                     and there are some {.val NA} values. Please check the data."
     )
```
Moohan, line 69: I think we maybe want a 'sub-check' for if `ipdc` is supplied but *all* are `NA`.

```r
if (all(is.na(ipdc))) {
   cli::cli_abort("In Acute records, {.var ipdc} is required to assign an smrtype,
                    but all values are {.val NA}. Please check the code/data."
}
```

R/add_smr_type.R
```diff
@@ -137,7 +137,8 @@ add_smr_type <- function(recid,
     smrtype <- dplyr::case_when(
       recid == "01B" & ipdc == "I" ~ "Acute-IP",
       recid == "01B" & ipdc == "D" ~ "Acute-DC",
-      recid == "GLS" & ipdc == "I" ~ "GLS-IP"
+      recid == "GLS" & ipdc == "I" ~ "GLS-IP",
+      TRUE ~ "Acute-Unknown"
```
Moohan, line 141:
```suggestion
      recid == "GLS" & ipdc == "I" ~ "GLS-IP",
      recid == "GLS" ~ "GLS-Unknown",
      .default = "Acute-Unknown"
```

Moohan (changes requested):

PR #646, opened 2023-05-02
Fix bug with monthly cost/beddays

Jennit07:
Issue here was that the `convert_monthly_rows_to_vars` function was returning an error due to the way this function was selecting variables ending `xxx_cost` and `xxx_beddays` which did not exist in the data e.g 2022/23 file where all months are not available. This has now been corrected using the `any_of()` function. I also have brought this branch up to date with #644 as otherwise this would have created conflicts and i was testing the same data and running into the issues in #644


Moohan (approved):
Happy to approve this but I'm not going to merge it as there are still outstanding comments on #644 which are related.

PR #647, opened 2023-05-02
Fix another bug related to #646

Moohan:
See https://github.com/Public-Health-Scotland/source-linkage-files/pull/646. This is the same issue cropping up in a different function. #646 resolved it for acute but this will resolve it for Social Care (and others)


lizihao-anu (approved):
Awesome fix!

PR #656, opened 2023-05-15
Fix home care cost bug

Jennit07:
Small typo of cost`s` instead of `cost` causing the tests to show a drop in `cost_total_net` by 100%

### DIFF
--- a/R/process_extract_home_care.R
+++ b/R/process_extract_home_care.R
@@ -66,7 +66,7 @@ process_extract_home_care <- function(
     # remove cost variables not from current year
     dplyr::select(-(tidyselect::contains("hc_cost_2"))) %>%
     # create cost total net
-    dplyr::mutate(cost_total_net = rowSums(dplyr::across(tidyselect::contains("hc_costs_q"))))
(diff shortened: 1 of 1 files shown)


R/process_extract_home_care.R
```diff
@@ -66,7 +66,7 @@ process_extract_home_care <- function(
     # remove cost variables not from current year
     dplyr::select(-(tidyselect::contains("hc_cost_2"))) %>%
     # create cost total net
-    dplyr::mutate(cost_total_net = rowSums(dplyr::across(tidyselect::contains("hc_costs_q"))))
+    dplyr::mutate(cost_total_net = rowSums(dplyr::across(tidyselect::contains("hc_cost_q"))))
```
Moohan, line 69: We might as well swap this to use `pick` whilst making this change. Using `across` without a function will be deprecated at some point.
```suggestion
    dplyr::mutate(cost_total_net = rowSums(dplyr::pick(tidyselect::contains("hc_cost_q"))))
```

Moohan (approved):
Looks good to me, simple issue, simple fix!

PR #660, opened 2023-05-15
Update geography lookups to use parquet format

Jennit07:
Update to use the new parquet files for geographies. Tested the files and happy they are working against our code. I've also updated some of the code to reflect those deprecated in tidyselect 1.2.0.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -58,6 +58,7 @@ Eileanan
 envir
 feb
 fileext
+Finalise
 fst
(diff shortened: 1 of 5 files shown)


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4981509865/attempts/1).

Moohan (approved):
Just tested and it produces the same results in all cases I could find.

I'll go back the the geog team with that answer

PR #663, opened 2023-05-16
Update to use the csv version of gpprac

Moohan:
The new csv file is in a sub folder so I've had to take that into account. The variable names have also changed. I've updated the tests and added some new ones to track the variable names incase something happens to them in the future.

### DIFF
--- a/R/get_lookup_paths.R
+++ b/R/get_lookup_paths.R
@@ -130,9 +130,12 @@ get_datazone_pop_path <- function(file_name = NULL, ext = "rds") {
 #' @export
 #'
 #' @family lookup file paths
-get_gpprac_ref_path <- function(ext = "sav") {
-  gpprac_dir <-
-    fs::path(get_lookups_dir(), "National Reference Files")
+get_gpprac_ref_path <- function(ext = "csv") {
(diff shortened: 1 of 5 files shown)


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/4991517110/attempts/1).

R/get_lookup_paths.R
```diff
@@ -130,9 +130,8 @@ get_datazone_pop_path <- function(file_name = NULL, ext = "rds") {
 #' @export
 #'
 #' @family lookup file paths
-get_gpprac_ref_path <- function(ext = "sav") {
-  gpprac_dir <-
-    fs::path(get_lookups_dir(), "National Reference Files")
+get_gpprac_ref_path <- function(ext = "csv") {
+  gpprac_dir <- fs::path(get_lookups_dir(), "National Reference Files")
```
Jennit07, line 134: This looks good. My only query - in the email this said there was a `csv files` folder, should we reference this here? I had a look and there are csv files outside of the `csv files` folder and also inside of this.

PR #665, opened 2023-05-17
care home fix

lizihao-anu:
There are two issues(I feel I am Moles using a magnifier to go through every inch of the suspicious area -__- ), which are in the create_monthly_costs and process_sc_all_social_care respectively. 

- create_monthly_costs. This issue was introduced in the patch fixing the bug for daily cases in A&E. There are two main modifications for this issue. The first is to use `yearstay==0.33` to determine whether it is a daily case. The second is to manually add columns for monthly costs. This is because when there is no daily case, dplyr::pivot_wider cannot produce 12 columns for monthly costs.
-  process_sc_all_social_care. When processing all social care data, the script did not correctly pick up the right episode for duplicated episodes.

After this fix, the tests produce reasonable results. The only thing worth mentioning is the items, the mean_monthly_costs, look strange because SPSS data use NA rather than 0 where the costs are 0.


R/create_monthly_costs.R
```diff
@@ -52,17 +57,24 @@ create_monthly_costs <- function(data,
...
-          paste0("_cost")
-      ),
+      tidyselect::any_of(full_cost_col),
       "daycase_check"
     )
 
   available_months <- setdiff(names(daycase_cost_months), "daycase_check")
+  add_months <- setdiff(full_cost_col, available_months)
+
+  add_months_df <- data.frame(matrix(0, nrow = nrow(data), ncol = length(add_months)))
```
Moohan, line 67: Using `tibbles` for consistency with other code.
```suggestion
add_months_df <- dplyr::as_tibble(
  matrix(0, nrow = nrow(data), ncol = length(add_months))
)```

File: R/create_monthly_costs.R [hunk 2]
```diff
@@ -52,17 +57,24 @@ create_monthly_costs <- function(data,
...
+  colnames(add_months_df) <- add_months
```
Moohan, line 68: I think `names()` is the same as `colnames()` here? If so I find `names()` more intuitive.
```suggestion
  names(add_months_df) <- add_months
```

R/create_monthly_costs.R
```diff
@@ -52,17 +57,24 @@ create_monthly_costs <- function(data,
...
+
+  daycase_cost_months <- daycase_cost_months %>%
+    cbind(add_months_df) %>%
```
Moohan, line 71: Prefer `dplyr::bind_cols` as it's faster and does some other checks on names and types etc.
I think it should work with no issues here.
```suggestion
    dplyr::bind_cols(add_months_df) %>%
```

R/process_sc_all_care_home.R
```diff
@@ -125,8 +125,22 @@ process_sc_all_care_home <- function(
...
+    dplyr::arrange(
+      desc(.data[["period"]]),
+      desc(.data[["ch_discharge_date"]]),
+      desc(.data[["ch_provider"]]),
+      desc(.data[["record_date"]]),
+      desc(.data[["qtr_start"]]),
+      desc(.data[["ch_name"]]),
+      desc(.data[["ch_postcode"]]),
+      desc(.data[["reason_for_admission"]]),
+      desc(.data[["type_of_admission"]]),
```
Moohan, line 137:
```suggestion
      dplyr::desc(.data[["period"]]),
      dplyr::desc(.data[["ch_discharge_date"]]),
      dplyr::desc(.data[["ch_provider"]]),
      dplyr::desc(.data[["record_date"]]),
      dplyr::desc(.data[["qtr_start"]]),
      dplyr::desc(.data[["ch_name"]]),
      dplyr::desc(.data[["ch_postcode"]]),
      dplyr::desc(.data[["reason_for_admission"]]),
      dplyr::desc(.data[["type_of_admission"]]),
```

Moohan (changes requested):
Great work! No issues really, just some code tweak suggestions.

One additional thing to do would be to write some tests for `create_monthly_costs` we put this off last time but that backfired so I think we should add some tests now.

Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/5059587144/attempts/1).

R/create_monthly_costs.R
```diff
@@ -52,17 +57,26 @@ create_monthly_costs <- function(data,
...
-      tidyselect::any_of(
-        month.abb[c(4:12, 1:3)] %>%
-          tolower() %>%
-          paste0("_cost")
-      ),
+      tidyselect::any_of(full_cost_col),
       "daycase_check"
     )
 
   available_months <- setdiff(names(daycase_cost_months), "daycase_check")
```
Moohan, line 64: I think this line is obsolete now?
```suggestion
```

Moohan (approved):

Moohan (approved):

PR #667, opened 2023-05-19
Tweaks to DN

Moohan:
The main change here is to recode `hbtreatcode` back to using HB2018 variables. This should correct the issue we saw in tests where some people (and their costs) had moved between Lanarkshire and Glasgow. I suspect that the datamart used to use HB2018 and has only recently switched to HB2019 as the R code seemed to be the same as the SPSS.

Some other changes I made were to expose the cost lookup as a parameter and add this to targets so that it now explicitly depends on the cost lookup. Otherwise, some minor tweaks to make the code more readable etc.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -11,6 +11,7 @@ beddays
 birthtime
 bodyloc
 boxi
+callr
 Canx
 carehome
 careinspectorate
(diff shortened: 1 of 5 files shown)


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/5023075064/attempts/1).

Jennit07 (approved):
Looks good to me!

Issue #625, opened 2023-04-17
Review the `attendance_status` variable


Moohan:
The `attendance_status` variable comes from GPOoHs (`Consultation Recorded`) and Outpatients (`Clinic Attendance (Status) Code`).

It was tidied up in this commit https://github.com/Public-Health-Scotland/source-linkage-files/commit/192a5c293b7d7b6855ed1702ed079d0854ab0b18 so that it now uses the [SMR codes / definitions](https://publichealthscotland.scot/services/national-data-catalogue/data-dictionary/search-the-data-dictionary/attendance-status/).

At one point the SLF team had discussed using factors to store these types of variables with (numeric) codes so that we could have the labels alongside. I think we moved away from the factor idea as 1) It doesn't really work for codes which are not exclusively numeric 2) Working with factors might not be obvious to users 3) It has the potential to break a lot of existing code in non-obvious ways.

The issue was not taken forward 'fully' due to `attendance_status` being used in the deaths cleanup code. https://github.com/Public-Health-Scotland/source-linkage-files/blob/f047aab76a4f6b4ceec3d33e0a7a5020ee43f5c9/_SPSS_archived/2014-15/C06%20Death%20fixes.sps#L33 

This issue is now probably redundant as the deaths cleanup has been handled differently and I don't think you'll want to change the coding of `attendance_status`. It would be worth a look through the code - following `attendance_status` and just making sure it still makes sense and is consistent.

[All uses of `attendance_status`](https://github.com/search?q=repo%3APublic-Health-Scotland%2Fsource-linkage-files%20attendance_status&type=code)

<details>
<summary>Old issue text</summary>

We should probably change the GP OoH `attendance_status` (`"Consultation Recorded"`) and the outpatient `attendance_status` ("Clinic Attendance (Status) Code") to both be a factor with something like:
`factor(<ooh_var>, levels = c("Y", "N"), labels = c("Patient Seen", "DNA"))` and
`factor(<outpatient_var>, levels = c(1, 5, 8), labels = c("Patient Seen", "CNW", "DNA"))`.

But this would also involve tweaking some other code which uses the 1 / 8 values, so out of scope for this PR. For now, I think it's sufficient to alter the GP OoH values this just needs to be done at the right place.

</details>

Moohan:
@SwiftySalmon I update the issue description above.

PR #670, opened 2023-05-24
Include episode file when processing targets

Jennit07:
I've tidied up the targets code here to include all extracts and this also needs testing in the targets pipeline but first pass at processing the episode file in targets.


R/process_extract_gp_ooh.R
```diff
@@ -82,6 +82,10 @@ process_extract_gp_ooh <- function(year, data_list, write_to_disk = TRUE) {
     dplyr::mutate(
       record_keydate1 = as.Date(.data$record_keydate1),
       record_keydate2 = as.Date(.data$record_keydate2)
+    ) %>%
+    # change attendance_status into numeric
+    dplyr::mutate(
+      attendance_status = as.double(attendance_status)
```
Moohan, line 88: I thought we had already resolved this... not sure how the episode files have been running properly if not?

That said, issue #625 relates to this but in the meantime, rather than introducing new code here, could we just make sure it's read in as a numeric for GP OoHs.

- [ ] Change [this line](https://github.com/Public-Health-Scotland/source-linkage-files/blob/b1442a01e3ef79783edb460e48416812cc4b15e5/R/read_extract_ooh_consultations.R#L16) to  `"Consultation Recorded" = readr::col_integer()`.
- [ ] Remove [this block of code](https://github.com/Public-Health-Scotland/source-linkage-files/blob/b1442a01e3ef79783edb460e48416812cc4b15e5/R/process_extract_consultations.R#L39).

_targets.R
```diff
@@ -19,7 +19,7 @@ tar_option_set(
   memory = "persistent" # default option
 )
 
-years_to_run <- c("1920")
+years_to_run <- c("1819")
```
Moohan, line 22: Just a reversion as it isn't a meaningful change.
```suggestion
years_to_run <- c("1920")
```

Moohan (changes requested):

R/process_extract_gp_ooh.R
Moohan: I was just going to make this change but realised it should already be a numeric by this point... So I think this bit of code can just be removed

Moohan (approved):

PR #673, opened 2023-05-25
Add tests to the targets pipeline

Moohan:
Outstanding tasks
- [x] Create a fix for tests where there is no data in that year e.g. DN 2223

### DIFF
--- a/R/process_tests_cmh.R
+++ b/R/process_tests_cmh.R
@@ -7,6 +7,11 @@
 #'
 #' @export
 process_tests_cmh <- function(data, year) {
+  if (identical(data, tibble::tibble())) {
+    # Deal with years where we have no data
+    return(data)
+  }
+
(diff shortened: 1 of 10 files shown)


Moohan:
Although the tests have mostly worked I think it would be worth amending the `write_tests` function to write a new file per year (per update). Targets writes the tests as soon as it's done which looks like a pretty random order.

Moohan:
> @check-spelling-bot apply updates.

Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/5132257424/attempts/1).

Jennit07:
`write_tests` function now merged so tests should write out per year now

Jennit07 (approved):
Looks good, tested this while running targets and working as expected

PR #675, opened 2023-05-30
Add some tests for the DD file

Moohan:
This will hopefully highlight any issues with the DD file immediately. The new tests:
 * Check the file is an `rds`. (Might need updating in the future!)
 * Check the file has more than 150K rows.
 * Checks that at least 98% of the CHIs are valid.
 * Check the variable names and types are what we expect (using a snapshot).


Jennit07 (approved):
Looks good!

PR #676, opened 2023-05-31
Fix bug in delayed discharges

Jennit07:
Fixed a bug which caused the processed delayed discharges output to have 0 rows. The issue was that the raw extract had changed to date format for the key variables and our code was wanting to change this to dates causing the variables to return NA. This NA was passed through which could not be filtered on causing the drop.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -17,6 +17,7 @@ carehome
 careinspectorate
 cattend
 CCYY
+cennum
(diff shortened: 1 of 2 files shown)


Moohan (approved):

Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/5131629875/attempts/1).

PR #677, opened 2023-05-31
D01 2 zihao

lizihao-anu:
The new version `aggregate_by_chi` is in the branch `d01_2_zihao` with the file name ` aggregate_by_chi`_zihao.R`. The reason I do so is not to cause confusion or corruption to the current structure. Feel free to rename it and merge the branch if you satisfy with the new version.

I conducted a benchmark comparison between the old version and the new version of the `aggregate_by_chi` function. The benchmark results are based on a test dataset with 10,000 (10^4) rows, as the complete benchmark on the real dataset is still pending due to time and resource limitations. However, I did successfully run the new version on the real dataset, which consisted of approximately 10 million (10^7) rows.

Here is the summary of benchmark results for the test dataset:

```
Benchmark Results:
-------------------------------------
Expression |    Time     |  Memory
-------------------------------------
old        |  2.33 min   |  953 MB
new        | 18.38 sec   |  546 MB
```

It is important to note that these benchmark results were obtained using a test dataset with 10,000 rows. Given that our actual datasets typically have around 10 million rows, we can estimate substantial improvements in time and memory consumption when applying the new function to real data sets. Although the exact reduction estimates for the real datasets may vary, based on the observed improvements in the benchmark results, we can anticipate a boost in performance and efficiency at a similar level. 

However, one issue is loading package `data.table`. Simply `load_all()` seems not working and I have to `source(R\aggregate_by_chi_zihao.R)` to make it work. Also, I am still hesitating about involving `data.table`.

Feel free to comment.

------------------
Update on Wednesday, 31/05/2023
The resource consumption is reduced to 82GB and 49 min to run the whole `create_individual_file()` with the new version of `aggreate_by_chi()` on the whole 2223 data with 10^7 rows. Below are the memory and time used.

image
```
> source("~/source-linkage-files/zihao/individual_file_test.R")
ℹ Loading createslf
ℹ Clean individual file function started at 2023-05-31 17:50:53
ℹ Aggregate by CHI function started at 2023-05-31 17:50:53
ℹ Recode Gender function started at 2023-05-31 17:50:53
ℹ Clean up CH function started at 2023-05-31 17:50:53
ℹ Aggregate ch episodes function started at 2023-05-31 17:50:53
ℹ Add all columns function started at 2023-05-31 17:50:53
ℹ Add cij columns function started at 2023-05-31 17:50:53
ℹ Remove blank CHI function started at 2023-05-31 17:50:53
Time difference of 49.14286 mins
```


lizihao-anu:
Importing `data.table` in the `NAMESPACE` and remove redundant columns of DoB, gpprac and postcode. Now it should be cool for review again.

R/aggregate_by_chi_zihao.R
```diff
@@ -0,0 +1,195 @@
...
+  chunk_size <- min(nrow(episode_file), 5e7)
+  # Adjust the chunk size as per your system's memory capacity
+  n_chunks <- nrow(episode_file) %/% chunk_size
+
+
+  # colums specification
+  cols2 <- names(episode_file)[grepl("postcode$|DoB$|gpprac$",
+    names(episode_file),
+    ignore.case = TRUE
+  )]
```
Moohan, line 64: I think this is equivalent, `str_ends` is makes the regex more readable I think.
```suggestion
  cols2 <- names(episode_file)[stringr::str_ends(
  string = names(episode_file),
  pattern = stringr::regex(
    "postcode|DoB|gpprac",
    ignore_case = TRUE
  )
)]
```

R/aggregate_by_chi_zihao.R
```diff
@@ -0,0 +1,195 @@
...
+      # "DN$",
+      "NHS24$",
+      "PCC$",
+      "_dnas$",
+      "^SDS_option",
+      sep = "|"
+    ),
+    names(episode_file),
+    ignore.case = TRUE
+  )]
```
Moohan, line 101: As above, can we use `str_ends` here instead of `grepl` + `$` in the regexp?

Also is the `ignore.case = TRUE` necessary? I think all variables should be lowercase (camel_case) and if they're not we should make them, rather than having to check for any case.

R/aggregate_by_chi_zihao.R
```diff
@@ -0,0 +1,195 @@
...
+  # cols5 <- cols5[!(cols5 %in% c("sc_send_lca", "sc_latest_submission"))]
+  cols6 <- c(
+    condition_cols(),
+    # "death_date",
+    # "deceased",
+    "year",
+    names(episode_file)[grepl("_Cohort$|end_fy$|start_fy$",
+      names(episode_file),
+      ignore.case = TRUE
+    )]
```
Moohan, line 113: Now I've seen this block a few times, could we make it a function?

```r
vars_end_with <- function(data, vars) {
  names(episode_file)[stringr::str_ends(
    names(episode_file),
    stringr::regex(paste(vars, sep = "|"), ignore.case = TRUE)
  )]
}

```

R/aggregate_by_chi_zihao.R
```diff
@@ -0,0 +1,195 @@
...
+      by = chi
+    ]
+    chunk_agg <- cbind(
+      chunk_cols1,
+      chunk_cols2[, chi := NULL],
+      chunk_cols3[, chi := NULL],
+      chunk_cols4[, chi := NULL],
+      # chunk_cols5[, chi := NULL],
+      chunk_cols6[, chi := NULL]
+    )
```
Moohan, line 160: `dplyr::bind_cols` is often faster and does some better error checking (i.e. it's more strict which can be useful for not having weird surprises).
```suggestion
    chunk_agg <- dplyr::bind_cols(
      chunk_cols1,
      chunk_cols2[, chi := NULL],
      chunk_cols3[, chi := NULL],
      chunk_cols4[, chi := NULL],
      # chunk_cols5[, chi := NULL],
      chunk_cols6[, chi := NULL]
    )
```

R/link_delayed_discharge_eps.R
```diff
@@ -271,37 +271,47 @@ link_delayed_discharge_eps <- function(data, year) {
...
+    dplyr::mutate(has_delay = dplyr::if_else(
+      .data$chi != "" & !is.na(.data$cij_marker),
+      .data$smrtype == "DD-CIJ",
+      NA
+    )) %>%
+    dplyr::group_by(chi, cij_marker) %>%
+    dplyr::mutate(cij_delay = max(has_delay)) %>%
+    dplyr::ungroup()
+  # tidy up and rename columns to match the format of episode files
+  dplyr::select(
```
Moohan, line 284:
```suggestion
    dplyr::ungroup() %>%
  # tidy up and rename columns to match the format of episode files
  dplyr::select(
```

PR #679, opened 2023-06-01
Update `write_tests_xlsx` to use a workbook per FY

Moohan:
The workbook was pretty crowded with tests when running multiple years so now this function will create a 'test workbook' per financial year plus another one for lookups etc.]

It will now sort the workbook so that tests of a given dataset will appear together and it should generally be easier to use.

A final addition was to add an alert message when the tests have been successfully written.


Jennit07 (approved):
Looks good! will approve/merge

PR #680, opened 2023-06-01
Create tests for the individual file

Moohan:

PR #682, opened 2023-06-08
Reduce the memory usage of the tests

Moohan:
The main saving here is for the episode file tests (the same improvements should be applied to the individual file tests). By only keeping the variables which are actually needed for the tests some memory is saved (~60%) but importantly we then only read those variables in from the 'existing' file which makes that part faster and also use less memory.

Other changes are minor and only speed things up in a small way.


Jennit07 (approved):
Changes look really good and the memory consumption is greatly improved! Changes should also be implemented in #680. Happy to approve/merge

PR #683, opened 2023-06-09
Fixing some variables in the episode file

Moohan:
Some variables were being duplicated by a `*_join`, me and @Jennit07 identified these:

- `location`, was being 'stored' but now it's used in the DD linkage part of the code
- `amended_dates`, was being kept from the DD data but isn't required as the data is encoded in `dd_quality`
- For the `service_use` and `demographic` cohorts, all the variables were being matched on but we only include the actual cohort variables in the SLF.


Jennit07 (approved):
Looks good! Happy to merge

PR #690, opened 2023-06-19
Add `cij_delay`

Moohan:
This was added to Zihao's individual file branch but should be included generally.


Jennit07 (approved):
Looks good

PR #693, opened 2023-06-20
Update the GP practice lookup

Moohan:
I noticed the test failed as the team had changed the variable names in the gpprac.csv file (again). I made some other changes including to the Open Data as `{phsopendata}` has now received an update which lets us extract all resources!

 * Open data function now extracts all resources instead of the most recent 20.
 * Open data code has been cleaned up with redundant comments removed and some code lines trimmed.
 * I tidied the documentation for `process_lookup_gpprac`.
 * Postcodes in the gpprac ref are now forced to PC7 format for consistency.
 * I've used the `col_select` argument from `read_parquet` to read the needed cols from the SPD much quicker.
   * I updated `read_file` to document the `read_parquet` arguments and to throw an error when they are used on a non-parquet file.
 * I swapped the matching order - I think this was wrong. It now starts with the GPprac ref and matches on cluster info if we have it. This results in many more GP practices being in the lookup, some without cluster info but still with postcode and hbprac info.
 * I've added error checks to the joins to highlight any issues in the future.


R/read_file.R
Moohan, line 1: I think there's scope for us to use `col_select` and even `as_data_frame = FALSE` in more places, now we're using `parquet` as standard.

`as_data_frame` works by returning an 'Arrow Table' instead of a tibble which you then use `{dplyr}` functions on and finally a `dplyr::collect()`, at which point it actually reads the (needed) data from the file and runs the commands thus making it more efficient. It's equivalent to using `{dbplyr}` on a database but for `parquet` files.

R/read_file.R
```diff
@@ -10,11 +10,12 @@
 #'  * `.parquet` uses [arrow::read_parquet()].
 #'
 #' @param path The file path to be read
+#' @inheritParams arrow::read_parquet
 #' @param ... Addition arguments passed to the relevant function.
 #'
 #' @return the data a [tibble][tibble::tibble-package]
 #' @export
-read_file <- function(path, ...) {
+read_file <- function(path, col_select = NULL, as_data_frame = TRUE, ...) {
```
Jennit07, line 18: I tested this with the code: `ep1718 <- read_file(slf_path, col_select = "recid")` but this still read in all of the variables. Have i done this wrong? Other than that i think the changes on this PR look good

Moohan: Nope, not wrong! I should have fixed it in this commit: https://github.com/Public-Health-Scotland/source-linkage-files/pull/693/commits/93ab9ed71028225bb3ad58ec908cc1299bac6128

Jennit07: Working as expected!

Jennit07 (approved):
Looks good! ready to merge

PR #695, opened 2023-06-20
Use `col_select` from `read_parquet`

Moohan:
Making better use of the parquet format.


R/get_slf_lookup_paths.R
```diff
@@ -54,7 +56,8 @@ get_slf_gpprac_path <- function(update = latest_update(), ...) {
 get_slf_chi_deaths_path <- function(update = latest_update(), ...) {
   slf_chi_deaths_path <- get_file_path(
     directory = fs::path(get_slf_dir(), "Deaths"),
-    file_name = stringr::str_glue("chi_deaths_{update}.parquet"),
+    file_name = stringr::str_glue("chi_deaths_{update})
+    ext = "parquet",
```
Moohan, line 60:
```suggestion
    file_name = stringr::str_glue("chi_deaths_{update}"),
    ext = "parquet",
```

R/fill_geographies.R
```diff
@@ -126,8 +126,10 @@ fill_postcode_geogs <- function(data) {
 }
 
 fill_gpprac_geographies <- function(data) {
-  gpprac_ref <- read_file(get_slf_gpprac_path()) %>%
-    dplyr::select("gpprac", "cluster", "hbpraccode")
+  gpprac_ref <- read_file(
+    get_slf_gpprac_path(),
+    col_select = c("gpprac", "cluster", "hbpraccode")
+  )
```
Jennit07, line 132: I've checked other uses for `col_select` in this PR and they are working. For this one, it seems to read in 8 columns? I'm not too sure why as the code looks right to me

Jennit07 (changes requested):
One comment, everything else looks good

R/fill_geographies.R
Moohan: I think the issue here is that the existing file is an `rds` so it reads that and doesn't use the `col_select` parameter at all. I introduced an error so that if you use the `col_select` or `as_data_frame` parameters on a non `parquet` file it tells you.

I've updated this branch so hopefully you will also see that error... It should all work out once the files have been written out!

Jennit07 (approved):
Ready to approve/merge. One thing to note that due to the gpprac and pc lookup not being ran in parquet format yet, this is causing the tests to fail. I think this will correct itself when we have these files

PR #700, opened 2023-06-21
Remove the anon_chi functions

Moohan:
These are no longer needed as they are implemented in `{slfhelper}` and we can just import from there instead of maintaining two versions.

I checked and as far as I can tell there are no uses of these functions in the code that needed to be removed/changed.


Jennit07 (approved):
Looks good/ ready to merge

PR #704, opened 2023-06-26
Update some parameters and documentation

Moohan:
This is to harmonise the documentation and make it slightly easier to maintain.


Jennit07 (approved):
Looks good! I agree this should be called `data` to make the functions less confusing

PR #707, opened 2023-06-27
Updates to SC client

Moohan:
So far some minor updates to ensure all the client flags are kept from the data mart. However, I would like to change how the client file is used from the current: Matching it on to the Social Care data sets, to using it like the LTC, deaths, HHG etc lookups and matching it on for all CHIs in the episode and individual files. This second part hasn't been implemented yet, I'm not sure whether to merge these changes and do the second at some point later or to leave this and do everything in one go!


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/5443535341/attempts/1).

PR #708, opened 2023-06-28
Include write to disk option for individual file 

Jennit07:
Write to disk was missing from the individual file so i have added this back in along with a year parameter.


Moohan (approved):
Looks good

PR #724, opened 2023-07-12
Join slf lookups onto individual file

Jennit07:
Join slf lookups onto individual file. This includes pc lookup and gpprac lookup. Ive tested this by matching onto 2021 individual file and i think its working as expected.


R/create_individual_file.R
```diff
@@ -760,3 +762,28 @@ clean_up_gender <- function(individual_file) {
...
+    dplyr::left_join(
+      read_file(
+        get_slf_gpprac_path()
+      ) %>%
+        dplyr::select(c("gpprac", "cluster", "hbpraccode")),
+      by = "gpprac"
+    )
+
+  return(individual_file)
+}
```
Moohan, line 789: So it works nicely with targets, could we have the lookups passed as parameters please e.g.

```suggestion
join_slf_lookup_vars <- function(
  individual_file,
  slf_postcode_lookup = read_file(get_slf_postcode_path()),
  slf_gpprac_lookup = read_file(
                       get_slf_gpprac_path(), 
                       col_select = c("gpprac", "cluster", "hbpraccode")
                       )
  ) {
  individual_file <- individual_file %>%
    dplyr::left_join(
      slf_postcode_lookup,
      by = "postcode"
    ) %>%
    dplyr::left_join(
      slf_gpprac_lookup %>%
        dplyr::select("gpprac", "cluster", "hbpraccode"),
      by = "gpprac"
    )

  return(individual_file)
}
```

Moohan: I think we can resolve the `hbrescode` here by adding `dplyr::rename(hbrescode = "HB2018")` - in fact could we have the hbrescode variable as a parameter so it's simpler to change in the future? e.g. `dplyr::rename(hbrescode = hbrescode_var)`.

Jennit07:
@Moohan Changes made, could you please review again? thanks

R/create_individual_file.R
Moohan, line 1: I think this function also needs to be added to the `create_individual_file()` flow, but otherwise is looking good!

Jennit07: I think i had already added this on line 60?

Moohan: Yep, odd, I feel like that didn't show for me last time!

R/create_individual_file.R
```diff
@@ -42,7 +42,8 @@ create_individual_file <- function(episode_file, year, write_to_disk = TRUE) {
       "sc_latest_submission",
       "hc_hours_annual",
       "hc_reablement",
-      "ooh_case_id"
+      "ooh_case_id",
+      "hbrescode"
```
Moohan, line 46:
```suggestion
      "ooh_case_id"
```

Moohan (approved):

R/create_individual_file.R
Jennit07: Thanks @Moohan, that's strange!

PR #726, opened 2023-07-13
Write out as a partitioned arrow dataset

Moohan:
This will allow us to test if this format provides any additional benefits. It is still also written out as a (non-partitioned) parquet file.


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/5573366782/attempts/1).

Jennit07 (approved):
Looks good

PR #727, opened 2023-07-13
Some updates to `read_file()`

Moohan:
Some small updates to `read_file` 

- Explicit handling of `.csv.gz` and `.rds.gz` extensions - previously it assumed any file ending `.gz` was a CSV.
- A nicer workaround for `read_parquet`, see: https://github.com/apache/arrow/issues/36658
- Updated so that reading an `.fst` file will always return a tibble.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -94,6 +94,7 @@ itle
 iwalk
 jaccard
 jan
+jennifer
 jul
 keydate
 keyring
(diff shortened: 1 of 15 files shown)


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/5573365929/attempts/1).

Jennit07 (approved):
Looks good, ready to merge

PR #747, opened 2023-07-17
Make episode file output with `anon_chi`

Moohan:
I've added this as a parameter so you can output CHI if desired, but the default is for anon_chi.

For the tests, it swaps back to CHI as there are some tests which specifically us the CHI number.


Jennit07 (approved):
Looks good. Ready to merge

PR #748, opened 2023-07-17
Output the individual file with `anon_chi`

Moohan:
By default it expects the input episode file to have `anon_chi` and it will output `anon_chi` but I've added parameters to deal with other situations if required.


Jennit07 (approved):
happy to merge

PR #750, opened 2023-07-18
Write the episode file as a partitioned dataset

Moohan:
This PR reverts #726 and moves the code to targets. Since this code was just meant as a test, I think that's fine and doing it this way will make sure it doesn't hold up the rest of the episode file (or individual file) processing.


Jennit07 (approved):
Makes sense to do it this way instead to prevent extra processing times. Will approve/merge

PR #752, opened 2023-07-19
Fix for anon_chi missing

Moohan:
`slfhelper::get_anon_chi` converts `NA` to "TkE=" which would then be converted back to "e1" 🤯

This fixes that by making NA CHIs blank `""`, then making blank anon_chis NA after.

slfhelper should be updated so that it always converts NA to NA.


Jennit07 (approved):
Looks good. Ready to merge

PR #755, opened 2023-07-20
Fix for years with no DN data

Moohan:
21/22 and 22/23 we're failing because they didn't have `total_no_dn_contacts` this is a bit of a crude fix but should work for any year, and if the variable doesn't exist in the episode file it will be created as `NA` in the individual file.


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/5614513614/attempts/1).

Issue #689, opened 2023-06-12
Simplify `create_hscp_test_flags`


Moohan:
I think this could be simplified to return `TRUE`/`FALSE` instead of `1`/`0` as that will sum the same anyway. It would make the code:
```r
Aberdeen_City = {{hscp_var}} == "S37000001"
```
instead of:
```r
Aberdeen_City = dplyr::if_else(
        {{ hscp_var }} == "S37000001",
        1L,
        0L
      )
```
i.e. a lot shorter and easier to read

_Originally posted by @Moohan in https://github.com/Public-Health-Scotland/source-linkage-files/pull/638#discussion_r1226650735_

PR #760, opened 2023-07-26
Add tests for delayed discharges extract

Jennit07:
Addresses #671


R/process_tests_delayed_discharges.R
```diff
@@ -0,0 +1,53 @@
...
+#' for creating test flags
+#' @seealso calculate_measures
+produce_source_dd_tests <- function(data) {
+  test_flags <- data %>%
+    dplyr::mutate(
+      n_delay_episodes = 1L,
+      code9_episodes = dplyr::case_when(
+        primary_delay_reason == "9" ~ 1L,
+        TRUE ~ 0L
+      )
```
Moohan, line 44:
```suggestion
      code9_episodes = dplyr::if_else(.data$primary_delay_reason == "9", 1L, 0L)
```

Moohan: I think that actually, this will be sufficient.
```suggestion
      code9_episodes = .data$primary_delay_reason == "9"
```

As the `TRUE`/`FALSE` will get summed as though it's `1`/`0`. Similar to #689

Moohan (changes requested):
Looks good, just one minor suggestion.

Also, could you add it to the targets pipeline please (and delete the `# TODO`)?

Moohan (approved):

PR #767, opened 2023-07-31
Add simple scripts for running targets as a workbench job

Moohan:
I was using these at the last update, they're not required but were pretty useful for creating a workbench job for each year.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -22,3 +22,4 @@
 ^_targets\.R$
 ^_targets\.yaml$
 ^_SPSS_archived$
+^run_targets_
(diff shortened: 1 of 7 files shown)


Moohan:
I've opened it as a draft but please review and merge it if you're happy.

One thing that is related would be to use this code / repo: https://github.com/Public-Health-Scotland/posit-workbench-job-scheduler 

Basically, we could write a script to programmatically create the workbench jobs, which would be nice but isn't super simple to set up 😂

SwiftySalmon (commented):
looks fine, I'm just not confident enough to approve :(

PR #772, opened 2023-08-01
Simplify `create_hscp_test_flags`

Jennit07:
Addresses #689


Moohan:
This looks good. I was going to test it and realised that it isn't actually used in any of the test functions... Would you be able to (test and) add it to some of the tests, please?

Also could you make the same changes to `create_hb_test_flags`

Moohan (approved):

PR #778, opened 2023-08-02
Create HRI Variables

shintoLamp:
I have written some code to add the HRI variables to the individual file, but I need someone to check the functionality.

The main issue is that the individual file I was using for testing (2019/20) has 5,983,038 rows, and when the HRIs are matched on there's 5,986,774. So we've gained 3736 rows. I imagine that the `left_join` at the end of `add_hri_variables` is doing something unexpected.

Could someone run through the code please and try and figure out what's going on?


R/add_hri_variables.R
```diff
@@ -0,0 +1,137 @@
...
+    dplyr::mutate(
+      dummy_postcode = .data$postcode %in% c("BF010AA", "NF1 1AB", "NK010AA") |
+        stringr::str_sub(.data$postcode, 1, 4) %in% c("ZZ01", "ZZ61"),
+      eng_prac = .data$gpprac %in% c(99942, 99957, 99961, 99976, 99981, 99995, 99999),
+      keep_flag = dplyr::case_when(
+        .data$scot_flag == 0L ~ 0L,
+        (is_missing(.data$postcode) | .data$dummy_postcode) & is.na(.data$gpprac) ~ 2L,
+        !.data$eng_prac & is_missing(.data$postcode) ~ 3L,
+        !.data$eng_prac & .data$dummy_postcode ~ 4L,
+        TRUE ~ 1L
```
Moohan, line 45:
```suggestion
        .default = 1L
```

R/add_hri_variables.R
```diff
@@ -0,0 +1,137 @@
...
+  # residents or not
+  return_data <- data %>%
+    dplyr::mutate(pc_area = stringr::str_match(postcode, "^[A-Z]{1,3}")) %>%
+    dplyr::left_join(pc_areas, by = "pc_area") %>%
+    dplyr::mutate(
+      dummy_postcode = .data$postcode %in% c("BF010AA", "NF1 1AB", "NK010AA") |
+        stringr::str_sub(.data$postcode, 1, 4) %in% c("ZZ01", "ZZ61"),
+      eng_prac = .data$gpprac %in% c(99942, 99957, 99961, 99976, 99981, 99995, 99999),
+      keep_flag = dplyr::case_when(
+        .data$scot_flag == 0L ~ 0L,
```
Moohan, line 41:
```suggestion
      scot_flag = TRUE
    ) %>%
    dplyr::distinct(pc_area, scot_flag)

  # Create a flag, 'keep_flag', to determine whether individuals are Scottish
  # residents or not
  return_data <- data %>%
    dplyr::mutate(pc_area = stringr::str_match(postcode, "^[A-Z]{1,3}")) %>%
    dplyr::left_join(pc_areas, by = "pc_area") %>%
    dplyr::mutate(
      dummy_postcode = .data$postcode %in% c("BF010AA", "NF1 1AB", "NK010AA") |
        stringr::str_sub(.data$postcode, 1, 4) %in% c("ZZ01", "ZZ61"),
      eng_prac = .data$gpprac %in% c(99942, 99957, 99961, 99976, 99981, 99995, 99999),
      keep_flag = dplyr::case_when(
        .data$scot_flag ~ 0L,
```

R/add_hri_variables.R
```diff
@@ -0,0 +1,137 @@
...
+}
+
+#' Add HRI variables to an SLF Individual File
+#'
+#' @details Filters the dataset to only include Scottish residents, then
+#' creates a lookup where HRIs are calculated at Scotland, Health Board, and
+#' LCA level. Then joins on this lookup by chi/anon_chi.
+#'
+#' @param data An SLF individual file
+#' @param chi_variable Defaults to "anon_chi"
```
Moohan, line 60: Why do we need this?

R/add_hri_variables.R
```diff
@@ -0,0 +1,137 @@
...
+      "mh_episodes",
+      "gls_episodes",
+      "op_newcons_attendances",
+      # op_newcons_dnas,
+      "ae_attendances",
+      "pis_paid_items",
+      "ooh_cases"
+    ) %>%
+    flag_non_scottish_residents() %>%
+    dplyr::filter(keep_flag == 0L) %>%
```
Moohan, line 86: I'm a bit confused as to why we do all the above, to then only keep people with a Scottish postcode.

It probably makes the most sense for `keep_flag` to be called `scottish_resident` and be T/F. Then we can do `dplyr::filter(scottish_resident)` and the code will read nicely.

R/add_hri_variables.R
```diff
@@ -0,0 +1,137 @@
...
+      lca_pct = (health_net_cost / lca_cost) * 100
+    ) %>%
+    dplyr::arrange(dplyr::desc(health_net_cost), .by_group = TRUE) %>%
+    dplyr::mutate(hri_lcap = cumsum(lca_pct)) %>%
+    dplyr::ungroup() %>%
+    # Add HRI flags
+    dplyr::mutate(
+      hri_scot = hri_scotp <= 50L,
+      hri_hb = hri_hbp <= 50L,
+      hri_lca = hri_lcap <= 50L,
```
Moohan, line 116: Good work for being explicit, but these aren't actually integers, so we should be specific like this:
```suggestion
      hri_scot = hri_scotp <= 50.0,
      hri_hb = hri_hbp <= 50.0,
      hri_lca = hri_lcap <= 50.0,
```

Moohan (commented):
Looks great, I haven't tested in R and I couldn't obviously see where you might be getting duplicate rows.

One suggestion would be to use the new error checking on the joins (https://www.tidyverse.org/blog/2023/01/dplyr-1-1-0-joins/#unmatched-rows)

For a lookup you can be really explicit and do something like:
```r
left_join(
data,
lookup,
by = "lookup_var",
na_matches = "never",
unmatched = "drop",
relationship = "one-to-one")
```
I think the `relationship` arg is the only none default one there. https://dplyr.tidyverse.org/reference/mutate-joins.html#arguments

R/add_hri_variables.R
```diff
@@ -0,0 +1,137 @@
...
+#' to [get_slf_postcode_path()]
+#'
+#' @return A data frame with the variable 'keep_flag'
+flag_non_scottish_residents <- function(data,
+                                        slf_pc_lookup_path = get_slf_postcode_path()) {
+  check_variables_exist(data, c("postcode", "gpprac"))
+
+  # Make a lookup of postcode areas, which consist of the first characters
+  # of the postcode
+  pc_areas <- arrow::read_parquet(slf_pc_lookup_path) %>%
```
Moohan, line 24: Instead of passing the path, pass the actual data, we can set a default so if we don't pass anything it will still work as expected but it lets us set up the targets pipeline nicely.

```suggestion
                                        slf_pc_lookup = read_file(get_slf_postcode_path())) {
  check_variables_exist(data, c("postcode", "gpprac"))

  # Make a lookup of postcode areas, which consist of the first characters
  # of the postcode
  pc_areas <- slf_pc_lookup %>%
```

lizihao-anu:
The duplicates in individual files may contribute to the issue. I am investigating it.

Moohan:
> The main issue is that the individual file I was using for testing (2019/20) has 5,983,038 rows, and when the HRIs are matched on there's 5,986,774. So we've gained 3736 rows. I imagine that the `left_join` at the end of `add_hri_variables` is doing something unexpected.

@lizihao-anu identified that the individual file has a number of duplicated rows, this looks to be (at least in part) because the SC client extract wasn't unique. Zihao is looking at this issue separately.

I've checked the HRI code on a 'deduplicated' individual file (I just did `distinct(chi)`) and the HRI code doesn't produce any new duplicates, so I'm generally happy with it pending my other comments and a final review 😄

Moohan:
> The duplicates in individual files may contribute to the issue. I am investigating it.

Jennit07:
@lizihao-anu Check this PR please, the test would be to check the numbers against the archived 1718 fst file in hscdiip

Jennit07:
> > @lizihao-anu Check this PR please, the test would be to check the numbers against the archived 1718 fst file in hscdiip
> 
> As I check, if the duplicated rows in the 1718 individual file were eliminated, then there would not be a difference in row numbers between HRI and the individual file. It means the code is fine from this perspective.
> 
> The new individual file produced by R for 1718 is around 0.11% rows than the old file produced by SPSS. So it should be cool!

Thanks Zihao! Great work, we can discuss this tomorrow!

Jennit07 (approved):
Looks good, i think this is ready to merge!

PR #779, opened 2023-08-02
757 investigate large increase in total preventable beddays

lizihao-anu:
There is a logic difference between the SPSS code and R code, which can explain the preventable_beddays difference. If we align the R code with the SPSS logic, then the gap between `total_preventable_beddays` is reduced to 3.3%.

Not sure if it is a bug or if we decided to change the logic. @Moohan


R/aggregate_by_chi.R
```diff
@@ -127,7 +127,7 @@ aggregate_by_chi <- function(episode_file) {
     .(
       preventable_beddays = ifelse(
         max(cij_ppa, na.rm = TRUE),
```
Moohan, line 129: I was testing using this data:
```R
library(tidyverse)
library(data.table)

year <- "2223"
data <- as.data.table(tibble(
  chi = c(rep(TRUE, 4), rep(2, 4)),
  cij_total = c(TRUE, 2, 2, 3, TRUE, TRUE, TRUE, 2),
  cij_start_date = as.Date(c(
    "2023-01-01",
    "2023-02-01",
    "2023-02-01",
    "2023-03-01",
    "2023-0TRUE-01",
    "2023-0TRUE-01",
    "2023-0TRUE-01",
    "2023-03-01"
  )),
  cij_end_date = as.Date(c(
    "2023-0TRUE-31",
    "2023-02-28",
    "2023-02-28",
    "2023-05-01",
    "2023-02-01",
    "2023-02-01",
    "2023-02-01",
    "2023-04-01"
  )),
  cij_ppa = c(FALSE, TRUE, TRUE, NA, TRUE, TRUE, TRUE, TRUE)
))
```
And it seems like there might be another issue where people with `NA` for `cij_ppa` will have their beddays counted (Because of this weird behaviour `ifelse(max(NA, na.rm = TRUE), 1, 2)`).
I checked in the real data and the only `cij_ppa`s which are NA are for Delayed Discharges, which I think is a separate issue that needs to be resolved in the DD code.

This suggestion will catch that issue
```suggestion
        !all(is.na(cij_ppa)) && max(cij_ppa, na.rm = TRUE),
```

R/aggregate_by_chi.R
```diff
@@ -127,7 +127,7 @@ aggregate_by_chi <- function(episode_file) {
...
-        max(cij_end_date) - min(cij_start_date),
+        min(cij_end_date, end_fy(year)) - max(cij_start_date, start_fy(year)),
         NA_real_
```
Moohan, line 131:
```suggestion
        as.integer(min(cij_end_date, end_fy(year)) - max(cij_start_date, start_fy(year))),
        NA_integer_
```

Moohan (changes requested):
Yep, that definitely looks like an issue! We only want to count the beddays 'within the year' so it definitely should be limited like that.

This looks good, two small suggestions.

R/aggregate_by_chi.R
lizihao-anu: Great spot! Thanks for your suggestion and your enlightening, I have got a better idea for this.

lizihao-anu, line 129:
```suggestion
        any(cij_ppa, na.rm = TRUE),
```

PR #780, opened 2023-08-03
Rewrite case when statements

SwiftySalmon:
changed `case_when` to `case_match` as it's easier to read


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/5752014211/attempts/1).

Moohan (approved):
Looks (and works) great.

I added a test for `convert_sending_location_to_lca` as I noticed we didn't have one for that.

PR #781, opened 2023-08-03
Update R-CMD-check.yaml

Moohan:
Adding `development` so it will run on that branch too


Jennit07 (approved):
Looks good

Jennit07:
CMD check is failing, but i think we can merge this and fix the CMD check separately

PR #783, opened 2023-08-07
Add social care client tests

Jennit07:
Addresses #672 

I've updated the code for social care client lookup. This used to live in each year folder, however i have changed this so that the lookups live next to the demographic lookup in `/hscdiip/social_care` and supply the update month. This is so that we can have a comparison of the sc client lookup data at the current quarter vs the previous update.


R/process_tests_sc_client_lookup.R
```diff
@@ -0,0 +1,63 @@
...
+      n_sc_social_worker_not_known = .data$sc_social_worker == "Not Known",
+      n_sc_meals_yes = .data$sc_meals == "Yes",
+      n_sc_meals_no = .data$sc_meals == "No",
+      n_sc_meals_not_known = .data$sc_meals == "Not Known",
+      n_sc_day_care_yes = .data$sc_day_care == "Yes",
+      n_sc_day_care_no = .data$sc_day_care == "No",
+      n_sc_day_care_not_known = .data$sc_day_care == "Not Known",
+    ) %>%
+    # remove variables that won't be summed
+    dplyr::select(.data$Aberdeen_City:.data$n_sc_day_care_not_known) %>%
```
Moohan, line 58:
```suggestion
    dplyr::select("Aberdeen_City":"n_sc_day_care_not_known") %>%
```

R/process_tests_sc_client_lookup.R
```diff
@@ -0,0 +1,63 @@
...
+#' (data is from [get_source_extract_path()])
+#' @param max_min_vars variables used when selecting 'min-max' from [calculate_measures()]
+#' @return a dataframe with a count of each flag.
+#'
+#' @family social care test functions
+produce_tests_sc_client_lookup <- function(data) {
+  test_flags <- data %>%
+    # create test flags
+    create_sending_location_test_flags(.data$sending_location) %>%
+    dplyr::mutate(
```
Moohan, line 39: We need to do a sort to make sure the `unique_sc_id` counts properly.
```suggestion
    dplyr::arrange(.data$sending_location, .data$social_care_id) %>%
    dplyr::mutate(
```

Jennit07:
waiting for #782

Moohan (commented):
@Jennit07 Could you add this to targets too please?

R/process_tests_sc_client_lookup.R
```diff
@@ -0,0 +1,64 @@
...
+#' Social care Client lookup Tests
+#'
+#' @description Produce the test for the social care Client all episodes
+#'
+#' @param data new or old data for testing summary flags
+#' (data is from [get_source_extract_path()])
+#' @param max_min_vars variables used when selecting 'min-max' from [calculate_measures()]
+#' @return a dataframe with a count of each flag.
+#'
+#' @family social care test functions
```
Moohan, line 34: If using this 'family' (I think it's a new one) you should also edit `_pkgdown.yml` to make sure they appear there.

R/process_tests_sc_client_lookup.R
```diff
@@ -0,0 +1,64 @@
+#' Social care client lookup tests
+#'
+#' @description This script takes the processed social care client lookup and
+#' produces a test comparison with the previous data. This is written to
+#' disk as a CSV.
```
Moohan, line 5:
```suggestion
#' produces a test comparison with the previous data. This is written to
#' disk in the tests workbook.
```

Moohan (approved):
Looks good, I've approved but it definitely needs adding to targets, there's also so other minor comments about documentation.

PR #785, opened 2023-08-08
Amend `get_boxi_extract_path` function for archiving DN and CMH data 

Jennit07:
I've created a new folder for archiving DN and CMH data: 

'/conf/hscdiip/SLF_Extracts/Archived_data' 

I've amended `get_boxi_extract_path` which will now change to the new directory when DN and CMH are selected. The directory will stay the same for other extracts e.g. Acute. 

Happy to have a discussion/make amendments about this if anyone has other ideas about how this should be done. My idea was that by doing it this way, we should not have to update targets.


R/get_boxi_extract_path.R
```diff
@@ -83,3 +59,46 @@ get_dummy_boxi_extract_path <- function() {
...
+#'         now in an archived folder which is set up to point to this in
+#'         the selections.
+#' @export
+#'
+get_csv_gz_path <- function(year, type, ext = c("csv", "csv.gz")) {
+  if (type %in% c("DN", "CMH")) {
+    dir <- fs::path(get_slf_dir(), "Archived_data")
+  } else {
+    dir <- get_year_dir(year, extracts_dir = TRUE)
+  }
```
Moohan, line 79: I'm not sure what the benefit of a new function is. What would be the downside of just including this snippet in `get_boxi_extract_path`, in place of the current `year_dir <- ...` bit?

Jennit07: Thanks @Moohan I think i had over complicated this and just pushed the changes. I had tried that but then the function was not working properly for me. Possibly due to having too many things open at the time. I could try this again and remove the new function

Jennit07:
@Moohan Changes made, could you please have a look again? thanks

Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/5856792420/attempts/1).

PR #787, opened 2023-08-08
Add new 'final' file path functions

Jennit07:
Addresses #725


R/get_final_file_paths.R
Moohan, line 1: We'll want to allow `...` and then pass this through to `get_file_path` (see how the other file path functions work).

When that's done they should be `check_mode = "write"` when they're used which ensures that we have write access to the file (check for read access is the default).

SwiftySalmon (approved):

SwiftySalmon (approved):
sorted out the conflicts and tentatively approving for merge into `development`

PR #789, opened 2023-08-08
added solve for hscp names

SwiftySalmon:
In processed extract variable is called hscp, and in final SLF it's called hscp2018.

Fixed with nested if statement


Jennit07 (approved):
Looks good! Ready to merge

Issue #509, opened 2022-12-15
Check for snake_case once all SPSS has stopped


Jennit07:
I had an issue when trying to use `get_source_extract_path` function where i wanted to read the SPSS version supplying `ext = "zsav"` but this would not work due to the file name having capital letters. I have amended this on mainR so that the files will work. However, we will need to go back and use lowercase when SPSS is no longer supported. This can be done by simply changing the 'type' to snake case in the `get_source_extract_path` function.

Jennit07:
PR completed #793  will close this issue now

PR #793, opened 2023-08-09
Check scripts are in snake case

Jennit07:
Addresses #509. 

I think i have caught all instances in the code where this would need to be changed. `get_boxi_extract_path` is the raw data extract and `get_source_extract_path` is the processed extract written to disk. I have also updated this in targets to use snake_case.

I have not updated the file names in sourcedev which may cause some tests to fail but this is because we will be clearing this out again soon for the next update. 

This also has changes from #785 as i think this may cause a conflict


R/get_source_extract_path.R
```diff
@@ -39,24 +39,24 @@ get_source_extract_path <- function(year,
...
-    "Homelessness" ~ "homelessness_for_source",
-    "Maternity" ~ "maternity_for_source",
-    "MH" ~ "mental_health_for_source",
-    "DD" ~ "DD_for_source",
-    "Outpatients" ~ "outpatients_for_source",
-    "PIS" ~ "prescribing_file_for_source",
-    "SDS" ~ "SDS-for-source"
+    "acute" ~ "acute_for_source",
+    "ae" ~ "a&e_for_source",
+    "at" ~ "alarms-Ttlecare-for-source",
```
Moohan, line 44:
```suggestion
    "at" ~ "alarms-telecare-for-source",
```

SwiftySalmon (approved):
sorted conflicts with the readr functions being called

SwiftySalmon (approved):
all good

PR #794, opened 2023-08-09
Handle OpenData extracts better

Moohan:
This change was inspired by @SwiftySalmon having issues with targets where the Homelessness target was failing - this was because of an issue extracting some OpenData (an LA code to name lookup).

The OpenData extract function was a bit buried in the Homelessness function - It's now passed to the homelessness function which means it will be grabbed separately. In targets, this would mean we'd get a clearer error saying 'open data failed for x' and 'homelessness failed because of bad input'.

I've also added some tests to the two OpenData functions.


Jennit07 (approved):
Looks good!

PR #796, opened 2023-08-09
Use `get_source_extract_path` in homelessness

Moohan:
This was already set up, just not used for some reason. Note that this will switch from using a `.rds` to `.parquet` (unless you do `get_source_extract_path(year, "Homelessness", ext = "rds")`).

I also added an error check to `get_source_extract_path` as I was accidentally doing `get_source_extract_path("Homelessness")` i.e. missing out year, and it was doing some odd things!

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -91,6 +91,7 @@ hjust
 hms
 homecare
 homev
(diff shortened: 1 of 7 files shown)


Jennit07 (approved):
Ready to merge

PR #797, opened 2023-08-09
fix warning on `:=`

### LINKED ISSUE #790: Fix/investigate build warning
Assigning this to @lizihao-anu as he wrote the data.table individual file code.

There's a persistent build warning e.g.
https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/5799684816/job/15720207031#step:6:78 
```
Note: possible error in '`:=`(ch_no_cost = max(ch_no_cost), ': unused arguments (ch_no_cost = max(ch_no_cost), ch_ep_start = min(record_keydate1), ch_ep_end = max(ch_ep_end), ch_cost_per_day = mean(ch_cost_per_day)) at aggregate_by_chi.R:206 
```

It's [this line in `aggregate_ch_episodes()`](https://github.com/Public-Health-Scotland/source-linkage-files/blob/bb3dbd5d0e056249c94252dab19636a9ef44e494/R/aggregate_by_chi.R#L206) - it works but the check doesn't like it, so I wonder if it could be tweaked or written differently?

lizihao-anu:
I hope this can fix the warning. Close #790

### DIFF
--- a/R/aggregate_by_chi.R
+++ b/R/aggregate_by_chi.R
@@ -203,12 +203,18 @@ aggregate_ch_episodes <- function(episode_file) {
   data.table::setDT(episode_file)
 
   # Perform grouping and aggregation
-  episode_file[, c(
-    ch_no_cost := max(ch_no_cost),
-    ch_ep_start := min(record_keydate1),
-    ch_ep_end := max(ch_ep_end),
-    ch_cost_per_day := mean(ch_cost_per_day)
-  ), by = c("chi", "ch_chi_cis")]
+episode_file[, c(
+    "ch_no_cost",
+    "ch_ep_start",
+    "ch_ep_end",
+    "ch_cost_per_day"
+  ) := list(
+    max(ch_no_cost),
+    min(record_keydate1),
+    max(ch_ep_end),
+    mean(ch_cost_per_day)
+  ),
+  by = c("chi", "ch_chi_cis")]
 
   # Convert back to tibble if needed
   episode_file <- tibble::as_tibble(episode_file)
--- a/R/aggregate_by_chi.R
+++ b/R/aggregate_by_chi.R
@@ -203,7 +203,7 @@ aggregate_ch_episodes <- function(episode_file) {
   data.table::setDT(episode_file)
 
   # Perform grouping and aggregation
-episode_file[, c(
+  episode_file[, c(
     "ch_no_cost",
     "ch_ep_start",
     "ch_ep_end",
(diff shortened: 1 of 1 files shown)


R/aggregate_by_chi.R
```diff
@@ -203,11 +203,11 @@ aggregate_ch_episodes <- function(episode_file) {
...
-    ch_no_cost = max(ch_no_cost),
-    ch_ep_start = min(record_keydate1),
-    ch_ep_end = max(ch_ep_end),
-    ch_cost_per_day = mean(ch_cost_per_day)
+  episode_file[, c(
+    ch_no_cost := max(ch_no_cost),
+    ch_ep_start := min(record_keydate1),
+    ch_ep_end := max(ch_ep_end),
+    ch_cost_per_day := mean(ch_cost_per_day)
   ), by = c("chi", "ch_chi_cis")]
```
Moohan, line 211: This does fix the build warning but I don't think it works. I ran it against some dummy data and it wasn't working (but the original code was).

I looked at `help(":=", "data.table")` and then tested this code and I think this is a replacement that works and also doesn't trigger a build warning.
```suggestion
episode_file[, c(
    "ch_no_cost",
    "ch_ep_start",
    "ch_ep_end",
    "ch_cost_per_day"
  ) := list(
    max(ch_no_cost),
    min(record_keydate1),
    max(ch_ep_end),
    mean(ch_cost_per_day)
  ),
  by = c("chi", "ch_chi_cis")]
```

Could you have a check please Zihao?

lizihao-anu: Many thanks for this! I tested this way using dummy data and it works perfectly.

lizihao-anu:
Inspired by James, I just wanted to say more about the way adding/changing columns in `data.table`. There are 3 ways of writing `:=` I know so far. Please see the code below.

```
rm(list = ls())
# Load the data.table package
library(data.table)

# Create a sample data.table
dt <- data.table(ID = 1:5, Value = c(10, 20, 15, 30, 25))

dt[,
   `:=`(l1 = 3,
        l2 = 4)]
dt[,
   c("l3", "l4") := list(5, 6)]
dt[,
   c("l5", "l6") := .(7, 8)]
# > dt
#    ID Value l1 l2 l3 l4 l5 l6
# 1:  1    10  3  4  5  6  7  8
# 2:  2    20  3  4  5  6  7  8
# 3:  3    15  3  4  5  6  7  8
# 4:  4    30  3  4  5  6  7  8
# 5:  5    25  3  4  5  6  7  8
```

Moohan (approved):

PR #798, opened 2023-08-10
Remove incorrect references to rds

Moohan:
Since we (mostly) don't use rds anymore these references are incorrect and potentially confusing.

I've updated lots of documentation to remove the reference to rds.

I've also updated many comments that mentioned rds (these were probably the most confusing).


SwiftySalmon (approved):
mostly notes that are changed. and renaming "outfile" to something more sensible. all looks fine to me

PR #800, opened 2023-08-10
Add a new function to set up keyring

Moohan:
I've tested this by deleting my `.Renviron` and deleting my keyring `keyring::keyring_delete("createslf")` and it seems to work. Would be great to have someone with an existing set-up (Jen) test it, and to have someone who doesn't have it set up to test it.

The code looks complicated but I've just tried to catch every scenario, so the process should be smooth and clear (from the user's point of view).

I've also expanded the code relating to the username, which will now hopefully work in more cases.


Moohan:
> @check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/5824423711/attempts/1).

Jennit07 (approved):
This works for me. Would be good for someone else to test to make sure! Will approve for now

SwiftySalmon (approved):
Tested the code by deleting keyring and environ, and then running read_sc. Works fine!

PR #803, opened 2023-08-11
Rename `run_episode_file()` -> `create_episode_file()`

Moohan:
This improves consistency! When speaking to Megan we noted that having the two 'main' functions with different names was needlessly confusing!


SwiftySalmon (approved):
looks fine to me

SwiftySalmon (approved):
all looks fine. conflict should be resolved now

PR #807, opened 2023-08-14
Tests/it extract path

Moohan:
This makes the check on the IT reference slightly stricter, but the main change is to add a few more tests.


Jennit07 (approved):
Looks good/all working

PR #808, opened 2023-08-14
Add additional tests for `get_file_path`

Moohan:
This adds some additional tests for `get_file_path`. They should also be able to be ran on GitHub which is nice.


Jennit07 (approved):
Looks good!

PR #811, opened 2023-08-15
Update news for September 23 update

Jennit07:
WIP - Feel free to add/amend any points for the September 23 update. I've also added info for March and June updates.


Moohan (approved):
Looks good to me, nothing obviously missed.

Just need to remember to add the few other bits as we go.

PR #821, opened 2023-08-18
Remove future from the cohorts code

Moohan:
I spoke to @Jennit07 about this, it's possibly causing an issue. I didn't have chance to get to the bottom of it but here's a draft PR in case you think it is causing an issue and want to try without the code!


Jennit07 (approved):
I've tested this and can confirm this code using future was causing an issue in targets. Ready to merge this back into sept-23 update

PR #825, opened 2023-09-06
Lizihao anu patch 1

lizihao-anu:


Jennit07 (approved):
Ill approve and merge this, i think i took out the chi variable in another PR but it makes sense to add back in

PR #833, opened 2023-09-26
transform the python script for sorting BI extracts to R

lizihao-anu:
I am sure this R script is absolutely super faster than the original Python script. The reason is Python is running on the local laptop. The specific steps are downloading the files from the server and then uploading the files to the new folder in the server. Instead, R moves files directly on the server. The time consumption is pretty much similar to that of what we do on the local laptop. 

In light of this, when we want to move files between folders on the server, the best way to do is to use Posit. There are two ways to achieve this. 

1. Use the R command `file.move()`, `file.copy`, etc. 
2. Use the Posit user interface to do so. In file table, select "More", and we can move, duplicate, or copy files on the server.

To be short, if we do something via Microsoft Explorer, it is highly likely that we are downloading and uploading files and draining the VPN. If we do something via Posit, it is done directly within the server, which is fast and less resource-consuming.

Feel free to make comments.

### DIFF
--- a/00_Sort_BI_Extracts.R
+++ b/00_Sort_BI_Extracts.R
@@ -1,5 +1,5 @@
 # Define the source directory and financial year pattern
-compress_files = FALSE
+compress_files <- FALSE
 source_dir <- "/conf/sourcedev/Source_Linkage_File_Updates/Extracts Temp"
 pattern <- "-20(\\d{4})\\.csv"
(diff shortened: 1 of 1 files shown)


Jennit07 (approved):
I tested this using an extract from boxi and it is working as expected. Happy to approve/merge!

Issue #688, opened 2023-06-12
Rename `convert_sending_location_to_lca` for clarity


Moohan:
This could be renamed to make it clear that sending location is from Social Care e.g. `convert_sc_sending_location_to_lca <- function(sc_sending_location)`

_Originally posted by @Moohan in https://github.com/Public-Health-Scotland/source-linkage-files/pull/638#discussion_r1220007876_

SwiftySalmon:
Jen sorted thi issue September 2023

PR #839, opened 2023-09-29
Rename function to `convert_sc_sending_location_to_lca`

Jennit07:
Addresses #688

I have set this to merge into Master so that we can then bring development up to date with master once this merges. The reason is to try and avoid multiple merge conflicts like the last time we renamed a function. Happy to help if this doesn't make sense!


SwiftySalmon (approved):
all looks good. no conflicts

PR #845, opened 2023-10-04
change boxi file names back to capitals

SwiftySalmon:
A previous pull request changed all capitals to lowercase - however boxi file names have capitals so it was no longer reading in files. This is a fix

### DIFF
--- a/R/get_boxi_extract_path.R
+++ b/R/get_boxi_extract_path.R
@@ -41,19 +41,19 @@ get_boxi_extract_path <- function(
 
   file_name <- dplyr::case_match(
     type,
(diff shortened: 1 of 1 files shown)


Jennit07:
I think this is fine as a quick fix for reading in the files. However, i think the files need renaming in sourcedev to lowercase then it should work. For consistency, the boxi extract function should be in lowercase as we do everything in R in lowercase. When this is applied, this will automatically save the processed files in lowercase the next time we update the files. 

The boxi extracts may need renaming to lowercase when we take these directly from boxi for this also to work which may be why you guys were having issues with it as the csv files in the `extract` folder will automatically have this capitalised when downloaded from boxi.

PR #847, opened 2023-10-18
Keep population

lizihao-anu:
A first draft of transform from SPSS to R for keep_population. Help is needed to check the logic is correct.


R/add_keep_population_flag.R
```diff
@@ -0,0 +1,163 @@
...
+    year_available <- pop_estimates %>%
+      dplyr::pull(year) %>%
+      unique()
+    if (calendar_year %in% year_available) {
+      pop_estimates <- pop_estimates %>%
+        dplyr::filter(year == calendar_year)
+    } else {
+      previous_year <- sort(year_available, decreasing = TRUE)[1]
+      pop_estimates <- pop_estimates %>%
+        dplyr::filter(year = previous_year)
```
Jennit07, line 34:
```suggestion
        dplyr::filter(year == previous_year)
```

R/add_keep_population_flag.R
```diff
@@ -0,0 +1,163 @@
+#' Add keep_popluation flag
+#'
+#' @description Add keep_population flag to individual files
+#' @param data A data frame
+#'
+#' @return A data frame with keep_population flags
+#' @family individual_file
+
+add_keep_population_flag <- function(individual_file, year) {
+  calendar_year <- paste0("20", substr(year, 1, 2)) %>% as.integer()
```
Jennit07, line 10: I wonder if we need this in a function or do we only use calendar year for population estimates?

R/add_keep_population_flag.R
```diff
@@ -0,0 +1,163 @@
...
+
+  if (!check_year_valid(year, "NSU")) {
+    individual_file <- individual_file %>%
+      dplyr::mutate(keep_population = 1L)
+  } else {
+    ## Obtain the population estimates for Locality AgeGroup and Gender.
+    pop_estimates <-
+      readRDS(get_datazone_pop_path("DataZone2011_pop_est_2011_2021.rds")) %>%
+      dplyr::as_tibble() %>%
```
Jennit07, line 19:
```suggestion
      readr::read_rds(get_datazone_pop_path("DataZone2011_pop_est_2011_2021.rds")) 
```
Not sure why we need `dplyr::as_tibble()` here? i can read in the file okay when running this line of code. I am suggesting to use `readr::read_rds` here instead as i think we do that everywhere else for consistency

R/add_keep_population_flag.R
Jennit07, line 34: This is good and works as expected. One typo i noticed

R/add_keep_population_flag.R
```diff
@@ -0,0 +1,163 @@
...
+          age >= 45 & age <= 54 ~ "45-54",
+          age >= 55 & age <= 64 ~ "55-64",
+          age >= 65 & age <= 74 ~ "65-74",
+          age >= 75 & age <= 84 ~ "75-84",
+          age >= 85 ~ "85+"
+        )
+      ) %>%
+      dplyr::left_join(
+        get_locality_path() %>%
+          readRDS() %>%
```
Jennit07, line 67: I think this could be re-written as..
```suggestion
        readr::read_rds(get_locality_path()) 
```

R/add_keep_population_flag.R
```diff
@@ -0,0 +1,163 @@
...
+    if (calendar_year %in% year_available) {
+      pop_estimates <- pop_estimates %>%
+        dplyr::filter(year == calendar_year)
+    } else {
+      previous_year <- sort(year_available, decreasing = TRUE)[1]
+      pop_estimates <- pop_estimates %>%
+        dplyr::filter(year = previous_year)
+    }
+
+    pop_estimates <- pop_estimates %>%
```
Jennit07, line 37:
```suggestion
    pop_estimates_filtered <- pop_estimates %>%
```

R/add_keep_population_flag.R
```diff
@@ -0,0 +1,163 @@
...
+      columns = c(
+        "chi",
+        "locality",
+        "age",
+        "gender",
+        # "nsu",
+        "death_date"
+      )
+    ) %>%
+      dplyr::mutate(nsu = 0L) # delete this before merge
```
Jennit07, line 87: Here you are reading the individual file from hscdiip when this should be the current individual file at the point of processing. This used to be called `temp4` but you have set the data parameter as `individual_file` which you can use instead as we want the estimates from the new file we are creating
```suggestion
    individual_file <- individual_file 
```

R/add_keep_population_flag.R
```diff
@@ -0,0 +1,163 @@
...
+          age >= 0 & age <= 4 ~ "0-4",
+          age >= 5 & age <= 14 ~ "5-14",
+          age >= 15 & age <= 24 ~ "15-24",
+          age >= 25 & age <= 34 ~ "25-34",
+          age >= 35 & age <= 44 ~ "35-44",
+          age >= 45 & age <= 54 ~ "45-54",
+          age >= 55 & age <= 64 ~ "55-64",
+          age >= 65 & age <= 74 ~ "65-74",
+          age >= 75 & age <= 84 ~ "75-84",
+          age >= 85 ~ "85+"
```
Jennit07, line 102: I think this section could be put into a new function as this was also done above in lines 53-62

R/add_keep_population_flag.R
```diff
@@ -0,0 +1,163 @@
...
+        )
+      )
+
+    set.seed(100)
+    mid_year <- lubridate::dmy(stringr::str_glue("30-06-{calendar_year}"))
```
Jennit07, line 107: please see the function called `midpoint_fy` i think we could use that instead and remove this line

R/add_keep_population_flag.R
Jennit07, line 87: Why do we set nsu to 0 here?

R/add_keep_population_flag.R
```diff
@@ -0,0 +1,163 @@
...
+    ## issues with age being negative
+    # If they don't have a locality, they're no good as we won't have an estimate to match them against.
+    # Same for age and gender.
+    nsu_keep_lookup <- individual_file_1 %>%
```
Jennit07, line 111: I got a bit lost from here onwards as the nsu variable only had 0 for each row. We can have chat about this on tuesday

R/add_keep_population_flag.R
lizihao-anu: midpoint_fy gives the mid of a financial year but this is the mid of a calendar year. So not equivalent.

lizihao-anu:
Had a test run and showed the code works fine.

R/add_keep_population_flag.R
```diff
@@ -0,0 +1,146 @@
...
+      ) %>%
+      dplyr::group_by(locality, age_group, gender) %>%
+      dplyr::summarize(population_estimate = sum(population_estimate)) %>%
+      dplyr::ungroup()
+
+    # Step 2: Work out the current population sizes in the SLF for Locality, AgeGroup, and Gender
+    # Work out the current population sizes in the SLF for Locality AgeGroup and Gender.
+    individual_file_1 <- individual_file %>%
+      dplyr::mutate(age = as.integer(age)) %>%
+      add_age_group("age")
```
Jennit07, line 64: Im a bit confused whats happening here with the naming of objects and matching them back on to things. I am going to make some suggestions to simplify the code, could you confirm if these are doing as you expect? thanks

R/add_keep_population_flag.R
```diff
@@ -0,0 +1,146 @@
...
+          dplyr::select("locality" = "hscp_locality", datazone2011),
+        by = "datazone2011"
+      ) %>%
+      dplyr::group_by(locality, age_group, gender) %>%
+      dplyr::summarize(population_estimate = sum(population_estimate)) %>%
+      dplyr::ungroup()
+
+    # Step 2: Work out the current population sizes in the SLF for Locality, AgeGroup, and Gender
+    # Work out the current population sizes in the SLF for Locality AgeGroup and Gender.
+    individual_file_1 <- individual_file %>%
```
Jennit07, line 62:
```suggestion
    individual_file <- individual_file %>%
```

R/add_keep_population_flag.R
```diff
@@ -0,0 +1,146 @@
...
+      dplyr::mutate(age = as.integer(age)) %>%
+      add_age_group("age")
+
+
+    set.seed(100)
+    mid_year <- lubridate::dmy(stringr::str_glue("30-06-{calendar_year}"))
+    ## issues with age being negative
+    # If they don't have a locality, they're no good as we won't have an estimate to match them against.
+    # Same for age and gender.
+    nsu_keep_lookup <- individual_file_1 %>%
```
Jennit07, line 72:
```suggestion
    nsu_keep_lookup <- individual_file %>%
```

R/add_keep_population_flag.R
```diff
@@ -0,0 +1,146 @@
...
+          scaling_factor > 1 ~ 1,
+          .default = scaling_factor
+        ),
+        keep_nsu = rbinom(1, 1, scaling_factor)
+      ) %>%
+      dplyr::filter(keep_nsu == 1L) %>%
+      dplyr::ungroup()
+
+    # step 3: match the flag back onto the slf
+    individual_file <- individual_file_1 %>%
```
Jennit07, line 101:
```suggestion
    individual_file <- individual_file %>%
```

R/add_keep_population_flag.R
Jennit07: I think we can just remove `individual_file_1` and this will still work as intended but simplifying the dataframes. From my understanding the only difference between `individual_file_1` and `individual_file` is that we want to have a column for the age groups? Please let me know your thoughts on this and if i have misunderstood. Thanks

R/add_keep_population_flag.R
```diff
@@ -0,0 +1,146 @@
+#' Add keep_popluation flag
+#'
+#' @description Add keep_population flag to individual files
+#' @param data A data frame
+#'
+#' @return A data frame with keep_population flags
+#' @family individual_file
+
+add_keep_population_flag <- function(individual_file, year, chi_var_name = "chi") {
```
Jennit07, line 9:
```suggestion
add_keep_population_flag <- function(individual_file, year") {
```
I dont think we need to specify the chi variable here as this should still be `chi` at this stage in the processing. However i know why you may have added this here if you were testing an existing individual file.

R/add_keep_population_flag.R
```diff
@@ -0,0 +1,146 @@
+#' Add keep_popluation flag
+#'
+#' @description Add keep_population flag to individual files
+#' @param data A data frame
+#'
```
Jennit07, line 5: Could you update the documentation please and add individual_file and year as parameters. Thanks

R/add_keep_population_flag.R
```diff
@@ -0,0 +1,146 @@
...
+        {{ age_var_name }} >= 35 & {{ age_var_name }} <= 44 ~ "35-44",
+        {{ age_var_name }} >= 45 & {{ age_var_name }} <= 54 ~ "45-54",
+        {{ age_var_name }} >= 55 & {{ age_var_name }} <= 64 ~ "55-64",
+        {{ age_var_name }} >= 65 & {{ age_var_name }} <= 74 ~ "65-74",
+        {{ age_var_name }} >= 75 & {{ age_var_name }} <= 84 ~ "75-84",
+        {{ age_var_name }} >= 85 ~ "85+"
+      )
+    )
+  return(individual_file)
+}
```
Jennit07, line 146: Looks good, i think you just need to add documentation for this function

R/add_keep_population_flag.R
lizihao-anu: Sure

SwiftySalmon (commented):
logic all seems good to me. I can't find the original spss script to compare it with though!

Jennit07 (approved):
Tested and ready to merge. If we find any bugs identified by the consultancy team then we can fix them during the update

PR #849, opened 2023-10-25
Make improvements to test workbooks

Jennit07:
Opening as a draft PR for now to see changes. I have tested the episode file and happy it works as expected but having trouble with the individual file tests. I would like to test these before marking ready for review.

Test workbooks to check:
- [x] episode file
- [x] individual file
- [x] extracts
- [x] lookups


Jennit07:
Okay, i've tested the workbooks and think im happy with the new set up. I've identified a few bugs along the way and fixed them on this branch. An overview of this:

Improvements: 
- new workbooks for sets of tests: 
- episode file tests workbook
- individual file tests workbook
- extract tests workbook
- lookup tests workbook

Bugs:
- Changed tests to use anon_chi when testing the episode/individual files. It was taking too long to test for chi using SLFhelper (this has now been fixed and PR waiting for review: https://github.com/Public-Health-Scotland/slfhelper/pull/68) Although slfhelper has been fixed i have left this to deal with anon_chi for faster processing! 
- Fixed a bug in `get_source_extract_path` which said it couldnt read the file in sourcedev. This was because year was in the format `1920` instead of `201920` in the file name. 
- A bug i thought we had fixed before was dealing with the variable `hscp` in tests. Basically, in the extract tests this looks for `hscp` but in the episode/individual file tests this looks for `hscp2018` so it throwing an error. There was a fix in the function `get_existing_data_for_tests` but this would break either the extract tests or episode tests depending on which one you run. I have implemented a fix where this will rename the extracts to `hscp2018` which will also match the episode/individual file tests.

R/process_tests_individual_file.R
```diff
@@ -61,11 +60,19 @@ produce_individual_file_tests <- function(data) {
 
   test_flags <- data %>%
     # use functions to create HB and partnership flags
-    create_demog_test_flags() %>%
+    dplyr::mutate(
```
SwiftySalmon, line 63: are these tests different to the create_demog_test_flags?

SwiftySalmon (commented):
Looks good to me! Will be much easier having tests split out that way. 
What's going on with the hscp2018 changes - seems like a lot of repitition?

R/process_tests_acute.R
```diff
@@ -12,11 +12,18 @@
 process_tests_acute <- function(data, year) {
   old_data <- get_existing_data_for_tests(data)
 
+  if ("hscp" %in% names(data)) {
+    data <- data %>%
+      dplyr::rename("hscp2018" = "hscp")
+  } else {
+    data <- data
+  }
+
```
Jennit07, line 21: put this into a function

Jennit07:
Changes made to put rename hscp into a function

SwiftySalmon (approved):

Issue #842, opened 2023-09-29
Create tests for each 'all' sc extracts


Jennit07:
Looking at this issue, i have discovered we already have `create_sending_location_test_flags` in the 'year' social care tests. However this is called `create_lca_test_flags` instead. The reason for this is because we create the `sending_lca` variable using `sending_location` at this stage in the 'processed' extracts. This function provides a count for each LCA area.  

We need to keep `create_sending_location_test_flags` as the client lookup does not have the `lca` variable so i think it is useful to keep both functions. 

I have also identified by looking at this issue that there are no tests for the 'all' sc files. I will make another issue to have a look at this again. However, on the branch i created i could apply `create_sending_location_test_flags` so i will push this and connect to this issue!

_Originally posted by @Jennit07 in https://github.com/Public-Health-Scotland/source-linkage-files/issues/818#issuecomment-1740815685_

Tests for each sc 'all' extract: 

- [x] All Alarms Telecare
- [x] All Care Home
- [x] All Home Care 
- [x] All SDS

Jennit07:
completed in #854

PR #854, opened 2023-11-06
Create tests for `all` social care extracts

Jennit07:
addresses #842 

Produces tests for social care `all` extracts. The data is the processed extract taken from the social care platform containing all years worth of data. 

Still to do: 
- Add these tests into the targets pipeline. (@SwiftySalmon happy to have a call with you to share how to do this!)


Jennit07:
JT - still to test HC and SDS. Also need to have a call to put this into the targets pipeline.

Jennit07:
@SwiftySalmon This one is ready to check/merge

SwiftySalmon (approved):

PR #868, opened 2023-11-22
Social Care demographics chi validation

SwiftySalmon:
Do chi and dob match and if not it looks to see if either chi or dob match. There are a lot of cases where it still doesn't match and the issue of multiple chis per social care id isn't fixed yet - social care are now aware of it though - but this is potentially the best we can do for the moment


R/process_lookup_sc_demographics.R
```diff
@@ -29,29 +29,135 @@ process_lookup_sc_demographics <- function(
...
+      # if only one option is available for DOB then choose that
+      dob = ifelse(chi_date_of_birth == submitted_date_of_birth | is.na(submitted_date_of_birth), chi_date_of_birth,
+        ifelse(is.na(chi_date_of_birth), submitted_date_of_birth, NA)
+      ),
+      dob = lubridate::as_date(dob)
+    ) %>%
+    dplyr::arrange(chi, dob) %>%
+    dplyr::group_by(social_care_id, sending_location) %>%
+    tidyr::fill(chi, .direction = c("down")) %>%
+    tidyr::fill(dob, .direction = c("down")) %>%
```
rchlv, line 74: This work looks good, well done! I don’t fully understand the data and fill function yet, but is there a chance this could fill in a different chi or dob in a blank field if the row above belongs to a different person?

SwiftySalmon: So this is grouping by social_care_id and sending_location and only filling in chi and dob if one exists for that social_care_id and sending_location combination. 
This isn't great for combinations with two chis attached to them...but my thinking is that it will fix more cases than it will break...? This is another problem for another day.

rchlv: That makes sense, thank you!

PR #872, opened 2023-11-30
Update SPARRA year to pick up new 2324 file

Jennit07:
Merge this PR when we have the new SPARRA file available


rchlv (approved):
Both Jen and Rachel have checked this.

PR #874, opened 2023-12-04
fixed bugs in the keep population flag.

SwiftySalmon:
- filter out where gender not 1 or 2
- deal with negative ages where dob is after mid point of year
- previously filtered cases where dod was before midpoint, but was also accidentally removing cases with missing death dates! Now fixed.
- added filter for nsu only once calculations are done

### DIFF
--- a/R/add_keep_population_flag.R
+++ b/R/add_keep_population_flag.R
@@ -102,7 +102,7 @@ add_keep_population_flag <- function(individual_file, year) {
     # step 3: match the flag back onto the slf
     individual_file <- individual_file %>%
       dplyr::left_join(nsu_keep_lookup,
-        by = "anon_chi",
+        by = "chi",
         suffix = c("", ".y")
       ) %>%
       dplyr::select(-contains(".y")) %>%
--- a/R/add_keep_population_flag.R
+++ b/R/add_keep_population_flag.R
@@ -24,6 +24,7 @@ add_keep_population_flag <- function(individual_file, year) {
     year_available <- pop_estimates %>%
       dplyr::pull(year) %>%
       unique()
+
     if (calendar_year %in% year_available) {
       pop_estimates <- pop_estimates %>%
         dplyr::filter(year == calendar_year)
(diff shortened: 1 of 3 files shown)


lizihao-anu (dismissed):
Thanks for the fixing

R/add_keep_population_flag.R
```diff
@@ -137,7 +139,7 @@ add_age_group <- function(data, age_var_name) {
   data <- data %>%
     dplyr::mutate(
       age_group = dplyr::case_when(
-        {{ age_var_name }} >= 0 & {{ age_var_name }} <= 4 ~ "0-4",
+        {{ age_var_name }} >= -1 & {{ age_var_name }} <= 4 ~ "0-4",
```
rchlv, line 142: Thanks for figuring out this issue, well done!
For age, do we sometimes get people with age -1?

SwiftySalmon: I think it's due to age being calculated at the midpoint of the year. So if someone is born in September their age on 30th June is -1. Very fun for us.

R/add_keep_population_flag.R
```diff
@@ -70,17 +71,18 @@ add_keep_population_flag <- function(individual_file, year) {
...
-      # anyone who died 5 years before the file shouldn't be in it anyway...
-      dplyr::filter(death_date > mid_year | nsu != 0) %>%
+      dplyr::filter(death_date > mid_year | is.na(death_date) | nsu != 0) %>%
       # Calculate the populations of the whole SLF and of the NSU.
       dplyr::group_by(locality, age_group, gender) %>%
       dplyr::mutate(
         nsu_population = sum(nsu),
         total_source_population = dplyr::n()
       ) %>%
+      dplyr::filter(nsu == 1) %>%
```
Jennit07, line 85: This was a good spot here! I've checked SPSS and this matches up. I think im happy to approve this now!

Jennit07 (approved):

PR #878, opened 2023-12-07
Bug - keep population filter and nsu pop not working

Jennit07:
We had a call to discuss the filter for deaths at the mid point of calendar year. This includes the changes we discussed. There was also the issue of the function `rbinom` not working correctly and selecting out whole groups rather than random samples from each group. This now works. I will run the code tomorrow to produce some of the tables but Rachel and I checked the keep population flag counts for 1819 and it seems a lot more reasonable with the figures. Great work everyone! Thanks


SwiftySalmon (approved):
Looks much better now!

PR #881, opened 2023-12-12
Fix duplicate episode sc send lca

lizihao-anu:
See the create_episode_file, 

line 34, `episode_file <- dplyr::bind_rows(processed_data_list) %>%`, the `bind_rows` matched all episode records by chi, sending_location, sc_send_lca, etc. 
line 37, `store_ep_file_vars(` drops not very necessary columns/variables, including `sc_send_lca`, and will join back at line 137.
line 136, `join_sc_client(year, sc_client = sc_client, file_type = "episode") %>%` add sc_send_lca column again, which cause two sc_send_lca columns.

Since line 136 only matches by sending_location, it seems not as reliable as line 34. I prefer to drop `sc_send_lca` column added by line 136. Feel free to comment.

### DIFF
--- a/R/create_episode_file.R
+++ b/R/create_episode_file.R
@@ -438,7 +438,7 @@ join_sc_client <- function(data,
     # Match on client variables by chi
     data_file <- data %>%
       dplyr::left_join(
-        sc_client %>% dplyr::select(-sc_send_lca),
+        sc_client,
         by = "chi",
         relationship = "many-to-one"
       )
--- a/R/create_episode_file.R
+++ b/R/create_episode_file.R
@@ -435,7 +435,8 @@ join_sc_client <- function(data,
   if (file_type == "episode") {
     # Match on client variables by chi
     data_file <- data %>%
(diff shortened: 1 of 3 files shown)


R/create_episode_file.R
```diff
@@ -437,7 +437,8 @@ join_sc_client <- function(data,
   if (file_type == "episode") {
     # Match on client variables by chi
     data_file <- data %>%
-      dplyr::left_join(sc_client,
+      dplyr::left_join(
+        sc_client %>% dplyr::select(-sc_send_lca),
```
Jennit07, line 441: Great work finding the problem here. I have a suggestion that we could remove this from `process_lookup_sc_client` instead as this is the source. Other social care scripts create this which is why it appears in the bind rows. It would be lines 153-155 in the function `process_lookup_sc_client`. I think this would achieve the same as your suggestion here as in client processing we add this variable on but here we are just removing it. Any thoughts?

Jennit07 (approved):

PR #892, opened 2024-01-12
Fix targets

lizihao-anu:
## Issues fixed so far:

1. In tests_sc_client_lookup, the variable `sc_send_lca` not found.
2. Fix the legacy of the replacement of get_datazone_pop_path by get_pop_path. Some scripts were not completely changed.
3. Fix the issue that the variable year is not explicitly specified in functions as an argument. This bug is ridiculously difficult to find and identify because scripts with this sort of bug run perfectly without `targets`. The logic of this bug is as follows.
When using targets, arguments/constants in a function need to be specified explicitly. However, baseR does not need to explicitly specify the arguments/constants in a function. For example,
```
a = 1
f = function(x){return(x+a)}
f(1)
# 2
```
It works fine. But in `targets` it does not. A demo targets script is as follows.
```
library(targets)
f = function(x) {
  return(x + a)
}

list(
  list(a = 1),
  tar_target(y, f(1))
)
```
When we run the command `tar_make()`, an error is reported.
```
> tar_make()
▶ start target y
✖ error target y
▶ end pipeline [0.958 seconds]
Error:
! Error running targets::tar_make()
  Error messages: targets::tar_meta(fields = error, complete_only = TRUE)
  Debugging guide: https://books.ropensci.org/targets/debugging.html
  How to ask for help: https://books.ropensci.org/targets/help.html
  Last error: object 'a' not found
```

## Remaining issue:
The remaining issue is the limited RAM (128GiB = 131072 MiB). Running the function `create_individual_file` without targets is perfectly fine but it requires more RAM to get it running in `targets`. Otherwise, an error is reported as below.
```
> tar_make_future(
+   names = (targets::contains("1920"))
+ )
▶ +0000 UTC 2024-01-17 19:18 17.38 start target compress_extracts_1920
● +0000 UTC 2024-01-17 19:18 44.90 built target compress_extracts_1920 [0.513 seconds]
▶ +0000 UTC 2024-01-17 20:47 12.58 start target individual_file_1920
Error in `get_result(output = out, options)`:
! callr subprocess failed: could not start R, exited with non-zero status, has crashed or was killed
ℹ See `$stderr` for standard error.
Type .Last.error to see the more details.
```
I have tried the function `create_individual_file` with `targets` in a session with 240GiB RAM, and it worked perfectly. Hence, it confirms the insufficient RAM. There are two temporary and quick solutions to this issue.

- Running the current targets script within a session with 240GiB RAM; or
- Comment out the bit of creating individual files and their tests, and append the process to `run_targets_{year}.R` after `tar_make()`.

Feel free to comment. (This comment is updated on 18/01/2024)


lizihao-anu:
This is ready for review.

SwiftySalmon:
Looks good! Not sure how much I understand about the targets stuff but I'll take another look next week. 
Is the plan to add the individual file code back into targets? or will we run that manually now?

R/aggregate_by_chi.R
```diff
@@ -7,7 +7,7 @@
 #' @importFrom data.table .SD
 #'
 #' @inheritParams create_individual_file
-aggregate_by_chi <- function(episode_file, exclude_sc_var = FALSE) {
+aggregate_by_chi <- function(episode_file, year, exclude_sc_var = FALSE) {
```
Jennit07, line 10: Good spot!

R/create_individual_file.R
```diff
@@ -82,7 +82,7 @@ create_individual_file <- function(
     individual_file <- individual_file %>%
       aggregate_ch_episodes() %>%
       clean_up_ch(year) %>%
-      aggregate_by_chi(exclude_sc_var = FALSE)
+      aggregate_by_chi(year = year, exclude_sc_var = FALSE)
```
Jennit07, line 85: Good spot!

R/process_lookup_sc_client.R
```diff
@@ -150,6 +150,9 @@ process_lookup_sc_client <-
       )) == "Not Known")) %>%
       dplyr::arrange(chi, count_not_known) %>%
       dplyr::distinct(chi, .keep_all = TRUE) %>%
+      dplyr::mutate(
+        sc_send_lca = convert_sc_sending_location_to_lca(sending_location)
+      ) %>%
```
Jennit07, line 155: This can be removed as this will introduce duplicate code. We previously fixed this in this PR https://github.com/Public-Health-Scotland/source-linkage-files/pull/881/files

R/create_episode_file.R
```diff
@@ -31,6 +31,9 @@ create_episode_file <- function(
     sc_client = read_file(get_sc_client_lookup_path(year)),
     write_to_disk = TRUE,
     anon_chi_out = TRUE) {
+  sc_client <- sc_client %>%
+    dplyr::select(-sc_send_lca)
+
```
Jennit07, line 36: These lines are not needed here and will cause duplicates. See https://github.com/Public-Health-Scotland/source-linkage-files/pull/881/files

run_targets_1718.R
```diff
@@ -0,0 +1,20 @@
...
+)
+
+# use targets to create individual files due to RAM limit
+library(createslf)
+
+episode_file <- arrow::read_parquet(get_slf_episode_path(year))
+
+# Run individual file
+create_individual_file(episode_file, year = year) %>%
+  process_tests_individual_file(year = year)
```
Jennit07, line 20: I think overall this PR is ready to merge. However i have some questions about the set up here. When you are back from A/L can we have a call to discuss? 

I wonder if we need to have separate set up scripts for the individual file if this runs out of memory during the processing? 

There are also workbench job scripts in the folder Run_SLF_Files_targets. so i think these would need to go into that folder to replace each `year` script.

Jennit07 (changes requested):
PR almost ready. I just have a few questions to ask zihao when he is back from A/L. Thanks!

SwiftySalmon (dismissed):

Jennit07 (approved):

PR #894, opened 2024-01-16
Remove redundant code

### LINKED ISSUE #603: remove this bit after SPSS stopped
replace NA by 0 in monthly costs

https://api.github.com/Public-Health-Scotland/source-linkage-files/blob/6eed5946ac5a61e5a3f2f8cb45779e05bdfe5974/R/process_tests_district_nursing.R#L12

```r

#' @export
process_tests_district_nursing <- function(data, year) {
  old_data <- get_existing_data_for_tests(data) %>%
    # TODO: remove this bit after SPSS stopped
    # replace NA by 0 in monthly costs
    dplyr::mutate(dplyr::across(
      dplyr::ends_with("_cost"),

```

Jennit07:
closes #603


R/process_tests_district_nursing.R
```diff
@@ -13,14 +13,7 @@ process_tests_district_nursing <- function(data, year) {
...
   }
 
-  old_data <- get_existing_data_for_tests(data) %>%
-    # TODO: remove this bit after SPSS stopped
-    # replace NA by 0 in monthly costs
-    dplyr::mutate(dplyr::across(
-      dplyr::ends_with("_cost"),
-      ~ tidyr::replace_na(.x, 0.0)
-    ))
-
```
lizihao-anu, line 23: Are we not using SPSS results any more? If so, we can remove this. I remember this bit of code is to fix the NA values in SPSS results, which will lead to abnormally higher mean in comparsions. Because the average of c(1,NA) is 1 rather than 0.5 as we expect from c(1, 0.5).

SwiftySalmon (approved):
okay

PR #896, opened 2024-01-22
Bug - match on client variables before NSU

Jennit07:
@SwiftySalmon spotted an issue where NSUs were showing a social care id attached. This was due to us matching on NSUs first and then SC Client variables at the end. Previously we matched SC client variables first and then NSUs onto the episode file. This difference occurred when we split up matching on client variables at the year specific extracts. This is a quick PR to reorder the functions to match our previous logic. I have tested the 1920 file to check this and the issue is no longer there.


SwiftySalmon (approved):
Looks good! Change makes sense

SwiftySalmon (approved):

PR #898, opened 2024-01-22
Review `create_demog_test_flags` function 

### LINKED ISSUE #850: Update `create_demog_test_flags` function
In the tests, i think this function would be useful to have a chi parameter for specifying use of chi or anon_chi. It needs changes elsewhere so opening an issue for now and will apply a fix for what i was working on.

Jennit07:
Improved the function `create_demog_test_flags` function to include a parameter for specifying `chi` or `anon_chi`. This is due to this function being used for testing the extracts (which calls chi) and also being used for testing the episode and individual file (which calls anon_chi). This now allows flexibility for interchanging between the two forms and simplifies the code by removing duplicated code.

Closes #850


SwiftySalmon (dismissed):

SwiftySalmon (approved):
didn't test but looks fine

Issue #758, opened 2023-07-26
Investigate extracts to identify areas of code which can be cut down for processing times


Jennit07:
Ones which stand out to me as worthy of some investigation to potentially speed up:

- [ ] Social Care Demographics
- [ ] Social Care Client
- [ ] Social Care Care Home
- [ ] Social Care Home Care
- [ ] Social Care SDS
- [ ] Social Care A&T
- [ ] District Nursing
- [ ] Prescribing

_Originally posted by @Moohan in https://github.com/Public-Health-Scotland/source-linkage-files/issues/549#issuecomment-1495650764_

lizihao-anu:
`targets` with `future` in Posit running time for each item:

- sc_demog_lookup [2.972 minutes]
- sc_client_lookup_1920 [1.614 minutes]
- all_care_home [11.471 minutes]
- all_care_home_extract [1.225 minutes]
- source_sc_care_home_1920 [6.794 seconds]
- all_home_care_extract [2.754 minutes]
- all_home_care [6.132 minutes]
- source_sc_home_care_1920 [5.734 seconds]
- all_sds_extract [50.751 seconds]
- all_sds [33.984 minutes]
- source_sc_sds_1920 [3.235 seconds]
- all_at [34.543 minutes]
- source_dn_extract_1920 [3.112 minutes]
- prescribing_data_file_1920 [1.198 seconds]
- prescribing_data_1920 [2.01 minutes]
- source_prescribing_extract_1920 [1.014 minutes]

So, `all_sds` and `all_at` will need investigation most. I will check if we could improve the efficiency.

Jennit07:
Further investigation is required when R 4.4 is available.

PR #899, opened 2024-01-22
758 investigate extracts to identify areas of code which can be cut down for processing times

lizihao-anu:
As discussed in #758, two functions, `process_sc_all_sds` and `process_sc_all_alarms_telecare`, take a very long time to execute. I have re-written the functions using `data.table` to improve the efficiency, including the speed and memory usage. The improvement is ridiculously significant.

Here's a summary comparing the improvements in two functions, `process_sc_all_sds` and `process_sc_all_alarms_telecare`:

| Function Name              | Old Method    | New Method    | Improvement   |
|---------------------------|---------------|---------------|---------------|
| `process_sc_all_sds`      | 24.94 mins | 1.69 mins | 94% reduction |
| `process_sc_all_alarms_telecare` | 31.81 mins | 0.81 mins | 98% reduction |

In both cases, the new methods produce identical results to the old methods but demonstrate significant reductions in running time. The `process_sc_all_sds` function saw a 94% reduction in running time, while the `process_sc_all_alarms_telecare` function experienced an even more impressive 98% reduction. These improvements are a testament to the effectiveness of the rewritten functions in optimizing their performance.


R/process_sc_all_sds.R
```diff
@@ -15,123 +15,152 @@ process_sc_all_sds <- function(
     sc_demog_lookup,
     write_to_disk = TRUE) {
   # Match on demographics data (chi, gender, dob and postcode)
-  matched_sds_data <- data %>%
+  matched_sds_data <- all_sds_extract %>%
```
SwiftySalmon, line 18: need to change this back to data

SwiftySalmon (changes requested):
Ran the sds code and it's much much faster. 

Only issue is that I've made a lot of changes to the sds code! Might need to merge my pull request and then make some changes to this one?

lizihao-anu:
> Ran the sds code and it's much much faster.
> 
> Only issue is that I've made a lot of changes to the sds code! Might need to merge my pull request and then make some changes to this one?

Yes. I agree. We can merge your changes to sds first, and the I can rewrite the code for sds.

R/process_sc_all_sds.R
```diff
@@ -15,123 +15,152 @@ process_sc_all_sds <- function(
...
-        sds_option == "SDS-4" ~ "SDS-4"
-      ),
-      recid = "SDS",
-      # Create person id variable
-      person_id = stringr::str_glue("{sending_location}-{social_care_id}"),
-      # Use function for creating sc send lca variables
-      sc_send_lca = convert_sc_sending_location_to_lca(.data$sending_location)
+      sds_option_1 = tidyr::replace_na(sds_option_1, 0),
+      sds_option_2 = tidyr::replace_na(sds_option_2, 0),
+      sds_option_3 = tidyr::replace_na(sds_option_3, 0)
```
Jennit07, line 30: Im wondering about this part of code as the variable `smrtype` codes the 4 options for SDS. If we are splitting them out into their own variables i think we would need to let people know in the email and also update the episode file layout workbook. Has this been split out due to how data table works?

Im not too familiar with working with code in data table. Do you have any links to any documentation to help my understanding? thanks

Jennit07: checked the final output and these variables are not there so i think it is fine. I think it is how i am reading the code!

Jennit07:
Hi @lizihao-anu PR #900 has now merged. When you are back from A/L would you be able to have a look at the SDS code again and make any changes to have this in data table code please? Thanks!

lizihao-anu:
Ready to merge

Jennit07 (approved):

PR #900, opened 2024-01-23
SC Demographics and SDS

SwiftySalmon:
Complete rewrite of SC Demographics. 

Changes include:
- only using chi from chi, 
- filling missing chi in if possible,
- `latest submissing`  is no longer accurate, so added code to choose the latest and made changes to `replace_sc_id_with_latest` function
- now selects latest sc id across all social care sections, not just whatever section is being run. will be consistent across sds, ch, hc, and at (hopefully) 
- cases where chi is missing were previously assigned the same social care id. This is fixed so it's a unique social care id allowing analysis to be done with these cases if needed. 

Things not fixed: 
- multiple chis per social care id. Data Management aware and working on it
- cases where start date is after period end date get removed in SDS (n = 4) These one are up to DM to contact local authority and as to resubmit 

I checked sds output against the social care publication and I am happy with it. Obviously demographics code might change again once I look at other sections.


R/process_lookup_sc_demographics.R
```diff
@@ -28,30 +28,45 @@ process_lookup_sc_demographics <- function(
...
+    dplyr::arrange(period, social_care_id) %>%
+    dplyr::group_by(social_care_id, sending_location) %>%
+    tidyr::fill(chi, .direction = ("updown")) %>%
+    tidyr::fill(dob, .direction = ("updown")) %>%
+    tidyr::fill(date_of_death, .direction = ("updown")) %>%
+    tidyr::fill(gender, .direction = ("updown")) %>%
+    tidyr::fill(chi_postcode, .direction = ("updown")) %>%
+    tidyr::fill(submitted_postcode, .direction = ("updown")) %>%
+    # format postcodes using `phsmethods`
+    dplyr::mutate(dplyr::across(tidyselect::contains("postcode"), ~ phsmethods::format_postcode(.x, format = "pc7"))) # are sc postcodes even used anywhere?
```
Jennit07, line 48: I think you need a `dplyr::ungroup()` here. Either before or after you do the formatting on postcodes

Jennit07: Also not sure about sc postcodes. But i guess it makes sense to format to `pc7` for consistency! id need to check what we do for the postcode lookup. But im sure`pc7` is the standard format

R/process_lookup_sc_demographics.R
```diff
@@ -28,30 +28,45 @@ process_lookup_sc_demographics <- function(
...
-      # use CHI sex if available
-      gender = dplyr::if_else(
-        is.na(.data$chi_gender_code) | .data$chi_gender_code == 9L,
-        .data$submitted_gender,
-        .data$chi_gender_code
-      ),
-      # Use CHI DoB if available
-      dob = dplyr::coalesce(.data$chi_date_of_birth, .data$submitted_date_of_birth)
+      latest_flag = ifelse((latest == period & last_sc_id == social_care_id) | is.na(chi), 1, 0),
+      keep = ifelse(latest_sc_id == period, 1, 0)
```
Jennit07, line 61: We have a function called `replace_sc_id_with_the_latest` and the logic uses the latest flag from the platform which we know is not accurate. Do we know how this would affect this function? It is applied to all sc datasets. Would this function also help here as we are flagging the latest sc id?

R/process_lookup_sc_demographics.R
```diff
@@ -102,26 +120,32 @@ process_lookup_sc_demographics <- function(
...
+      .data$latest_flag
     ) %>%
     # summarise to select the last (non NA) submission
     dplyr::summarise(
-      chi = dplyr::last(.data$upi),
       gender = dplyr::last(.data$gender),
       dob = dplyr::last(.data$dob),
-      postcode = dplyr::last(.data$postcode)
+      postcode = dplyr::last(.data$postcode),
+      date_of_death = dplyr::last(.data$date_of_death)
```
Jennit07, line 140: Do we know how accurate date of death is from the sc platform? I wonder if we need to do a check with the nrs and chi deaths extract?

R/process_lookup_sc_demographics.R
Jennit07: Didnt realise you had already made changes to this further down!

R/process_sc_all_sds.R
```diff
@@ -15,14 +15,17 @@ process_sc_all_sds <- function(
     sc_demog_lookup,
     write_to_disk = TRUE) {
   # Match on demographics data (chi, gender, dob and postcode)
-  matched_sds_data <- data %>%
-    dplyr::left_join(
+  matched_sds_data <- data %>% #
+    dplyr::filter(.data$sds_start_date_after_period_end_date != 1) %>%
```
Jennit07, line 19: agree this is better/makes more sense here

R/process_sc_all_sds.R
```diff
@@ -15,14 +15,17 @@ process_sc_all_sds <- function(
...
+    dplyr::right_join(
       sc_demog_lookup,
       by = c("sending_location", "social_care_id")
     ) %>%
     # when multiple social_care_id from sending_location for single CHI
     # replace social_care_id with latest
-    replace_sc_id_with_latest()
+    replace_sc_id_with_latest() %>%
+    dplyr::select(-latest_sc_id, -latest_flag, -sds_start_date_after_period_end_date) %>%
+    dplyr::distinct()
```
Jennit07, line 28: Do we need a distinct here if we have just matched the demographics onto SDS data?

R/process_sc_all_sds.R
```diff
@@ -59,14 +62,19 @@ process_sc_all_sds <- function(
       sds_end_date = fix_sc_end_dates(
         .data$sds_start_date,
         .data$sds_end_date,
-        .data$period
+        .data$sds_period_end_date
```
Jennit07, line 65: agree

R/process_sc_all_sds.R
```diff
@@ -59,14 +62,19 @@ process_sc_all_sds <- function(
...
+    dplyr::select(
+      -sds_period_start_date, -sds_period_end_date,
+      -sds_start_date_after_end_date
+    ) %>%
     # rename for matching source variables
     dplyr::rename(
       record_keydate1 = .data$sds_start_date,
       record_keydate2 = .data$sds_end_date
     ) %>%
+    dplyr::distinct() %>%
```
Jennit07, line 77: Im confused here why we need distinct again as we do this on line 88

Jennit07:
Made a few comments, I was confused in some points of the code so just made a few comments as i went along. Posit was really slow for me today but i managed to run through the code and it was all working

R/process_lookup_sc_demographics.R
SwiftySalmon: added in. still need to push

R/process_sc_all_sds.R
SwiftySalmon: I think I did this in case there are cases that are duplicate but with two different sc_ids that are the same chi. So once sc_id is changed to the latest it will remove duplicates. I don't think there are any... but just in case there ever are?

R/process_sc_all_sds.R
SwiftySalmon: I just love distinct. If I've done a lot of mutates I tend to do a distinct to get rid of any duplicates that mutate might have introduced. Not sure I need as many as I've added.... I just really love the distinct function

R/process_lookup_sc_demographics.R
SwiftySalmon: Not accurate at all. Just thought it would be interesting to have it in while I do my investigations. Can remove at a later date

Jennit07 (approved):
Ran through the code and it works fine. Logic looks good. Happy to approve/merge!

PR #904, opened 2024-01-30
Sc all at speedup

lizihao-anu:
As discussed, separate the pull request from sds pull request


Jennit07 (approved):
Tested this and it works as expected. Great work speeding up this function, this is going to make a big impact on our processing time for `create_episode_file`. Ready to approve/merge!

PR #908, opened 2024-02-05
Bug - NSUs being assigned psychiatry in the service use cohort

Jennit07:
Due to a coding error, the `dplyr::pick` was not working correctly when trying to take the average for assigning the label `unassigned`. This was causing NSUs to have the label `psychiatry` attached as this was the next name available. The fix is to specify all of the columns needed to take an average for assigning the label `unassigned`. This has a big impact on the figures. 

The  current 'live' files:

```
  service_use_cohort       n
   <chr>                <int>
 1 Elective Inpatient  488510
 2 Geriatric           293863
 3 Limited Daycases    725538
 4 Maternity           206488
 5 Multiple Emergency 1140659
 6 Outpatients        1102837
 7 Prescribing        3982957
 **8 Psychiatry         1601193**
 9 Routine Daycase     178739
10 Single Emergency   1116014
**11 Unassigned          250917**
12 Unscheduled Care   1718283
13 NA                   55697

```

after the fix: 

```
   service_use_cohort       n
   <chr>                <int>
 1 Elective Inpatient  488521
 2 Geriatric           293869
 3 Limited Daycases    725583
 4 Maternity           206488
 5 Multiple Emergency 1140688
 6 Outpatients        1102878
 7 Prescribing        3982958
 **8 Psychiatry          111859**
 9 Routine Daycase     178742
10 Single Emergency   1116042
**11 Unassigned         1740237**
12 Unscheduled Care   1718269
13 NA                   55180
```

Ive asked the source team if they can run through a test to see the impact on their figures so i am opening this as a draft for now incase further changes are needed. 

closes #907

PR #911, opened 2024-02-07
Bug - `high_cc` in demographic cohort showing `NAs` instead of `TRUE/FALSE`

### LINKED ISSUE #910: Bug- demographic cohort showing NAs
the variable `high_cc` shows NAs but this should be `TRUE/FALSE`. Picked up by the source team during discussions of #907

Jennit07:
Issue was caused by the conditional statement not recognising `spec == G5`. The fix is to wrap the code in a `case_when` statement to make sure everything is being captured. 

This is also updated in the code for #908 so that we can test the demographic cohorts

closes #910

### DIFF
--- a/R/create_demographic_lookup.R
+++ b/R/create_demographic_lookup.R
@@ -344,18 +344,21 @@ assign_d_cohort_high_cc <- function(dementia,
                                     liver,
                                     cancer,
                                     spec) {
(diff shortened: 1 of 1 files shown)


SwiftySalmon (approved):
looks fine

PR #859, opened 2023-11-07
711 pop estimate eg datazone

lizihao-anu:
Will close #711

PR #913, opened 2024-02-14
Bug - Fix get pop path failing and preventing the indiv file from running. 

Jennit07:
Found an issue which was preventing the individual file from running. I think this was missed on the PR #859 where the function works well but we forgot to apply this within the code. Tested the changes while looking at missing variables and it works fine now.


SwiftySalmon (approved):
looks fine! good catch.

PR #914, opened 2024-02-14
Bug - deal with missing variables

Jennit07:
While looking into the old files I noticed there were some missing variables which include: 
Individual file: 
- `op_newcons_dnas`
- `op_cost_dnas`

Latest year file: 
- `ch_sc_id_cis`
- `ch_postcode`
- `person_id`
- `sc_latest_submission`
- `sc_send_lca`

This PR corrects the code for dealing with these variables

### DIFF
--- a/R/create_episode_file.R
+++ b/R/create_episode_file.R
@@ -143,8 +143,9 @@ create_episode_file <- function(
     episode_file <- episode_file %>%
       dplyr::mutate(
         ch_chi_cis = NA,
-        sc_id_cis = NA,
+        ch_sc_id_cis = NA,
         ch_name = NA,
+        ch_postcode = NA,
         ch_adm_reason = NA,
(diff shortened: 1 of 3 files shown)


R/create_episode_file.R
```diff
@@ -159,7 +160,9 @@ create_episode_file <- function(
         hc_cost_q4 = NA,
         hc_provider = NA,
         hc_reablement = NA,
-        sds_option_4 = NA,
```
SwiftySalmon, line 162: has sds_option 4 been removed?

SwiftySalmon (approved):

R/create_episode_file.R
Jennit07: It is coded in smrtype instead with the other 3 options so i removed this from here

Issue #906, opened 2024-02-01
Investigate Homelessness processing code


Jennit07:
Check `process_extract_homelessness.R`

An issue we have noticed is that we are filtering data in line with the published SG figures. It is unclear why we filter the data and also add east ayrshire as an exception to the rule. We need to investigate this further. From the homlessness analyst team's point of view, they are interested in whether an individual has had a homelessness application in the FY and we should be including all records we have. To Do: 

- [x] contact James when he is back in the office to discuss the methodology
- [x] contact the SG for an update on the total HL1 assessments broken down by FY and Quarter (requested 7/02/24)

See this code below for where we do this:
```
  if (!is.null(completeness_data)) {
    filtered_data <- data %>%
      dplyr::left_join(completeness_data,
        by = c("year", "sending_local_authority_name")
      ) %>%
      dplyr::filter(
        dplyr::between(.data[["pct_complete_all"]], 0.90, 1.05) |
          .data[["sending_local_authority_name"]] == "East Ayrshire"
      )
  } else {
    filtered_data <- data
  }
```

PR #931, opened 2024-03-22
Bug - HL1 filtering to east ayrshire data only in 22/23

Jennit07:
Due to a bug in the file, the 22/23 file was filtering data for east ayrshire only. This is due to an error with matching to the SG completeness file. Improvements to the code here is necessary as this was not picked up in the automated process and was a manual change when this file updated. This is related to the issue #906 

The episode and individual files for 22/23 need to be re-ran to correct this.


SwiftySalmon (approved):
all good

PR #945, opened 2024-04-29
Major update of Care Home script

SwiftySalmon:
lots of changes to the care home processing script. Details in the document on sharepoint. Needs thouroughly checked!!!


R/process_sc_all_care_home.R
```diff
@@ -28,53 +29,63 @@ process_sc_all_care_home <- function(
     spd_path = read_file(get_spd_path()),
```
Jennit07, line 29: on line 28 `ch_name_lookup_path` the `read_file` is not working due to the document being xlsx

R/process_sc_all_care_home.R
```diff
@@ -85,47 +96,63 @@ process_sc_all_care_home <- function(
       -"min_ch_provider",
       -"max_ch_provider"
     ) %>%
-    # tidy up ch_provider using 6 when disagreeing values
     tidyr::fill(.data[["ch_provider"]], .direction = "downup") %>%
```
Jennit07, line 99: Do we need this line? when running this, we change mismatching min/max cases to 6 on lines 89-93. There also shouldnt be any NA values as this was corrected on line 70. Im not sure what the fill is doing as my understanding is that we have already done this on lines 89-93.

R/process_sc_all_care_home.R
```diff
@@ -28,53 +29,63 @@ process_sc_all_care_home <- function(
...
     # sort data
     dplyr::arrange(
-      "sending_location",
-      "social_care_id",
-      "ch_admission_date",
-      "period"
+      .data[["sending_location"]],
+      .data[["social_care_id"]],
+      .data[["period"]],
+      .data[["ch_admission_date"]]
```
Jennit07, line 77: Agree this new sort is working correctly and now the dates are in the right order for later.

R/process_sc_all_care_home.R
```diff
@@ -28,53 +29,63 @@ process_sc_all_care_home <- function(
...
+    replace_sc_id_with_latest() %>%
+    dplyr::select(-latest_flag, -latest_sc_id)
 
+
+  # cleaning and matching care home names -  Zihao looking into care home names
   name_postcode_clean <- fill_ch_names(
     ch_data = ch_clean,
     ch_name_lookup_path = ch_name_lookup_path,
     spd_path = spd_path
   )
```
Jennit07, line 64: I skipped this part as i couldnt read in the ch names workbook and knew this was open in a separate PR

R/process_sc_all_care_home.R
```diff
@@ -28,53 +29,63 @@ process_sc_all_care_home <- function(
...
     dplyr::group_by(
       .data[["sending_location"]],
       .data[["social_care_id"]]
     ) %>%
+    # work out the min and max ch provider in an episode
     dplyr::mutate(
       min_ch_provider = min(.data[["ch_provider"]]),
       max_ch_provider = max(.data[["ch_provider"]]),
+      # if care home provider is different across cases, set to "6".
+      # tidy up ch_provider using 6 when disagreeing values
```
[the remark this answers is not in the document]
Jennit07: In saying that, i think the code is working as expected. I was just wondering whether we are using the correct code or if it should be recoded.

R/process_sc_all_care_home.R
```diff
@@ -85,47 +96,63 @@ process_sc_all_care_home <- function(
...
-    dplyr::ungroup()
+    # create a count of each time the nursing provision changes between records with the same admission date
+    dplyr::mutate(split_episode_counter = ifelse(split_episode == TRUE, dplyr::row_number(), NA)) %>%
+    dplyr::group_by(
+      .data[["social_care_id"]],
+      .data[["sending_location"]]
+    ) %>%
+    # fill split episode counter. This will create a new id number for each different nursing provision within an episode
+    tidyr::fill(split_episode_counter, .direction = c("down")) %>%
+    dplyr::select(-previous_nursing_care_provision, -split_episode)
```
Jennit07, line 151: Well done for dealing with this bit of code. It is very complex and took me a while to get my head around what is going on! I think the new logic is working well to split up episodes where they have changed due to a change in nursing provision or a new start date. I followed along with your example chi and could see where changes were happening. The fill `down` then resolves the counter and produces 3 continuous episodes in the particular example i was following. Very complex logic to work with!

R/process_sc_all_care_home.R
```diff
@@ -138,8 +165,8 @@ process_sc_all_care_home <- function(
       dplyr::desc(.data[["period"]]),
       dplyr::desc(.data[["ch_discharge_date"]]),
       dplyr::desc(.data[["ch_provider"]]),
-      dplyr::desc(.data[["record_date"]]),
-      dplyr::desc(.data[["qtr_start"]]),
+      dplyr::desc(.data[["period_end_date"]]),
+      dplyr::desc(.data[["period_start_date"]]),
```
Jennit07, line 169: Agree with these variables

R/process_sc_all_care_home.R
```diff
@@ -150,62 +177,39 @@ process_sc_all_care_home <- function(
...
-          "reason_for_admission",
-          "type_of_admission"
-        ),
-        dplyr::first
-      ),
+      dplyr::across(c(
+        "ch_discharge_date",
+        "ch_provider",
+        "period_end_date",
+        "period_start_date",
```
Jennit07, line 184: Agree with these

R/process_sc_all_care_home.R
```diff
@@ -150,62 +177,39 @@ process_sc_all_care_home <- function(
...
       ),
       # If it's the last episode(s) then keep the discharge date(s), otherwise
       # use the end of the quarter
-      ch_discharge_date = dplyr::if_else(
-        .data[["sum_latest_submission"]] == max(.data[["sum_latest_submission"]]),
-        .data[["ch_discharge_date"]],
-        .data[["record_date"]]
+      ch_discharge_date = dplyr::if_else(is.na(.data[["ch_discharge_date"]]),
+        .data[["period_end_date"]],
+        .data[["ch_discharge_date"]]
```
Jennit07, line 203: agree with this code.

R/process_sc_all_care_home.R
```diff
@@ -150,62 +177,39 @@ process_sc_all_care_home <- function(
...
       )
     ) %>%
-    dplyr::ungroup()
+    dplyr::ungroup() %>%
+    dplyr::select(-period_start_date, -split_episode_counter)
+
 
   # Compare to Deaths Data
   # match ch_episode data with deaths data
+  # TO DO should this be boxi nrs death dates instead of IT extract deaths?
```
Jennit07, line 212: Yes - i do think this should be using the BOXI NRS dates instead of NRS dates given in the IT extract. However we will need to wait to make this change once we have further info on the deaths and a team decision is made

R/process_sc_all_care_home.R
Jennit07:
```
readxl::read_excel(get_slf_ch_name_lookup_path())
```
This seems to work in place

R/process_sc_all_care_home.R
```diff
@@ -237,38 +241,61 @@ process_sc_all_care_home <- function(
...
-    # ch_sc_id_cis
-    # uses the social care id and sending location so can be used for
+    # gives cases their unique CIS identifier
+    dplyr::mutate(ch_chi_cis = ifelse(continuous_stay_chi == FALSE, dplyr::row_number(), NA)) %>%
+    dplyr::group_by(
+      .data[["social_care_id"]],
+      .data[["sending_location"]]
+    ) %>%
+    # fills in CIS identifier for all cases
+    tidyr::fill(ch_chi_cis, .direction = c("down")) %>%
```
Jennit07, line 271: i think this is working but i have some questions about the example shown in the investigation document

R/process_sc_all_care_home.R
```diff
@@ -237,38 +241,61 @@ process_sc_all_care_home <- function(
 
   # Continuous Care Home Stays
   # Stay will be continuous as long as the admission date is the next day or
-  # earlier than the previous discharge date
-
+  # earlier than the previous discharge date.
+  # creates a CIS  flag for CHI across all of scotland
+  # and a CIS for social care ID and sending location for just that LA
   ch_markers <- matched_deaths_data %>%
```
Jennit07, line 247:
```suggestion
  ch_chi_markers <- matched_deaths_data %>%
```

R/process_sc_all_care_home.R
```diff
@@ -237,38 +241,61 @@ process_sc_all_care_home <- function(
...
+    dplyr::group_by(
+      .data[["social_care_id"]],
+      .data[["sending_location"]]
+    ) %>%
+    # fills in CIS identifier for all cases
+    tidyr::fill(ch_chi_cis, .direction = c("down")) %>%
+    # This is the same but uses the social care id and sending location so can be used for
     # episodes that are not attached to a CHI number
     # This will restrict continuous stays to each Local Authority
     dplyr::group_by(.data[["social_care_id"]], .data[["sending_location"]]) %>%
```
Jennit07, line 275:
```suggestion
    sc_ch_id_markers <- chi_chi_markers %>%
    dplyr::group_by(.data[["social_care_id"]], .data[["sending_location"]]) %>%
```
I think the sc_id_cis can be split onto a new object just to make it a bit clearer that its separate even though it uses the same logic as above. 

Alternatively we could put the logic into a function since we are repeating this for sc_id_cis

R/process_sc_all_care_home.R
```diff
@@ -277,32 +304,33 @@ process_sc_all_care_home <- function(
     ) %>%
     dplyr::mutate(
       ch_ep_start = min(.data[["ch_admission_date"]]),
+      # Creates a vector for the earliest date out of the end of period and discharge date.
+      # And will then select what ever is the latest date out of those
       ch_ep_end = max(
         pmin(
-          .data[["record_date"]],
+          .data[["period_end_date"]],
```
Jennit07, line 311: agree

R/process_sc_all_care_home.R
```diff
@@ -312,6 +340,15 @@ process_sc_all_care_home <- function(
...
     ) %>%
+    # recode the care home provider description
+    dplyr::mutate(ch_provider_description = dplyr::case_when( # from social care syntax
+      ch_provider == 1 ~ "LOCAL AUTHORITY/HSCP/NHS BOARD",
+      ch_provider == 2 ~ "PRIVATE",
+      ch_provider == 3 ~ "OTHER LOCAL AUTHORITY",
+      ch_provider == 4 ~ "THIRD SECTOR",
+      ch_provider == 5 ~ "NHS BOARD",
+      ch_provider == 6 ~ "OTHER"
+    )) %>%
```
Jennit07, line 351: Nice addition

R/process_sc_all_care_home.R
```diff
@@ -237,38 +241,62 @@ process_sc_all_care_home <- function(
...
+    dplyr::select(
+      -previous_discharge_date_chi, -continuous_stay_flag_chi, -previous_continuous_stay_flag_chi, -continuous_stay_chi,
+      -previous_discharge_date_sc, -continuous_stay_flag_sc, -previous_continuous_stay_flag_sc, -continuous_stay_sc,
+      -dis_after_death, -death_date_nrs, -death_date_chi
+    )
 
-  # Do a recode on the old reason for admission
+
+  # Do a recode on the old reason for admission for respite stays.
   adm_reason_recoded <- ch_markers %>%
```
SwiftySalmon, line 300:
```suggestion
  adm_reason_recoded <- sc_ch_id_markers %>%
```

R/process_sc_all_care_home.R
SwiftySalmon: fixed

R/process_sc_all_care_home.R
SwiftySalmon: I think this is a leftover from the code that was there before. Need to look at it when more awake

R/process_sc_all_care_home.R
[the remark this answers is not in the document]
SwiftySalmon: I think 6 now means unknown/mix of things

R/process_sc_all_care_home.R
SwiftySalmon: agree. made new issue for it

R/process_sc_all_care_home.R
```diff
@@ -19,62 +19,73 @@
...
+    replace_sc_id_with_latest() %>%
+    dplyr::select(-latest_flag, -latest_sc_id)
 
+
+  # cleaning and matching care home names
   name_postcode_clean <- fill_ch_names(
     ch_data = ch_clean,
     ch_name_lookup_path = ch_name_lookup_path,
     spd_path = spd_path
   )
```
Jennit07, line 64: This section is not working. Produces this error: 

```
Error in charToDate(x) : 
  character string is not in a standard unambiguous format
Called from: charToDate(x)
```

Jennit07: works after bringing the code up to date and resolving conflicts

R/process_sc_all_care_home.R
```diff
@@ -19,62 +19,72 @@
 #' @family process extracts
 #'
 #' @export
-#'
+
+
 process_sc_all_care_home <- function(
     data,
     sc_demog_lookup = read_file(get_sc_demog_lookup_path()) %>% slfhelper::get_chi(),
     it_chi_deaths_data = read_file(get_slf_chi_deaths_path()),
```
Jennit07, line 27:
```suggestion
    it_chi_deaths_data = read_file(get_slf_chi_deaths_path()) %>% slfhelper::get_chi(),
```
need to convert to chi as this causes an error matching later on

R/process_sc_all_care_home.R
```diff
@@ -237,72 +233,98 @@ process_sc_all_care_home <- function(
...
-      ch_sc_id_cis = cumsum(.data[["continuous_stay_sc"]])
+    # gives cases their unique CIS identifier
+    dplyr::mutate(ch_chi_cis = ifelse(continuous_stay_chi == FALSE, dplyr::row_number(), NA)) %>%
+    dplyr::group_by(
+      .data[["social_care_id"]],
+      .data[["sending_location"]]
     ) %>%
-    dplyr::ungroup()
+    # fills in CIS identifier for all cases
+    tidyr::fill(ch_chi_cis, .direction = c("down"))
```
Jennit07, line 263: I think this is working more like how we want it to. However, due to using the lag function and NA being assigned to the top row, it means that for some cases the first episode is being assigned 1 and the next episode would be assigned 2 even if it is part of the same stay. Its due to the missing date on the row before that it is not calculating this properly. The cases after the first row of each episode is working as we would expect.

I have shown @SwiftySalmon a few examples on a call and she is going to have another look at the logic. Other than this part of care homes, i think everything else is working as we would expect. Well done Megan for all your hard work on this complex PR

lizihao-anu (approved):
Looks ok to me. Happy to merge it when it is ready.

Jennit07:
Pushed a few changes for the deaths methodology. Ready to merge

PR #947, opened 2024-04-30
only select columns we want in ltc raw data

### LINKED ISSUE #944: Investigate extra variable in individual file
Extra variable called `...1` between `hl1_in_fy` and `arth_date` has appeared in the individual file. Investigate why this has pulled through and correct this. Check episode file to ensure this is also not there.

lizihao-anu:
easy fix but took a long time to identify the issue. Close #944

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -7,10 +7,13 @@ adtf
 arrivalmode
 arth
 atlassian
+atrialfib
 attendcat
 aut
 bedday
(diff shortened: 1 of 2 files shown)


lizihao-anu:
@check-spelling-bot apply [updates](https://github.com/Public-Health-Scotland/source-linkage-files/actions/runs/8897746003/attempts/1).

SwiftySalmon (approved):
nice work finding it

PR #948, opened 2024-05-03
latest social care ID update

SwiftySalmon:
for some reason the latest scid code was overwritten after the march update?? anyway, now it is fixed.

### DIFF
--- a/R/replace_sc_id_with_latest.R
+++ b/R/replace_sc_id_with_latest.R
@@ -7,31 +7,40 @@ replace_sc_id_with_latest <- function(data) {
   # Check for required variables
   check_variables_exist(
     data,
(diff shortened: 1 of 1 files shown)


rchlv:
Additional comment from SC team for information:
The period variable in demographics is taken from the header; it is not the same as in other sections and it does not show the period of activity necessarily and shouldn't be used for analysis. It is fine in terms of getting the latest record.
For example, some HSCPs resubmit the whole demographics file each time rather than just new records, so the period would always change to the header data rather than referring to the period of activity. Or in other cases, person A could have activity in Q1, 2 and 3 but only submitted in Q4; this would only show Q4.

rchlv (approved):
This looks fine to merge, thank you.

PR #949, opened 2024-05-06
Merge May24 NI update into June update branch

Jennit07:
Following completing a May update of years 2021, 2122, 2223 and 2324 for the National Indicators team, there was one small change required during the running process. This was to do with reading the client extract. For 2324 it did not like running the `dplyr::collect()` after manipulating the variables. I have changed the ordering of this and it seems to run fine. I would like to merge this into the June update in case this causes an issue during the next update.


SwiftySalmon (approved):
seems sensible

PR #951, opened 2024-05-10
Develop high level tests across years

Jennit07:
WIP - just need to write to disk for the June update. 

I think this can be developed further with some visualisations!


_targets.R
```diff
@@ -591,7 +591,11 @@ list(
         data = episode_file,
         year = year
       )
-    ) # ,
+    ),
+    tar_target(
+      cross_year_tests,
+      process_tests_cross_year(year = year)
+    ), # ,
```
Jennit07, line 598: I think i need to come back to this as im not sure this will ensure that all newly created ep files will be included in the test comparison

SwiftySalmon (approved):
looks fine. We'll run it and see how it goes

PR #952, opened 2024-05-17
Update `00_sort_bi_extracts` to write anon_chi

Jennit07:
As part of ensuring anon-chi is used on disk, an update to sort bi extract script. Opening as a draft for now as this works for extracts with `chi` variable available but we do take some extracts without chi e.g. gp ooh data. Some thought is needed to expand the code to work around this. Possibly a conditional statement.


lizihao-anu:
Update the `00_Sort_BI_Extracts.R` to 

- enable parallel computing
- tell whether a file have a chi column or not

00_Sort_BI_Extracts.R
```diff
@@ -20,31 +20,52 @@ extract_financial_year <- function(filename) {
...
+        financial_year_dir,
+        paste0("anon-", basename(csv_file))
+      )
+      read_file(csv_file) %>%
+        dplyr::rename_with(~ paste0("chi"), tidyselect::contains("UPI")) %>%
+        slfhelper::get_anon_chi() %>%
+        readr::write_csv(file = new_file_path)
+      cat("Replaced chi with anon chi:", csv_file, "to", new_file_path, "\n")
+    } else {
+      new_file_path <- file.path(financial_year_dir, basename(csv_file))
```
Jennit07, line 53: Remove this line

00_Sort_BI_Extracts.R
```diff
@@ -20,31 +20,52 @@ extract_financial_year <- function(filename) {
...
       dir.create(financial_year_dir)
     }
 
+    # set up new file path location to move each file to their destination.
+    chi_in_file <- is_chi_in_file(csv_file)
+    if (chi_in_file) {
+      new_file_path <- file.path(
+        financial_year_dir,
+        paste0("anon-", basename(csv_file))
+      )
```
Jennit07, line 46: Move this file path to line 40?

Jennit07:
> Update the `00_Sort_BI_Extracts.R` to
> 
> * enable parallel computing
> * tell whether a file have a chi column or not
> 
> It works well as I tested. But I do have a question about the new file name. Since we have added a prefix "anon-" for those files, will it break the existing data pipeline? If so, we should
> 
> * modify the functions which pick up those files. OR,
> * we just do not add the prefix to file names.

Hi @lizihao-anu Thanks for all your help with this. We want to keep the `anon-` prefix for consistency as the updated `get_boxi_extract` path has renamed this for all extracts when we are changing everything to `anon_chi`. I have made the changes we have discussed and also tested a file. I am happy the function is working as intended. 

I will mark this PR ready for review. Could you please do a final check to confirm you are happy with this and then approve/merge into the June24 update branch. Thanks! :)

lizihao-anu (approved):

PR #971, opened 2024-06-21
Add sys time to functions

SwiftySalmon:
Added in sys_time to functions when creating the ep and ind files because I like to see where it's at when I run things manually


Jennit07 (approved):
Looks good, like the idea!

PR #972, opened 2024-06-21
adding the death dates to activity after death cases

SwiftySalmon:


Jennit07 (approved):
Tested this through and working as expected. Happy to approve/merge

PR #980, opened 2024-07-03
Update Targets to include combined deaths processing

Jennit07:
After having a discussion about the processing flow, i have updated this to reflect the following: 

- Check the years specified in `targets.R`
- Run one year first to identify any potential bugs
- Work with another analyst to resolve any potential issues
- Once happy, check that all years specified in `targets.R` are correct
- One analyst can run the script `run_all_targets` to run the remaining years targets. ETA is tbc
- Once all targets have completed we can check through all of the tests for each of the extracts.
- Once we are happy, we can divide up the years between analysts and run the episode files. It is also recommended to run one year first to check for any potential bugs. 
- When we are happy with the episode files, we can then run the individual files. 

Summarised the process here but i will also update the SOP


lizihao-anu (dismissed):

_targets.R
```diff
@@ -542,106 +542,114 @@ list(
...
-    #     partitioning = c("hscp2018"),
-    #     compression = "zstd",
-    #     version = "latest"
-    #   )
-    # )
+    )
+  ),
+  tar_target(
+    combined_deaths_lookup,
+    process_combined_deaths_lookup()
```
lizihao-anu, line 556: I guess this needs to be moved out from targets.

SwiftySalmon: where does it go instead?

Run_SLF_Files_targets/run_all_targets.R
```diff
@@ -0,0 +1,5 @@
+library(targets)
+
+# use tar_make_future() to run targets for all years
+# This will run everything needed for creating the episode file.
+tar_make_future()
```
lizihao-anu, line 5:
```suggestion
tar_make_future()

combined_deaths_lookup = process_combined_deaths_lookup()
```

_targets.R
lizihao-anu: Sorry for not making it perfectly clear. We should move it to `run_all_targets.R`. By moving it, we can make sure that combining deaths lookup is after the death file for each year being produced. I add a code suggestion below.

SwiftySalmon (approved):
don't fully understand what is happening here but I trust yous.

PR #979, opened 2024-07-02
Add client flags

SwiftySalmon:
- removed code that wasn't needed.
- updated housing codes
- latest social care ID
- changed "mental health problems" to "mental health disorders" in line with PHS style guide

PR #981, opened 2024-07-03
Person id sds

SwiftySalmon:
added back in missing person_id for SDS. and also added latest_flag back into client lookup in targets


_targets.R
```diff
@@ -464,7 +464,7 @@ list(
         year = year,
         sc_demographics = sc_demog_lookup %>%
           slfhelper::get_chi() %>%
-          dplyr::select(c("sending_location", "social_care_id", "chi")),
+          dplyr::select(c("sending_location", "social_care_id", "chi", "latest_flag")),
```
Jennit07, line 467: I think you need to make this change on `process_lookup_sc_client` on this branch as well otherwise targets might break because its missing here

Jennit07 (changes requested):
One comment, but otherwise i think this is ready to merge!

_targets.R
Jennit07: Noticed you've done this in #979 but think it also needs to match on this one so that you dont get a conflict

SwiftySalmon: done!

PR #982, opened 2024-07-04
GPooh and acute cup

### LINKED ISSUE #775: Bring in the `cup_marker` from the UCD to GP OoHs

lizihao-anu:
close #775 #776


R/read_extract_acute.R
```diff
@@ -149,6 +149,7 @@ read_extract_acute <- function(year, file_path = get_boxi_extract_path(year = ye
       ipdc = "Inpatient Day Case Identifier Code",
       cij_ipdc = "CIJ Inpatient Day Case Identifier Code (01)",
       lineno = "Line Number (01)",
+      case_reference_number = "Case Reference Number [C]",
```
SwiftySalmon, line 152: Need to read in case ref no above this

SwiftySalmon: "Case Reference Number [C]" is missing from the boxi extract? where does this come from?

SwiftySalmon: nm. I was running a differnt year to you. Ran with 1920 and it works fine

R/add_activity_after_death_flag.R
```diff
@@ -125,8 +125,8 @@ add_activity_after_death_flag <- function(
 #'
 #'
 # Read data------------------------------------------------
-process_deaths_lookup <- function(update = latest_update(),
-                                  write_to_disk = TRUE, ...) {
+process_combined_deaths_lookup <- function(update = latest_update(),
+                                           write_to_disk = TRUE, ...) {
```
SwiftySalmon, line 129: are these meant to be different? I do not want to overwrite changes on the September branch

lizihao-anu: > are these meant to be different? I do not want to overwrite changes on the September branch

Yeah. It's something that Jen and I agreed but she forgot to change. `process_combined_deaths_lookup` make more sense and is more intuitive. This function is used in `run_all_targets.R` and the new name `combined` is already in use.

SwiftySalmon (approved):
looks good to me

PR #987, opened 2024-07-29
review deaths dates processing

### LINKED ISSUE #986: Death dates
R/process_it_chi_deaths.R 

For this we should be using chi_death_date and then joining with BOXI extract

Default should be death date from boxi nrs death date, and filling in affected death dates from chi_death_date from IT extract

Or if CHIs are affected by CHI seeding issue then use chi_death_date from IT extract (needs investigation on how to identify cases)

Jennit07:
Opening this as a draft PR for now to fix #986 

Previously, the IT deaths extract was using the nrs weekly date but if this was missing, it would use the chi death date. We agreed as a team that this was wrong and the boxi nrs deaths date was the most reliable. This was the previous code in process_it_chi_deaths: 

`death_date = dplyr::coalesce(.data$death_date_nrs, .data$death_date_chi)` 

I have removed this and selected chi and death_date_chi so that we do not use the weekly nrs death date provided by IT. This is unvalidated and found to be unreliable. Instead, the IT chi death date is left with no modification and has two variables: `chi` and `death_date_chi`. 

Later in the process we create a deaths lookup which is one row per chi for deaths that occurred within the financial year. This is used to match onto the episode file later in the process. By default, this was set up to use the boxi nrs death date which is submitted monthly and is validated. We agreed this should be the correct method going forward. However, this does not take into account the chi date. I have updated this in a way that: 

`if the BOXI NRS date does not match the chi death date, use the chi death date` here is the code which does this: 

```
    dplyr::mutate(
      death_date = dplyr::if_else(.data$record_keydate1 != .data$death_date_chi,
        .data$death_date_chi, .data$record_keydate1
      ),
      deceased = TRUE,
      .keep = "unused"
    )
```
When i checked this with 1920 data, there was only one case which changed from 31-03-2020 to 01-04-2020. This was not one of the chi's which had a 40year difference. The 40 year difference mainly was in the weekly nrs data. 

We possibly need to review this to say `use the chi date if X time between the boxi nrs date and chi date`

Another thing to check the methodology is on care home data which is explicitly using the nrs weekly data but if this is not available then use chi data. This is outstanding on this PR. The changes should be the same from the deaths lookup

TO DO: 
- [x] check time difference between boxi nrs and chi death dates
- [x] check care home methodology to reflect new changes

opening a draft for now to see the changes.


lizihao-anu:
Should be ready for review. @Jennit07 @SwiftySalmon

SwiftySalmon (approved):
seems fine

PR #988, opened 2024-08-02
Update older years to bring the data in line with our newest processes. 

### LINKED ISSUE #893: Run older files with the R code
In #860 this revealed that the older files have now become out of date with the R process. We need to run years 1415, 1516 and 1617 as an ad hoc update to bring this in line with the latest R version to correct any mismatching variables.

Jennit07:
Closes #893 

I have ran the old years: 1415, 1516 and 1617 and updated the code in line with our latest processes. Ive declared any missing variables and they should appear now automatically when running those years. 

There are a few benefits of spending the time to update these:
- fixes a bug on slfhelper when reading any old year alongside new years (creates a mismatch in variables)
- our code is consistent
- our code is up to date so that if we need to run these years again, this should now be faster and more efficient. 

The tests for each episode and individual files is available in hscdiip. I noticed a big difference in A&E in the number of females being a large difference. However this was an issue before that we could put this down to the methodology in SPSS to R code changing. 

Still to do:
- When checking through the PR, ensure the tests are also checked for any major differences.
- move the files to hscdiip once happy with this PR for analytical use.


Jennit07:
Will need to run these files again once the code for the September update is finalised. This should be straightforward to do now we know the code is working and able to produce the episode and individual files.

R/replace_sc_id_with_latest.R
SwiftySalmon, line 1: This one needs to be the same as in the care home branch. I think it should be period. latest flag is not accurate at all

SwiftySalmon (commented):
everything looks fine except replace_sc_id_with_latest

lizihao-anu (approved):

lizihao-anu (approved):

PR #989, opened 2024-08-07
Investigate care home cis markers

Jennit07:
In the care home methodology update #945 the cis markers were working but not for episodes nested within a stay, it would wrongly assign the 'next' episode counter. This was due to the lag function assigning NA to the first row.

This update fixes the nested cases and i think it is working properly now for all scenarios. Where the first row is NA for `previous_discharge_date` this will use the `ch_discharge_date` which will be the first episode. 

We also use the following code, adapted from Social care scripts to create the cis markers. This simplifies the code and seems to work for all scenarios. e.g. when the first episode linked to a chi is NA and the next two episodes are continuous, this will return 1,2,2 correctly. Previously this would return 1,2,3. 
```
mutate(continuous_marker = accumulate(new_episode[-1], .init = 1, ~ if (.y == "Y") {.x + 1} else {.x}))
```

### DIFF
--- a/R/process_sc_all_care_home.R
+++ b/R/process_sc_all_care_home.R
@@ -237,30 +237,44 @@ process_sc_all_care_home <- function(
   # creates a CIS  flag for CHI across all of scotland
   # and a CIS for social care ID and sending location for just that LA
   ch_chi_markers <- matched_deaths_data %>%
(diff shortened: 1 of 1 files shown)


SwiftySalmon (approved):
looks great!

PR #1014, opened 2024-10-07
Write temp data

lizihao-anu:


R/create_episode_file.R
```diff
@@ -32,11 +32,13 @@ create_episode_file <- function(
     slf_deaths_lookup = read_file(get_slf_deaths_lookup_path(year)) %>% slfhelper::get_chi(),
     sc_client = read_file(get_sc_client_lookup_path(year)) %>% slfhelper::get_chi(),
     write_to_disk = TRUE,
-    anon_chi_out = TRUE) {
+    anon_chi_out = TRUE,
+    test_mode) {
```
Jennit07, line 36: Do we need to set a default here?

R/create_individual_file.R
```diff
@@ -17,7 +17,8 @@ create_individual_file <- function(
     homelessness_lookup = create_homelessness_lookup(year),
     write_to_disk = TRUE,
     anon_chi_in = TRUE,
-    anon_chi_out = TRUE) {
+    anon_chi_out = TRUE,
+    test_mode) {
```
Jennit07, line 21: Again, wondering if we need to set a default here?

R/create_episode_file.R
```diff
@@ -140,11 +145,13 @@ create_episode_file <- function(
       year,
       slf_deaths_lookup
     ) %>%
+    write_temp_data(year, file_name = "ep_temp5", test_mode) %>%
```
Jennit07, line 148: I think we could move this temp5 up to just before joining on the deaths data?

R/create_individual_file.R
```diff
@@ -90,14 +92,17 @@ create_individual_file <- function(
     recode_gender() %>%
     clean_individual_file(year) %>%
     join_cohort_lookups(year) %>%
+    write_temp_data(data, year, file_name = "indiv_temp2", test_mode) %>%
```
Jennit07, line 95: i think we could move this up to line 84 and line 89 so that we save a temp file right after the aggregate? i remember that function takes a while to process

R/create_individual_file.R
Jennit07:
```suggestion
    test_mode = FALSE) {
```

R/create_episode_file.R
Jennit07:
```suggestion
    test_mode = FALSE) {
```

lizihao-anu:
Some ideas agreed:

- Keep 1920, 2324, 2425 in sourcedev, and clear all other folders (except extracts folders)
- Remove "TEST-" prefix for temp files.
- Add parameter `write_temp_to_disk` to `write_temp_file()`.

Subject to further changes:

- Write out all ep temp files for all years. Before running individual files, delete ep temp files to save storage. (Probably ok but double check)
-

Jennit07 (approved):
Looks good. Ready to merge

PR #1015, opened 2024-10-07
death join and distinct refined death

lizihao-anu:


Jennit07 (changes requested):
Think this is looking good. Although i noticed that `activity_after_death` is now showing `1` or `NA` but this should be `1` or `0`. Could you have another look please? I think this will be ready to merge after! Thanks

lizihao-anu:
ready for merge.

Jennit07 (approved):
Looks good

PR #1021, opened 2024-10-29
Clean test folder

Jennit07:
I've cleaned out the tests folder and updated the file paths accordingly. Should now be a bit more automated and also all tests are in the same naming convention. E.g. `Sep_2024_2021_ep_file_tests` or `Sep_2024_2021_extract_tests`. I zipped up 2023 tests with the idea that we should archive/zip tests once we have a full year available

Opening as a draft for now as still outstanding:

Still TO DO: 
 - clean costs tests
 - update costs tests write to disk


lizihao-anu:
Hi, is it ready for review?

Jennit07:
> Hi, is it ready for review?

Hi @lizihao-anu I've pushed some changes and this is now ready for review. I've updated this so that it will produce the output only once per quarterly update. Thanks

R/00-update_refs.R
```diff
@@ -10,6 +116,7 @@ latest_update <- function() {
   "Sep_2024"
```
Jennit07, line 116: I think we could automate this using `end_date`

R/00-update_refs.R
```diff
@@ -1,3 +1,109 @@
...
+#'
+#' @description Get the period for Delayed Discharge
+#'
+#' @return The period for the Delayed Discharge file
+#' as MMMYY_MMMYY
+#' @export
+#'
+#' @family initialisation
+get_dd_period <- function() {
+  "Jul16_Jun24"
```
Jennit07, line 103: I think we could also automate this - will add an issue for both

R/00-update_refs.R
```diff
@@ -1,3 +1,109 @@
...
+#' @export End date as dmy
+#'
+end_date <- function() {
+  ## UPDATE ##
+  # Last date in reporting period
+  # Q1 June = 30062024
+  # Q2 September = 30092024
+  # Q3 December = 31122024
+  # Q4 March = 31032024
+  lubridate::dmy(30062024)
```
lizihao-anu, line 32: Are we going to change it to 30092024 for the december update?
```suggestion
  lubridate::dmy(30092024)
```

Jennit07: Yeah! Thanks, good spot :)

lizihao-anu (approved):

PR #1039, opened 2024-12-24
1031 split tests

lizihao-anu:
This branch writes out tests into different xlsx files separately to avoid writing into one file simultaneously and then combine the tests outputs into files by years. So, the final outputs are the same as before. 

There is another branch [1031-remove-tests-from-targets](https://github.com/Public-Health-Scotland/source-linkage-files/tree/1031-remove-tests-from-targets), which takes out all tests from targets and do tests by sequence. Finishing all tests for each year would take around 25 minutes. It is time-consuming and hence this approach is much less favourable.

Summary of key changes:
- Functions added/changed include
    - `setup_tests_file_name`: It is taken out from `write_tests_xlsx` and set up paths for tests file and folder.
    - `combine_multi_xlsx`, `combine_extracts_tests_year`, `combine_lookup_tests`, `combine_tests`: Combine xlsx files together
- `targets` script has been restore for parallel computing for tests and there is no problem of writing tests into one file simultaneously. 
- `processs_tests_sds` has been added filter of `check_year_valid` to avoid corruption in `targets` for `tests_sds_2425`.

A thorough test has been done and successfully went through without any problem. So it is good for review and then merge.


Jennit07:
Hi Zihao, i have had a look through your changes. I think we can have a discussion about this when you are back from leave. I have a few questions: 

- when you say `targets script has been restore for parallel computing for tests and there is no problem of writing tests into one file simultaneously.` im a bit confused about how we should be running the tests. Is this still in targets but then it processes each extract separately? When would we combine them? 
- I am wondering about what happens to each individual workbook when the tests are combined? do we remove the individual workbooks? I am not sure how much space this will take up, although i know these are small files. However, i think we should have some structure/organisation in place as we are writing multiple tests each update. 
- In the combined tests final output, the formatting looks like it is missing even though the individual extracts do have the formatting in place. Im not sure where this is lost. 

Happy to have a discussion about this when you are back, Thanks :)

Jennit07:
todo - wed github session

PR #1043, opened 2025-01-14
Update pkgdown

Jennit07:


lizihao-anu:
Why do we do this though? Does anything malfunction or is it out of date?

Jennit07:
> Why do we do this though? Does anything malfunction or is it out of date?

This is related to the website i was sharing on the call. The main purpose/nice feature of this is there is a reference tab which links to all of the documentation for every function in the package. The Readme page is also out of date so it would be good to get things updated/more documentation on the package. If you look at the SLFhelper package website, it also has vignettes where we added memory consumption documentation. The idea is the same to have this available for the `createslf` package. Happy to discuss further if needed as this took me a while to figure out to get it working again. The link was initially broken/not updated in some time

lizihao-anu:
> > Why do we do this though? Does anything malfunction or is it out of date?
> 
> This is related to the website i was sharing on the call. The main purpose/nice feature of this is there is a reference tab which links to all of the documentation for every function in the package. The Readme page is also out of date so it would be good to get things updated/more documentation on the package. If you look at the SLFhelper package website, it also has vignettes where we added memory consumption documentation. The idea is the same to have this available for the `createslf` package. Happy to discuss further if needed as this took me a while to figure out to get it working again. The link was initially broken/not updated in some time

Great!

Jennit07:
reminder: before merging, need to remove `update_pkgdown` from pkgdown.yaml file as this was temporary.

PR #1049, opened 2025-02-06
add a new function link_ch2ae

lizihao-anu:
In fact, ("01B", "04B", "GLS", "AE2") are included to link with CH. Technically, DD episodes can also be linked with CH episodes as DD episodes are linked with AE episodes and others but we do not link DD to care home.


R/link_ch2ae.R
```diff
@@ -0,0 +1,119 @@
...
+
+  data <- ep %>%
+    dplyr::select(
+      ep_row_id_CE,
+      anon_chi,
+      recid,
+      record_keydate1,
+      record_keydate2,
+      keytime1,
+      keytime2,
```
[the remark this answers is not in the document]
Jennit07: ch_name, ch_postcode, cup_pathway and cij_pattype, also need to be added to the list

R/link_ch2ae.R
```diff
@@ -0,0 +1,119 @@
...
+    ) %>%
+    dplyr::mutate(
+      ch_name = dplyr::coalesce(ch_name, ch_name.y),
+      ch_postcode = dplyr::coalesce(ch_postcode, ch_postcode.y)
+    ) %>%
+    dplyr::select(-c(
+      dplyr::ends_with(".y"),
+      "test_date"
+    )) %>%
+    dplyr::mutate(ch2ae_flag = 1L)
```
Jennit07, line 83:
```suggestion
    dplyr::mutate(sc_ch_link_ae = 1L)
```
name change suggestion to have this in line with social care variables

Jennit07: Had a discussion about this and we think this might be more useful to create a flag linking care home episodes that result in an admission `sc_ch_link_adm`

R/link_ch_with_ae.R
```diff
@@ -0,0 +1,119 @@
...
+    # remove A&E records that don't lead to an admission,
+    # remove elective admissions,
+    # and remove day cases
+    dplyr::filter(recid == "AE2" &
+      grepl("A", cup_pathway) == TRUE |
+      recid != "AE2") %>%
+    dplyr::filter(!(recid != "AE2" &
+      cij_pattype == "Elective")) %>%
+    dplyr::filter(!(recid != "AE2" &
+      cij_ipdc == "D")) %>%
```
Jennit07, line 50: We will review this code as this was initially some code used that was specific to the query LIST had.

R/create_episode_file.R
```diff
@@ -73,9 +75,13 @@ create_episode_file <- function(
...
+        "ch_name",
+        "ch_postcode",
+        "cup_pathway",
         "cij_marker",
         "cij_start_date",
         "cij_end_date",
+        "cij_pattype",
         "cij_pattype_code",
         "cij_ipdc",
         "cij_admtype",
```
Jennit07, line 87: Im getting this error here: 

ℹ The file temp_ep_file_variable_store_2021.parquet did not exist in
/conf/sourcedev/Source_Linkage_File_Updates/2021, it has now been created.
Error in `check_variables_exist()` at source-linkage-files/R/create_episode_file.R:290:3:
! Variable "hscp" is required, but is missing from `data`.
Run `rlang::last_trace()` to see where the error occurred.

R/link_ch_with_adms.R
```diff
@@ -0,0 +1,161 @@
...
+        recid == "AE2"),
+      0L, sc_ch_link_ae
+      ),
+      sc_ch_link_adms = dplyr::if_else((
+        is.na(sc_ch_link_adms) &
+          recid %in% c("01B", "04B", "GLS")
+      ),
+      0L, sc_ch_link_adms
+      )
+    )
```
Jennit07, line 156: Just wondering about this last part, should the final flags show 1, 0,  NA? or just 1, 0?

The code seems to be working apart from the one error i get above with hscp. But i am wondering about the final output. In the episode file, 01b records have NA with `sc_ch_link_ae` is that right?

lizihao-anu: The final flags should show 1, 0, NA. It is the best practice. Imagine a general scenario -- if we assign zero to those which are supposed to NA and we need to get an average of the variable, the descriptive statistics will be misleading.

sc_ch_link_ae is supposed to be applied to AE2 only while sc_ch_link_adms is supposed to applied to 01B, 04B, and GLS. So, 01B records should have NA with sc_ch_link_ae while AE2 records should have NA with sc_ch_link_adms.

lizihao-anu:
Ready for review and merge

Jennit07 (approved):
Great work!

PR #1051, opened 2025-02-12
Sort it extracts

lizihao-anu:
During pre-processing, the new script `01_Sort_IT_Extracts.R` will do the following automatically and sequentially

- delete files from the previous request
- convert new csv files to parquet files with anonymous chi
- archive the parquet files from the current request and put it to archive folder.
- delete the csv files from the current request

The functions to get it extracts are changed accordingly.

### DIFF
--- a/R/process_lookup_homelessness.R
+++ b/R/process_lookup_homelessness.R
@@ -14,7 +14,7 @@ create_homelessness_lookup <- function(
     year,
     homelessness_data = read_file(get_source_extract_path(year, "homelessness"))) {
   # Specify years available for running
-  if (year < "1617") {
+  if (year <= "1516") {
     return(NULL)
   }
   homelessness_lookup <- homelessness_data %>%


Pre_processing_scripts/pre_prossessing_all.R
```diff
@@ -0,0 +1,2 @@
+source("Pre-processing_scripts/00_Sort_BI_Extracts.R")
+source("Pre-processing_scripts/01_Sort_IT_Extracts.R")
```
lizihao-anu, line 2: @Jennit07 , Would you prefer to have this single script to automate all, or not? If so, could you please add other pre-processing scripts to this please? Thank you!

Pre_processing_scripts/01_Sort_IT_Extracts.R
```diff
@@ -0,0 +1,129 @@
...
+    cli::cli_alert_danger("Detected 0 or more than 1 task numbers! Please check!")
+    return(NULL)
+  }
+}
+
+get_current_it_task_num <- function() {
+  new_csv_files <- list.files(it_extract_path)
+  ltcs_file <-
+    new_csv_files[stringr::str_detect(new_csv_files, "LTCs\\.csv$")]
+  task_num <- stringr::str_extract(ltcs_file, "SCTASK\\d+")
```
Jennit07, line 37:
```suggestion
  new_ltcs_file <-
    new_csv_files[stringr::str_detect(new_csv_files, "LTCs\\.csv$")]
  new_task_num <- stringr::str_extract(ltcs_file, "SCTASK\\d+")
```
Suggest to rename the object so its clear which ones are the new files

Pre_processing_scripts/01_Sort_IT_Extracts.R
```diff
@@ -0,0 +1,129 @@
...
+  if (length(task_num) == 1L) {
+    return(task_num)
+  } else {
+    cli::cli_abort("Detected 0 or more than 1 task numbers! Please check!")
+  }
+}
+
+previous_task_num <- get_previous_it_task_num()
+current_task_num <- get_current_it_task_num()
```
Jennit07, line 46: Can we get current task num to return an error please? currently i am running this and we do not have new extracts but its not clear that is the reason. It returns a generic error

Pre_processing_scripts/01_Sort_IT_Extracts.R
```diff
@@ -0,0 +1,129 @@
...
+
+
+# convert csv to anon parquet ---------------------------------------------
+csv_files <- list.files(it_extract_path,
+  pattern = "SCTASK[0-9]{7}_(PIS_20[0-9]{2}|Deaths|LTCs)\\.csv(\\.gz)?",
+  full.names = TRUE
+)
+
+convert_it_csv_to_parquet <- function(csv_file) {
+  parequet_file <- gsub("\\.csv(\\.gz)?$", ".parquet", csv_file)
```
Jennit07, line 63:
```suggestion
  parquet_file <- gsub("\\.csv(\\.gz)?$", ".parquet", csv_file)
```

Pre_processing_scripts/01_Sort_IT_Extracts.R
```diff
@@ -0,0 +1,129 @@
...
+  data <- read_file(csv_file)
+  new_file <- file.path(
+    it_extract_anon_path,
+    paste0("anon-", basename(parequet_file))
```
Jennit07, line 67:
```suggestion
    paste0("anon-", basename(parquet_file))
```

Pre_processing_scripts/01_Sort_IT_Extracts.R
```diff
@@ -0,0 +1,129 @@
...
+      dplyr::rename_with(
+        ~ paste0("chi"),
+        tidyselect::contains("UPI", ignore.case = FALSE)
+      ) %>%
+      slfhelper::get_anon_chi()
+  }
+  write_file(data, new_file)
+  cli::cli_alert_info("\n {basename(csv_file)} finished at {Sys.time()}")
+  # return(NULL)
+}
```
Jennit07, line 80:
```suggestion
  if (is_chi_in_file(csv_file)) {
    data <- data %>%
      dplyr::rename_with(
        ~ paste0("chi"),
        tidyselect::contains("UPI", ignore.case = FALSE)
      ) %>%
      slfhelper::get_anon_chi()
  } else {
  write_file(data, new_file)
  cli::cli_alert_info("\n {basename(csv_file)} finished at {Sys.time()}")
}
```
Missing a curly bracket here? Do we also want to return NULL?

Pre_processing_scripts/01_Sort_IT_Extracts.R
```diff
@@ -0,0 +1,129 @@
...
+
+  csv_files_size <- paste0(
+    round(file.size(csv_files) %>% sum() / 1024^3, 1),
+    "G"
+  )
+  message <- stringr::str_glue(
+    "There are {available_space} avaialable in hscdiip, and the rest of operation requires around {csv_files_size}."
+  )
+  cli::cli_alert_warning(message)
+}
```
Jennit07, line 101: Nice checks here!

Pre_processing_scripts/01_Sort_IT_Extracts.R
```diff
@@ -0,0 +1,129 @@
...
+)
+it_extract_anon_path <- file.path(
+  it_extract_path,
+  "anon-chi-IT"
+)
+
+is_chi_in_file <- function(filename) {
+  data <- read.csv(filename, nrow = 1)
+  return(grepl("UPI", names(data)) %>% any())
+}
```
Jennit07, line 16: I wonder if we can split some of the functions out into a separate functions script and use source to call them. Some of these can be applied to sort_BI_extracts and write_anon_chi_files

Pre_processing_scripts/01_Sort_IT_Extracts.R
Jennit07, line 46:
```suggestion
new_task_num <- get_current_it_task_num()
```
name change suggestion

Pre_processing_scripts/pre_prossessing_all.R
Jennit07: I think this is a good idea but i think having this as separate is more beneficial as we dont have to run every script each quarter in this folder and in any particular order. I agree we could clear this up/organise this more. Think we should also have some more documentation notes and structure. E.g. `write_anon_files` notes

R/get_it_extract_paths.R
```diff
@@ -13,16 +13,16 @@
...
   } else {
     it_reference <- check_it_reference(it_reference)
 
     it_ltc_path <- get_file_path(
-      directory = fs::path(get_slf_dir(), "IT_extracts"),
-      file_name = stringr::str_glue("SCTASK{it_reference}_LTCs.csv.gz")
+      directory = fs::path(get_slf_dir(), "IT_extracts/anon-chi-IT"),
+      file_name = stringr::str_glue("anon-SCTASK{it_reference}_LTCs.parquet")
     )
   }
```
Jennit07, line 27: This looks all good! I think once this branch has merged into the anon_chi branch, we need to do a check for any outstanding places in other scripts using chi or `slfhelper::get_chi/anonchi`

Jennit07 (changes requested):
Thanks Zihao, great work on this! I have made a few minor suggestions regarding name changes to make things a bit clearer. Overall, the logic and code looks really good! Thanks

Pre_processing_scripts/01_Sort_IT_Extracts.R
lizihao-anu: It basically converts chi to anon_chi. If there is no chi, we still write out data to parquet. So no need for `else` part.

Pre_processing_scripts/01_Sort_IT_Extracts.R
lizihao-anu: Yes. That's right. Put this into a new issue and get it done in the next update.

Jennit07 (approved):

PR #1063, opened 2025-03-26
remove joining na when linking chi with person_id in care home

### LINKED ISSUE #1061: Something wrong with Care Home data when CHI is missing
From James:

Something wrong with Care Home data when CHI is missing
Hey guys, I was doing some work with Social Care in the SLFs (just putting together some simple 'best-practice' examples, and I noticed some weirdness for episodes where the CHI (anon_chi) is missing. See this example:
 
library(slfhelper)
library(dplyr)
care_home <- read_slf_episode(
  year = "2223", 
  recids = "CH", 
  col_select = c("year", "anon_chi", "person_id", "social_care_id", "sc_send_lca"))
 
The person_id and the social_care_id are all appearing as 120-PER121222 / PER121222 but a lot of the rows clearly should be different people, e.g. they have a different sc_send_lca, not the 120 as person_id is implying. Also if you pull more columns you'll see different dob, postcode, etc. etc.

lizihao-anu:
Close #1061

### DIFF
--- a/R/create_episode_file.R
+++ b/R/create_episode_file.R
@@ -547,7 +547,8 @@ join_sc_client <- function(data,
       dplyr::left_join(
         sc_client,
         by = "anon_chi",
-        relationship = "many-to-one"
+        relationship = "many-to-one",
+        na_matches = c("never")
       )
   } else {
     data_file <- data %>%


Jennit07 (approved):
Looks good

PR #1064, opened 2025-03-26
change the code due to update of spd datazone

lizihao-anu:
Scottish Postcode Directory has been updated to version March_2025. `datazone2022` has been added and while the codes related to datazone are hard coded as `datazone2011`. 

I was thinking of automatically making the code automatically pick up latest datazone. However, some files downloaded from boxi are still having variables `datazone2011` and there is no option to add `datazone2022` yet. So, for now, it is better to use `datazone2011`.


Jennit07 (approved):
Looks good. We should review this again when we would need to use `datazone2022`

PR #1071, opened 2025-04-08
1070 create a common function is chi in file for pre processing

lizihao-anu:


Pre_processing_scripts/00_Sort_BI_Extracts.R
```diff
@@ -21,10 +21,8 @@ extract_financial_year <- function(filename) {
 }
 
 # Create a function to read variable names
-is_chi_in_file <- function(filename) {
-  data <- read.csv(filename, nrow = 1)
-  return(grepl("UPI", names(data)) %>% any())
-}
+# load is_chi_in_file() function
+devtools::load_all()
```
Jennit07, line 25: Think we can move load_all to the top of each script? otherwise looks good! Thanks

PR #1077, opened 2025-05-06
Sink console output

lizihao-anu:
divert targets/ep/ind console outputs to txt files


Jennit07 (approved):
Works as expected. Ready to merge

PR #1078, opened 2025-05-07
update is_chi_in_file to allow for parquet file

lizihao-anu:


Jennit07 (changes requested):
I tested this with the DD file we received from the team but as there is CHI in the file and not UPI this function is still failing, could we expand this to include some flexibility for CHI please?

Jennit07 (approved):
Looks good

PR #1115, opened 2025-05-30
Update class of variables 

Jennit07:
This PR should hopefully update the class of variables to match when the data is not available. This causes an issue when trying to read certain years with slfhelper. e.g combining 1415 with 1718. An error occurs when using `bind_rows` due to a mis match in variables.


Jennit07:
All episode and individual files updated. PR ready for review

PR #1154, opened 2025-07-08
fix 00b extract tests

lizihao-anu:
For SMR00, we use `cost_total_net_inc_dnas` from the existing episode file to compare with `cost_total_net` in the new extract. In `create_episode_file()`, `cost_total_net_inc_dnas ` is essentially a copy of `cost_total_net` in the extract and `cost_total_net` in the episode file excludes costs dnas.

### DIFF
--- a/man/produce_source_extract_tests.Rd
+++ b/man/produce_source_extract_tests.Rd
@@ -21,6 +21,9 @@ produce_source_extract_tests(
 \item{max_min_vars}{variables used when selecting 'min-max' from \code{\link[=calculate_measures]{calculate_measures()}}}
 
 \item{add_hscp_count}{Default set to TRUE. For use where \verb{hscp variable} is not available, specify FALSE.}
+
+\item{smr00_ep}{Only for Outpatients extract test.
+TRUE only when using old_data for outpatients}
 }
 \value{
 a dataframe with a count of each flag


R/get_existing_data_for_tests.R
```diff
@@ -31,6 +31,7 @@ get_existing_data_for_tests <- function(new_data, file_version = "episode") {
   if (file_version == "episode") {
     variable_names <- c(
       "anon_chi",
+      "cost_total_net_inc_dnas",
```
OluwatobiOni, line 34:
```suggestion
      "cost_total_net",
```

Test - showing Tobi how to review a PR

R/process_tests_outpatients.R
```diff
@@ -15,12 +15,14 @@ process_tests_outpatients <- function(data, year) {
     old_data = produce_source_extract_tests(old_data,
       sum_mean_vars = "cost",
       max_min_vars = c("record_keydate1", "record_keydate2", "cost_total_net"),
-      add_hscp_count = FALSE
+      add_hscp_count = FALSE,
+      smr00_ep = TRUE
```
Jennit07, line 19: Ive added this parameter for the new change

R/process_tests_outpatients.R
```diff
@@ -15,12 +15,14 @@ process_tests_outpatients <- function(data, year) {
...
     ),
     new_data = produce_source_extract_tests(data,
       sum_mean_vars = "cost",
       max_min_vars = c("record_keydate1", "record_keydate2", "cost_total_net"),
-      add_hscp_count = FALSE
+      add_hscp_count = FALSE,
+      smr00_ep = TRUE
```
Jennit07, line 25: Ive added this parameter for the new change

R/produce_source_extract_tests.R
```diff
@@ -30,7 +30,14 @@ produce_source_extract_tests <- function(data,
...
                                          ),
-                                         add_hscp_count = TRUE) {
+                                         add_hscp_count = TRUE,
+                                         smr00_ep = FALSE) {
+  if(smr00_ep){
+    data <- data %>%
+      dplyr::select(-"cost_total_net") %>%
+      dplyr::rename(cost_total_net = cost_total_net_inc_dnas)
+  }
+
```
Jennit07, line 40: We have tested the outpatients data to try an reproduce the tests but its not working. It needed the parameter above. But also there is an error coming through. Could you please have another look?

Error in dplyr::rename():
! Can't rename columns that don't exist.
✖ Column cost_total_net_inc_dnas doesn't exist.
Run rlang::last_trace() to see where the error occurred.

Jennit07 (dismissed):
Hi @lizihao-anu We have reviewed your PR but its not working right. Could you please have another look? Thanks

R/produce_source_extract_tests.R
OluwatobiOni: Thanks Zihao for fixing this. The error no longer exists and the code runs smoothly.

PR #1159, opened 2025-08-04
Demog chpc homepc

lizihao-anu:
Summary statistics of change:

- Around 0.08% demographics episode home postcode will be updated.
- Around 11.7% demographics episode will be updated. Mostly because of
  - social care id fix in Renfrewshire, 
  - then introduce of latest extract date. The previous method arrange the time order using period eg"2023Q4". But one cannot tell which is newer when there are two "2023Q4"s. So, extract_date may help in this scenario. 
  - The fix of social care id in Renfrewshire and introduction of extract_date are inspired by Social Care team.
- `right_join` and `full_join` have no difference on the files of SDS, CH, client (by year), etc.

### DIFF
--- a/R/process_lookup_sc_demographics.R
+++ b/R/process_lookup_sc_demographics.R
@@ -28,7 +28,7 @@ process_lookup_sc_demographics <- function(
     dplyr::pull(.data$pc7)
 
   # remove per in social_care_id in Renfrewshire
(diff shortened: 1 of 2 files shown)


Jennit07 (approved):
Thanks Zihao, ready to merge

PR #1171, opened 2025-08-18
Bug fix: slfhelper not combining multiple years

OluwatobiOni:
Created functions that contains the list of variables in the episode and individual files. The function is then called out before writing to disk to ensure columns are structured in the same order across years. This allows us to combine multiple years.


Jennit07 (commented):
Hey @OluwatobiOni this looks good! I had a thought that since we are reordering variables, we could put them into a more meaningful order where all the dataset variables are grouped e.g. HL1 variables grouped together, sc variables grouped together etc. Are you able to make this change? would be on both functions then need to update the variable layout lists as well for the customers. Thanks

OluwatobiOni:
Great idea @Jennit07, I'll look into that.

OluwatobiOni:
Hi @Jennit07 , I have now rearranged the ordering. Feel free to take a look when you have the time. Thanks!

R/order_ep_cols.R
```diff
@@ -0,0 +1,298 @@
...
+      "apr_cost",
+      "may_cost",
+      "jun_cost",
+      "jul_cost",
+      "aug_cost",
+      "sep_cost",
+      "oct_cost",
+      "nov_cost",
+      "dec_cost",
+      "social_care_id",
```
Jennit07, line 213: Could we please move `person_id` under `social_care_id`? i think that makes more sense as its a social care derived variable

Jennit07 (changes requested):
Hi @OluwatobiOni ive made a few suggestions, but otherwise looks great and ready to merge! Thanks

Jennit07:
Thanks @OluwatobiOni great work!

PR #1195, opened 2025-09-18
Social care - latest record flag

Jennit07:
I have tried to align our latest record flag methodology with the social care team as close as possible. However this still needs more work and also testing to see the impact when we join the sc demographics lookup onto each sc dataset. I am happy that the demographics lookup will be improved by adopting the sc team methodology and should bring our client counts closer. 

For example: the previous SLF methodology would keep both CHIs when one scid is associated with more than one CHI causing an over inflation in the SLFs. This new methodology aligns with the social care team by choosing the latest record and the associated CHI. The scid and CHI will be one-to-one which is also what the social care team do. We also need to check what happens when matching at the episode file stage and any blank CHIs. 

For info, here is the link to the social care team's code which deals with the latest record flag and care home postcode: 
https://github.com/Public-Health-Scotland/social-care-publication/blob/2425_common_extracts/Common%20Extracts/scripts/1_demo_processing.R

Please note: i have reordered the `process_sc_lookup_demographics` script to flow better. The latest record flag will be dealt with separately to the ch postcodes. This is also what the social care team do. This may result in a conflict when CH postcodes are investigated

I will open this as a PR just now. Feel free to make any comments/suggestions/changes

### DIFF
--- a/R/cost_uplift.R
+++ b/R/cost_uplift.R
@@ -1,16 +1,18 @@
 #' Uplift costs
 #'
-#' @param data episode data
+#' @param data episode or extract data
 #'
-#' @return episode data with uplifted costs
-#' @family episode_file
+#' @return data with uplifted costs
+#' @family uplift_costs
 apply_cost_uplift <- function(data) {
   data <- data %>%
     # attach a uplift scale as the last column
-    lookup_uplift() %>%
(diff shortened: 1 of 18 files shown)


Jennit07:
Closing as this has been incorporated in the PR #1202

Issue #1185, opened 2025-09-15
SC extract tests issue


OluwatobiOni:
The way we count client number is count distinct anon_chi. Lots of clients have only social_care_id and no anon_chi. So the number is less. AT, CH, HC, SDS, clients are all affected.

PR #1198, opened 2025-10-09
sc extract tests count fix

OluwatobiOni:
SC extract tests issue #1185

Two issues are fixed in this pull request:
1. The way we count client number is count distinct anon_chi. Lots of clients have only social_care_id and no anon_chi. So the number is less. AT, CH, HC, SDS, clients are all affected. This is fixed by considering distinct social_care_id alongside distinct anon_chi
2. n_missing_chi shows 1 in the sc extract tests. This is because we calculate distinct anon_chi and therefore all NAs are regarded as 1 (distinct). To fix this, the sum of NAs is first calculated and this is used in the process tests comparison table, therefore showing the correct number of n_missing_chi.

### DIFF
--- a/R/00-update_refs.R
+++ b/R/00-update_refs.R
@@ -48,30 +48,31 @@ end_date <- function() {
 #'
 #' @return A logical TRUE/FALSE
 check_year_valid <- function(
-    year,
-    type = c(
-      "acute",
-      "ae",
-      "at",
-      "ch",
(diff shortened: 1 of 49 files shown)


lizihao-anu:
Missing dob and missing postcode needs pre-calculation as well. Because distinct by chi will also remove dob and postcode being NA.

PR #1205, opened 2025-11-10
remove quality 19, 20 as they do not make sense

lizihao-anu:
improve `fill_ch_names()`. Remove quality 19 and 20, which do not make sense.


Jennit07 (approved):
Looks good

PR #1207, opened 2025-11-14
export console output to txt file

OluwatobiOni:
I have fixed the sink code block to export the console output to the correct directory making sure to close sink off when the process is complete or if code breaks due to error. I have tested it to make sure it works too.


Run_SLF_Files_manually/run_episode_file_1415.R
```diff
@@ -32,25 +32,35 @@ write_temp_to_disk <- TRUE
...
-# }
+# #-------------------------------------------------------------------------------
+# # save console outputs if `console_outputs == TRUE`
+if (console_outputs) {
+  update <- latest_update()
+
+  con_output_dir <- "/conf/sourcedev/Source_Linkage_File_Updates/_console_output/"
+
+  file_name <- stringr::str_glue(
+    "ep_{year}_{update}_update.txt"
```
Jennit07: I like this idea, i think it would be good if we could also incorporate the sys.time in the file name if the file runs again. This would be handy if we needed to compare outputs and check the latest log was successful

Issue #1162, opened 2025-08-12
investigate episode file tests


Jennit07:
Changes made to social care tests will also need to be applied to the episode file. Eg. currently counting number of submissions by area but we should try count the number of people/clients by area

PR #1219, opened 2026-01-13
improve episode file process tests

OluwatobiOni:
Following the release of the Dec25 update, we found that the episode file tests are counting the number of records for social care datasets instead of the number of clients like the extract tests. I think this may require some further investigation and some changes, however we would like this to match counting the number of clients in the episode file. Please could you pick this up and see if you could get this working? You may need to refactor the tests to get this working – that is absolutely fine, you can take a new branch off master to work on any of these developments.

Issue #1162 - Changes made to social care tests will also need to be applied to the episode file. Eg. currently counting number of submissions by area but we should try count the number of people/clients by area

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -45,7 +45,7 @@ jobs:
 
       - name: Deploy to GitHub pages 🚀
         if: github.event_name != 'pull_request'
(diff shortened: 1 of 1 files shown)


OluwatobiOni:
> Is it supposed to merge into development branch?

Into master as it is part of the main processing code. BYOC development merge into the development branch.

PR #1252, opened 2026-02-03
Run sdl fix

lizihao-anu:
Remove `devtools::load_all()` and use `library(createslf)` instead. So we need export some functions. 

More functions are expected to be exported.


LisaHopcroft (commented):
Just a couple of queries below.

Let me know your responses and then I can approve.

PR #1257, opened 2026-02-17
Bug - fix byoc file paths

Jennit07:
In PR #1253 this broke some of the file paths. for example: `read_extract_acute()` returned the error: 

```
Error in get_dummy_boxi_extract_path() : 
  argument "BYOC_MODE" is missing, with no default
Called from: get_dummy_boxi_extract_path()
```
This was due to the `BYOC_MODE` having no default. I have set this to `FALSE` for now in this PR as this allows us to still have functionality during the refactoring process. 

Could this please be reviewed but not merged into development for now until we have confirmation from NSS. Thanks

### DIFF
--- a/R/get_boxi_extract_path.R
+++ b/R/get_boxi_extract_path.R
@@ -39,9 +39,8 @@ get_boxi_extract_path <- function(
     # Since BOXI extracts will be only in sourcedev and not needed in Denodo,
     # BYOC_MODE here will be always FALSE.
 
-    # TODO: JT - Check here to see if this comment is still relevant. My understanding
-    #       is that any static files will be converted into Denodo VDBs. This code
-    #       may be redundant during refactoring.
+    # Hence, this function or the if-else part of BYOC_MODE being FALSE will
+    # be completely removed after refactoring is completed.
     if (type %in% c("dn", "cmh")) {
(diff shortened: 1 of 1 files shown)


Jennit07:
> That's not what we agreed before.
> 
> The error for `read_extract_acute()` is because it has not been refactored yet. See `read_extract_homelessness()` which has been refactored in the development. The `read_extract_xxx()` should be reframed to read from Denode. So, reading from local csv BOXI files is supposed to be removed.
> 
> Do you want to keep the functionality of reading from local csv files in `read_extract_xxx()` for UAT testing? @Jennit07 We can achieve that but need to agree first then we can apply the same principle to other scripts

Hi Zihao, happy to have a call to discuss this. I understand thats why the datasets would fail as they have not yet been refactored. However, i think we should keep the functionality for now to support UAT testing. This PR achieves this where it maintains the functionality but for other datsets currently in development e.g. homelessness and maternity, those should still be part of the refactored code. Due to the default being set as `FALSE` there is no need to apply this principle to other functions as they all work. I have tested this and can confirm. Happy to have a call. It would be helpful to include Lucy in the call as the default being set as `FALSE` was her idea and she is aware of what would be required for future refactoring. Thanks

PR #1259, opened 2026-02-23
BYOC_ID: 43920 Refactor maternity

lizihao-anu:
Refactoring maternity extracts:

- read_extract_maternity: read from Denodo
- process_extract_maternity: write processed extract to Denodo intermediate drive


R/read_extract_maternity.R
```diff
@@ -3,123 +3,83 @@
 #' @inherit read_extract_acute
 #'
 #' @export
-read_extract_maternity <- function(
-  year,
-  file_path = get_boxi_extract_path(year = year, type = "maternity")
-) {
+read_extract_maternity <- function(year,
+                                   denodo_connect,
```
Jennit07:
```suggestion
                                   denodo_connect = phs_connection()
```
I think we should hard code the connection as the parameter here and put it into a function. That way if someone wanted to run the function, it would all be set up

R/read_extract_maternity.R
```diff
@@ -3,123 +3,83 @@
...
+      postcode = "geo_postcode",
+      record_keydate1 = "date_of_admission",
+      record_keydate2 = "date_of_discharge",
+      selfharm_adm = "self_harm_related_admission",
+      sigfac = "significant_facility_code",
+      spec = "specialty_classification_code",
+      submis_adm = "substance_misuse_related_admission",
+      uri = "maternity_unique_record_identifier",
+      yearstay = "occupied_bed_days"
+    ) %>%
```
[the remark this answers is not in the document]
Jennit07: Please see `read_sc_all_alarms_telecare()` as an example

R/read_extract_maternity.R
LisaHopcroft: Is the intention that this connection will be created using the environment variables that we will be populated as part of the wrapper script?

**UPDATE**: Confirmed elsewhere: https://github.com/Public-Health-Scotland/source-linkage-files/pull/1259#discussion_r2911023418

LisaHopcroft (commented):
See comments below in response to review.

R/get_boxi_extract_path.R
```diff
@@ -10,75 +10,67 @@
...
+  )
 
-    boxi_extract_path_csv_gz <- fs::path(
-      dir,
-      stringr::str_glue("{file_name}-20{year}.csv.gz")
-    )
+  if (BYOC_MODE) {
+    return(stringr::str_glue("valid_denodo_extract_{file_name}-20{year}"))
+  } else {
+    boxi_extract_path_csv_gz <- fs::path(dir, stringr::str_glue("{file_name}-20{year}.csv.gz"))
```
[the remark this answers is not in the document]
Jennit07: Hi @LisaHopcroft If BYOC mode is TRUE then it will return the denodo extract name but if this is FALSE and running locally, this will point to the current directories. The denodo extract name may be superseeded by #1265

R/read_extract_maternity.R
Jennit07: @LisaHopcroft yes, the connection will be put into a function called `get_denodo_connection` PR open: #1263 and within the function, this will handle the environment variables. This is just passing the connection to each `read_XXX` script

LucyEmma22 (commented):
Approved

LucyEmma22 (commented):
Approved

LucyEmma22 (commented):
Approved

PR #1263, opened 2026-02-25
BYOC_ID: 44543 Get denodo connection

OluwatobiOni:
**Zihao:**

Hi Tobi,
 Would you be able to review the `get_connection_PHS_database.R` script in the Source Linkage Files repository? 

As you’ll see, it uses the keyring package to securely store a user’s credentials, meaning the username and password do not need to be entered each time a database connection is established. The current implementation relies on Windows (LDAP) credentials for connecting to the PHS databases.

For the automation work, we are planning to follow the same approach for Denodo connections. The idea is to use keyring to securely store each user’s Denodo credentials (with a different password), so that local runs can authenticate automatically — similar to how we currently handle the social care datasets.


OluwatobiOni:
As requested by @lizihao-anu, I have added the parameter to take credentials from system environment when running in BYOC_MODE and keyring when running locally.

lizihao-anu (approved):
Looks good. It works well both in local and BYOC.

PR #1265, opened 2026-03-05
BYOC_ID: 44544 add get_byoc_intermediate_path.R

OluwatobiOni:
Writing a dynamic path-generation function and mapping logic to standardise Denodo intermediate file locations, enabling the wrapper to selectively process and save specific datasets based on a provided list.


LisaHopcroft (commented):
This looks fine, though we will also need another mapping that links the dataset type to the S3 bucket that will be the source for the Snowpipe. This is something that could be implemented as another named list at your end or implemented as a view in Denodo at our end. My preference is the latter which will mean that if we change the S3 bucket used at our end, you wouldn't have to update your implementation.

I have created a ticket at our end to create this view. You can leave this PR open until we have created this view and you have used it to copy files to the relevant S3 buckets.

One other thing to mention, it's a little confusing that this function is called `get_byoc_intermediate_path.R` as there is a variable in the wrapper script called `BYOC_COPY_INTERMEDIATE_LOGS` which controls the printing of intermediate logs (which was necessary for other projects, but not likely to be used here). I am intending to remove the `BYOC_COPY_INTERMEDIATE_LOGS` variable from the wrapper script.

Can you confirm that you are not using `BYOC_COPY_INTERMEDIATE_LOGS` for any SDL logic?

OluwatobiOni:
Changes have been made as discussed with Lisa. The updates in this PR adds support for generating a dynamic list of BYOC output files so that the wrapper can selectively copy only the datasets produced in a given run.

Changes included:
- Added `get_byoc_intermediate_path()` to standardise generation of byoc intermediate output paths.
- Added helper function `get_byoc_output_files()` to build a named list of output file paths.
- Introduced `byoc_output_files` variable in run_sdl.R, which is passed to the byoc wrapper and used to determine which files should be copied to S3.

@Jennit07 @lizihao-anu

PR #1324, opened 2026-05-11
update get-byoc-output-files

OluwatobiOni:
Created a function to retrieve processed SDL datasets, supporting both BYOC and local modes:

- In BYOC mode, data is pulled directly from Denodo
- In local mode, datasets are loaded locally using appropriate file path functions

It covers all identified datasets and adds safeguards for unknown dataset types. This provides a single, consistent interface for accessing processed data regardless of environment.

TO-DO: This script/PR will need to be updated when BYOC_MODE is updated as a parameter to some of the get-file functions.


R/get_byoc_output_files.R
```diff
@@ -1,97 +1,140 @@
...
-  all_types <- c(year_specific_types, non_year_specific_types)
 
-  # if types not supplied, use all types
-  if (is.null(types)) {
-    types <- all_types
-  }
+  if (BYOC_MODE) {
+    sdl_name <- dplyr::recode_values(
+      type,
+      "acute"                        = "sdl_acute_processed",
```
LisaHopcroft, line 38: Where are you using the `*_processed` values? In my mapping spreadsheet, the `*_processed` values refer to tabs in the specification, not views. Is this function intending to identify the source view for each of the datasets? If so, the views are of the format `sdl_{dataset}_source`.

lizihao-anu (approved):

OluwatobiOni:
Reopened in Get SDL Helpers and a new function added to the PR.

PR #1330, opened 2026-05-22
Fix date format

lizihao-anu:
Changes made:
- Explicitly make date variables of date data type.
- Make `run_sdl.R` stop once targets error
- pkgdown error fix


Jennit07 (approved):
Looks good

PR #1321, opened 2026-05-04
BYOC_ID: 46026 - Refactor hhg

lizihao-anu:
Combined two tickets together

PR #1336, opened 2026-06-08
BYOC_ID: 46016 - Refactor SPARRA

Jennit07:
Split out content from #1321

SPARRA refactoring - risk date must be calculated from 1st April for the financial year. This is a projection score. e.g. 01/04/2025 would be included in the 2025/26 financial year.


R/get_sparra_hhg_paths.R
```diff
@@ -47,3 +47,32 @@ get_sparra_path <- function(year, ...) {
...
+  if (isTRUE(BYOC_MODE)) {
+    extract_sparra <- dplyr::tbl(
+      denodo_connect,
+      dbplyr::in_schema("sdl", "sdl_sparra_source")
+    ) %>%
+      ## TODO: Check filters/sparra risk year ##
+      collect()
+  } else {
+    extract_sparra <- read_file(get_sparra_path(year))
+  }
```
Jennit07: Discussed with LB: Should we always read from Denodo if BYOC_MODE is TRUE/FALSE

Jennit07:
Summary: 
- Split out sparra from #1321 
- kept get_sparra_path the same
- new function `get_sparra_data` for reading from denodo

TODO: 
- re-test data from changes in UAT 


PR is ready for checking/PHS testing

R/get_sparra_hhg_paths.R
lizihao-anu: My opinion is that we should always read from Denodo. We will no longer download BOXI extracts. Hence, we will not read from local for BOXI. For the consistency, we should also do that (ie read from Denodo) for other files we uploaded to Denodo.

lizihao-anu (commented):
See comments about reading from Denodo for local files

R/get_sparra_hhg_paths.R
OluwatobiOni: I agree.

R/get_sparra_hhg_paths.R
```diff
@@ -47,3 +47,36 @@ get_sparra_path <- function(year, ...) {
...
+#' @return The path to the SPARRA extract as an [fs::path()]
+#' @export
+#' @family extract file paths
+#' @seealso [get_file_path()] for the generic function.
+get_sparra_data <- function(year,
+                            denodo_connect = get_denodo_connection(BYOC_MODE = BYOC_MODE),
+                            BYOC_MODE) {
+  year <- check_year_format(year, format = "fyyear")
+  c_year <- convert_fyyear_to_year(year)
+
```
Jennit07, line 69: Add logger messages here

OluwatobiOni: This is pending and logger messages have not been added yet

R/get_sparra_hhg_paths.R
```diff
@@ -47,3 +47,36 @@ get_sparra_path <- function(year, ...) {
...
+    extract_sparra <- dplyr::tbl(
+      denodo_connect,
+      dbplyr::in_schema("sdl", "sdl_sparra_source")
+    ) %>%
+      # Filter on projection variable: `risk_financial_year` e.g.
+      # 01-04-2025 relates to financial year 2025/26.
+      dplyr::filter(risk_financial_year == c_year) %>%
+      dplyr::collect()
+
```
OluwatobiOni: I don't see anonymisation of chi being done in the code. 

Is it missing this step here? - slfhelper::get_anon_chi("patient_chi")

R/get_sparra_hhg_paths.R
```diff
@@ -47,3 +47,36 @@ get_sparra_path <- function(year, ...) {
...
+  on.exit(try(DBI::dbDisconnect(denodo_connect), silent = TRUE), add = TRUE)
+
+    extract_sparra <- dplyr::tbl(
+      denodo_connect,
+      dbplyr::in_schema("sdl", "sdl_sparra_source")
+    ) %>%
+      # Filter on projection variable: `risk_financial_year` e.g.
+      # 01-04-2025 relates to financial year 2025/26.
+      dplyr::filter(risk_financial_year == c_year) %>%
+      dplyr::collect()
```
OluwatobiOni:
```suggestion
      dplyr::collect() %>%
      slfhelper::get_anon_chi("patient_chi")
```

OluwatobiOni:
The code generally looks okay and I am able to read the data from denodo using the refactored code. It looks stacked by FY as expected and I am able to filter on `risk_financial_year`. 

The code and denodo variables available also look alright when the logic we used is compared with the original used to extract the data for us by the SPARRA team. It also looks okay to match on to an ep file using the test code provided by Zihao. 

Just a few things to flag/change:
1. Needs logger messages added
2. I don't see anonymisation of patient_chi in the code. Might have been missed

@Jennit07

Jennit07:
Update 15/09/2026

On hold. Do not merge into development until testing with the episode file

PR #1344, opened 2026-07-01
Update Refactor LTC

LucyEmma22:
Add year column to LTC data during processing.


LucyEmma22:
When testing locally I get the error `There were bad CHI numbers in the LTC file` which comes from line 58-60 of `read_lookup_ltc.R` and is caused by 5 records where `chi_check` is `Invalid checksum`.

Jennit07 (commented):
Getting this error when trying to run the `read_lookup_ltc`

`Error: 'recode_values' is not an exported object from 'namespace:dplyr'`

Jennit07:
I've tried running the code and i also get the folowing error: 

`Error: There were bad CHI numbers in the LTC file`

We should feedback to NSS at the stand up tomorrow

LucyEmma22:
Invalid CHI numbers have been removed by NSS.

LucyEmma22:
Passed PHS local testing.

Jennit07 (approved):

